Issue #18, opened 2016-01-27
Content container falls below left navigation in Safari 8.x


mattdeco:
The AMP docs have a layout issue in older versions of Safari that support Flexbox (Safari 8.x in particular) where the content div wraps below the left navigation. This happens because of a missing legacy `display: -webkit-flex;` style declaration in the `.main` rule in `_sass/content.scss`.

Steps to reproduce:
1. Navigate to https://www.ampproject.org using Safari 8.x / OS X 10.10.x ("Yosemite")
2. Click on Docs in the primary navigation menu
3. Observe that the main content area wraps below the left side navigation menu

See attached screenshot for more detail:
![safari8-amp-docs-layout](https://cloud.githubusercontent.com/assets/777787/12631115/8f9c1b58-c504-11e5-9fd4-4e7de5b58795.jpg)

pbakaus:
Thanks for catching this, merged your PR so closing this issue.

PR #19, opened 2016-01-27
Add flex container style declaration to support older versions of Safari

mattdeco:
This change addresses the Safari layout bug mentioned in #18


pbakaus:
Good find! LGTM.

Issue #21, opened 2016-01-28
Jekyll error occurs while generating amp-access-spec documentation


britice:
After importing reference docs with `import_reference_docs.js`, the following error occurs when running `jekyll serve`:

```
Configuration file: /<private>/docs/_config.yml
            Source: /<private>/docs
       Destination: /<private>/docs/_site
 Incremental build: disabled. Enable with --incremental
      Generating...
  Liquid Exception: Syntax Error in tag 'highlight' while parsing the following markup: amp-mustache``` 
  Valid syntax: highlight <lang> [linenos] in /<private>/docs/_reference/extended/amp-access-spec.md
```

britice:
It appears that [amp-access-spec.md](https://github.com/ampproject/amphtml/blob/master/extensions/amp-access/amp-access-spec.md) is not formatted in a way that is correctly parseable by the [`convertMarkdown` function of `import_reference_docs.js`](https://github.com/ampproject/docs/blob/master/_scripts/import_reference_docs.js#L40-L63).

The current build of the [amp-access-spec page of the docs website](https://www.ampproject.org/docs/reference/extended/amp-access-spec.html) also appears to have been incorrectly converted but it must have been built without errors. 

Here is a screen-shot sample from the page which appears to have been incorrectly converted to html:
![amp-access-spec](https://cloud.githubusercontent.com/assets/1638007/12658602/fe45c12a-c5d7-11e5-8502-212a3d9a446b.png)

Notice that the list is not correctly formatted and the code formatting seems to set on the wrong content.

britice:
@dvoytenko, it looks like you're the only one who has commits which affect the markdown file mentioned in this issue. I haven't dug into what's causing the formatting issue but it looks like things aren't appearing correctly on the docs site.

dvoytenko:
@andreban Any idea?

andreban:
What is breaking it is actually the `import_reference_docs.js` script. I did 2 things:
1 - Updated the docs with a workaround (added a comma to EOL), to avoid the ```amp-mustache```being treate as highlighted code and fixed some H2 (##) issues.
2 - Created a fix for the script that improves the regexp when handling highlights.

dvoytenko:
It's merged. PTAL.

pbakaus:
As all fixes are merged, I'm closing this. Please let me know if you still have issues.

PR #24, opened 2016-01-29
Improved importing script

andreban:
- Added \n to orderer lists, to make it compatible with Jekyll
- Improved regex to avoid catching part of table and copyright
- Fixed handling inline highlights as html code Issue #21


andreban:
Not working properly.

pbakaus:
This looks great, thanks!

PR #30, opened 2016-02-03
Update new boilerplate in amp docs.

sriramkrish85:


Meggin:
lgtm... with a couple of easy fixes. I also suspect we might need to tweak some more once we see these changes live. And I'm also wondering if we need to explain what this means a bit more somewhere. Thoughts?

sriramkrish85:
PTAL
- on the main repo (amphtml) i created a new file for the boilerplate code so that it would be easy to copy and paste. 
- I dont think we need to explain developers what this is (the css itself is pretty simple and understandable) - also we did not have any explanation for the previous boilerplate.

Meggin:
Sounds good. Merge away.

PR #36, opened 2016-02-09
Fix spelling/grammar mistake in validate.md

PrivateSniper:
Change loose to lose as this is the correct word for this context.


PrivateSniper:
I signed it!

jmadler:
Good spot!  Thanks for the PR.  LGTM, merging.

PR #66, opened 2016-03-23
amp-social-share-config.js PATH changed

ramtel2005:
old config file url went to 404, changed to new url


pbakaus:
The docs repo pulls these changes in automatically from the amphtml repo, changes need to be done here: https://github.com/ampproject/amphtml/blob/master/extensions/amp-social-share/amp-social-share.md

Thanks!

PR #82, opened 2016-04-19
Update validation_errors references to validator-main.protoascii

rzhw:
The validation errors doc currently refers to [validator.protoascii](https://github.com/ampproject/amphtml/blob/master/validator/validator.protoascii), which has since been split into [validator-main.protoascii](https://github.com/ampproject/amphtml/blob/master/validator/validator-main.protoascii) and files under `extensions/*/0.1/validator-*.protoascii`.

This PR replaces references from validator.protoascii to validator-main.protoascii in validation_errors.md.

<del>Instead of merging this as-is it might be more preferable to replace the old references with directions to visit either validator-main or extensions/.../*.protoascii depending on what tag the user wants docs for.</del> Extensions are referenced on the top of validator-main.

### DIFF
--- a/_reference/validation_errors.md
+++ b/_reference/validation_errors.md
@@ -25,7 +25,7 @@ The purpose of this document is to help you better understand
 and fix any validation errors you encounter
 when you [validate your AMP pages](/docs/guides/validate.html).
 For a complete overview of the validation errors,
-see the [AMP validator specification](https://github.com/ampproject/amphtml/blob/master/validator/validator.protoascii).
+see the [AMP validator specification](https://github.com/ampproject/amphtml/blob/master/validator/validator-main.protoascii).
 
 {% include toc.html %}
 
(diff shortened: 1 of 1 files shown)


pbakaus:
Did you fix this in the main amphtml repo (https://github.com/ampproject/amphtml)? Since these are auto-generated, your changes would be overwritten as soon as I rerun the generator.

rzhw:
Which file(s) make up the doc? A cursory search through the main repo for some phrases didn't seem to turn up any matches.

Meggin:
Shouldn't the built-in components also reference main? They aren't extensions. For example, see validation errors towards the end of amp-img: https://github.com/ampproject/amphtml/blob/master/builtins/amp-img.md

pbakaus:
Fixed this manually on the new megamerge branch, will be deployed soon! Thanks!

PR #108, opened 2016-06-01
always give hamburger width

erwinmombay:
@pbakaus @Meggin please review. will deploy asap once reviewed


Meggin:
Tested on Nexus 5 using DevTools and iphone 5 in iOS simulator-- both showing hamburger as intended, so lgtm.

PR #112, opened 2016-06-03
Arabic translations for "Get started" & analytics.

yunelee:
- Uses a different set of style sheets for layout elements dependent on the left or right side.
- Attempts to wrap multi-word inline LTR phrases.


pbakaus:
Hey Yune, I implemented this PR manually on the megamerge branch and fixed the rtl styles without causing a lot of duplication. Check it out!

Still missing is porting the translations in _jekyll/data/messages.yml over, but there's currently a bug in "grow extract", which is needed to generate the PO files. I'll leave the PR here open until we figure the last piece out.

pbakaus:
I've manually merged all of this into megamerge, so closing this one! Thanks!

PR #118, opened 2016-06-19
Fix missing quote on several html attributes

rajsite:


dvoytenko:
@rajsite Thanks! Merging.

PR #126, opened 2016-07-08
Update the validation doc.

Gregable:


pbakaus:
This is awesome, thanks much! We can figure out a fix for the images in the next step, will merge for now.

PR #138, opened 2016-07-16
Fix build time issue and two other tiny cleanups

jeremydw:
Hi @pbakaus, this fixes the build time issue. :) I was doing something very stupid and expensive when generating the nav menu (notice the removal of the `O(n)` check). The whole site (all locales) builds in about five minutes on my machine now, and I'm pretty happy with that time.

Next week I'll look into some additional caching or optimizations that can be done on the Grow side, but I think a five minute build for 3000 Markdown files is hopefully reasonable enough for now! [Still would like to make it faster though. ;)]

This PR also contains two minor other things
- Use `doc.locale.is_rtl` instead of checking a hardcoded `ar` locale.
- Update the README to remove `grow deploy` since it's not used in this workflow, nor does that command actually work without a destination anyway.

If you could double check the nav menu to ensure it's still being generated properly, that'd be great. I did a cursory check and I think it looks good.

Cheers!

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,11 +33,10 @@ $ grow run
 
 You can now open http://localhost:8080/ and continue working on the source files, then reload the page to see changes appear.
 
-### Deploy
+### Build
 
 ```sh
 $ grow build
-$ grow deploy
 ```
 
 This will generate a static, complete build of the site into the **build* folder.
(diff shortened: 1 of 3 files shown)


pbakaus:
Ha great find, such small optimizations, so much win :D awesome.

PR #144, opened 2016-07-20
Sidebar needs to open differently depending on lrt/rtl

aghassemi:


views/base.html
```diff
@@ -24,7 +24,7 @@
 
 {% set header_doc = g.doc('/content/includes/header.yaml', locale=doc.locale) %}
 
-<amp-sidebar id='sidebar' side='right' layout='nodisplay'>
+<amp-sidebar id='sidebar' side='{% if doc.locale.is_rtl %}left{% else %}right{% endif %}' layout='nodisplay'>
```
dvoytenko: I believe if side is left undefined, this is the default behavior.

aghassemi: Default is the opposite behaviour (Left sidebar for ltr and Right sidebar for rtl) but AMP Docs explicitly wants the opposite.

pbakaus:
Thanks, SGTM!

PR #162, opened 2016-07-30
Attempt to fix the layout of the code sections.

powdercloud:
I'm inserting newlines before the ```. Hoping this helps.


pbakaus:
If you want to go from "hope" to "know", I recommend trying "grow run" and checking it out locally to see how it looks :) I think this might fix it though, let's try.

PR #163, opened 2016-07-30
Switching the code blocks to pre tags and mild escaping &lt; &nbsp;.

powdercloud:


pbakaus:
Hi! There's actually a proper way to do source code on our site. Simply wrap your code in the following blocks:

[sourcecode]
...your source code
[/source]

or if you use a specific language, something like:

[sourcecode:html]
...your HTML source code
[/source]

Could you fix those? I'll merge afterwards. Thanks!

pbakaus:
Thanks, looks good!

powdercloud:
@pbakaus Thank you for your patience with this.

PR #167, opened 2016-08-03
Adding Drugs.com to Who page

NoodlesNZ:


pbakaus:
Could you tell us how you're using AMP and point us to some examples?

NoodlesNZ:
Sure thing, we have been using AMP since it was first announced, but we weren't included in SERPs as we're not a news site. If you use https://g.co/ampdemo and search for lipitor then it should show up our AMP page (https://www.drugs.com/amp/lipitor.html if it doesn't show up).

We've been rolling out AMP across our news pages and basic drug information. Once Google starts rolling AMP pages out to SERPs for everyone then we'll roll it out further.

We even got a mention here: https://techcrunch.com/2016/08/02/googles-amp-expands-beyond-news/

pbakaus:
Thanks for the clarification, looking good!

PR #170, opened 2016-08-12
Bunch of RTL fixes

aghassemi:
bunch of RTL fixes for the hamburger icon and nav.
`rtl` class on `body` was also not being applied (white-space bug)


pbakaus:
LGTM, thanks

PR #171, opened 2016-08-14
Including Namshi  in MENA

odino:
Greetings from Dubai! :)

We're an online retailer and are following the good work put on by eBay on AMP -- excited to see AMP growing for e-commerce! 

namshi-amp

### DIFF
--- a/content/includes/who.yaml
+++ b/content/includes/who.yaml
@@ -232,6 +232,10 @@ partners:
         - USA Today
         - Vox
         - Washington Post
+        
+    - title@: MENA
+      section_items:
+        - Namshi
 
 tech_companies:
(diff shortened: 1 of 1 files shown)


odino:
I signed it!

pbakaus:
Thanks much, we'd definitely like to have you on the list! That said, we're trying to secure a few other names for the MENA section before launching that section (so it's not that empty) – I'll keep you updated.

pbakaus:
Hi @odino,

I apologize for the late turnaround. In the meantime, we have decided to simplify the Who page and rename it to "Supported platforms". In the process, we removed all mentions to publishers – there are just too many to keep track of, and it doesn't provide immediate value to the website user.

PR #178, opened 2016-08-30
adds ATTN: to Who section

tgallant:
Hi There,

We have recently set up Google AMP on http://www.attn.com and would love to get added to the Who page!

Let me know if you need anything else

### DIFF
--- a/content/includes/who.yaml
+++ b/content/includes/who.yaml
@@ -188,6 +188,7 @@ partners:
         - 1-800-Flowers
         - 247Sports
         - ABC News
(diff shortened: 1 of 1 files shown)


tgallant:
I signed it!

pbakaus:
Hi @tgallant,

I apologize for the late turnaround. In the meantime, we have decided to simplify the Who page and rename it to "Supported platforms". In the process, we removed all mentions to publishers – there are just too many to keep track of, and it doesn't provide immediate value to the website user.

PR #184, opened 2016-09-16
Add reference to HTML layout spec; fixes #173

### LINKED ISSUE #173: ampproject.org: Include instructions on image size specs & image resizing
Prevent errors like blurred images, captions missing, images missing. Clarify what are common specs, what is the difference between responsiveness and rescaling, what can be consequences of misimplementations or - rephrasing this - highlight common errors.

bpaduch:
Added reference to HTML layout spec per issue #173


Meggin:
I would also recommend putting links to layout docs and spec in the amp-img file (that one is in amphtml) as per the issue raised. Might as well have it in both places.

bpaduch:
Updated reference docs in this request: https://github.com/ampproject/amphtml/pull/5063

PR #210, opened 2016-10-03
Update footer.html

james4388:
Missing close tag


james4388:
I signed it!

pbakaus:
Good catch! Thanks.

PR #226, opened 2016-10-12
Fix sidebar responsiveness

derekcecillewis:
When opening the sidebar in mobile viewport then making viewport wider, sidebar styles are removed. This PR removes the media query that restricts the sidebar styles to only "mobile", allowing for a consistent mobile experience (think switching from portrait to landscape with sidebar open on devices with unusually large screens). Compare the before and after images below.

960-before
960-after

### DIFF
--- a/assets/sass/_sidebar.scss
+++ b/assets/sass/_sidebar.scss
@@ -1,164 +1,160 @@
-@include max-screen($mobile-breakpoint) {
+amp-sidebar .hamburger {
+  display: inline-block;
+  background-image: url("/static/img/hamburger_white.svg");
+  background-repeat: no-repeat;
+  background-size: 30px 24px;
+  background-position: center center;
+  background-color: transparent;
+  width: 40px;
+  height: 40px;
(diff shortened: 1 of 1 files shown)


pbakaus:
I'm torn about this. I get where you're coming from, but I do think it is a scenario that almost never happens, and now we're giving Desktop browsers more work to do because they need to apply and parse more rules (theoretically at least). Thoughts?

derekcecillewis:
Although this scenario may occur rarely, it is awful when it does. I hope you don't intend on leaving it this way. Even if desktop browsers don't parse and apply rules in media queries that don't pertain to the current viewport dimensions, this problem is not exclusive to mobile devices. I actually discovered it on a desktop computer. Also, assuming your theory is correct, please keep in mind that desktop computers generally have more processing power—parsing and applying these [theoretically] additional rules would likely be trivial for desktop browsers.

pbakaus:
Alright convinced! Desktop should be able to handle the additional parsing overhead.

PR #231, opened 2016-10-18
Align globe icon on right for RTL Languages

navarr:
Fixes an issue where the globe icon is placed over text in the language dropdown when the docs are rendered in RTL mode

I'll be honest, I've only tested this in CSS not SCSS so I'm unsure if this particular change will work.  I do not currently possess the capabilities to test this diff.


pbakaus:
That was not the right fix :) Just applied the right fix in c398654. Thanks anyway!

navarr:
:thumbs-up: whatever it takes

PR #247, opened 2016-10-31
Fix wrong calculation for JS loading in docs

skatpgusskat:
The calculation for loading time was wrong.

> For example, if you have five ads on your page, and each of them cause three synchronous loads,
each with a 1 second latency connection,
you’re in **18** seconds of load time just for JS loading.

### My calculation
5 ads * 3 sync loading * 1 second = 15 seconds.


skatpgusskat:
@googlebot  I signed it!

pbakaus:
Good catch, thanks!

PR #248, opened 2016-10-31
Remove hx:before tag blocking scroll

skatpgusskat:
I found that I can't scroll the script box below on [this pages](https://www.ampproject.org/learn/how-amp-works/#dont-let-extension-mechanisms-block-rendering).
> ```<script async custom-element="amp-iframe" src="https://cdn.ampproject.org/v0/amp-youtube-0.1.js"></script>```

I'm using Chrome 54. 
The ```::before``` components block scrolling. When I delete that, Anything changed but I could scroll.
What actually do ```h(n):before``` styles for?

### DIFF
--- a/assets/sass/_global.scss
+++ b/assets/sass/_global.scss
@@ -270,3 +270,7 @@ figure {
 .center-content {
   text-align: center;
 }
+
+.code {
+  position: relative;
+}


pbakaus:
These are needed as a fix for hashbang urls, when linking to a certain header directly (otherwise the scroll position is bad).

I cannot reproduce your issue, but a potentially better fix is to make the code blocks position relative with z-index 1, so they sit on top of the invisible :before.

skatpgusskat:
@pbakaus That make sense, I will try that

skatpgusskat:
@pbakaus I fix PR to make the code blocks position relative, but nothing in z-index. Because when I set ```position: relative```, the scrollbar worked. Do I have to set ```z-index``` property?

pbakaus:
nope, might be enough then. Thanks!

PR #254, opened 2016-11-07
Typos in analytics guides: browser substitution

dbs:
* brower -> browser
* substition -> substitution (also affects navigation)

Fix navigation from the translations as well.

Signed-off-by: Dan Scott <dan@coffeecode.net>


pbakaus:
Good catch, thanks!

PR #266, opened 2016-12-09
Publisher Addition

darrennolan:
Adding Fox Sports Australia as a publisher using the AMP Project.


darrennolan:
I signed it!

pbakaus:
Heya, we moved away from mentioning individual publishers. Sorry it took us so long to reply, we were in the middle of a relaunch of our site. Thanks anyway for your contribution, and for investing in AMP!

PR #271, opened 2016-12-15
Correct bullets so they appear correctly

bpaduch:
Bullets needed a space above to render correctly on amproject site.

to/ @kul3r4 - ptal


kul3r4:
Thanks!
LGTM

Issue #272, opened 2016-12-16
Mustache syntax in codeblocks doesn't render.


bpaduch:
Occurs in [amp-mustache](https://www.ampproject.org/docs/reference/components/amp-mustache) and  [amp-form](https://www.ampproject.org/docs/reference/components/amp-form#successerror-response-rendering).

Any codeblocks that contain mustache syntax (double curly braces) do not render and a bunch of text is gobbled up after the codeblock.  Example codeblock that renders fine in Github but chokes in docs/Grow:

```html
<template type="amp-mustache">
  Hello {{world}}!
</template>
```

[import_docs.js](https://github.com/ampproject/docs/blob/master/scripts/import_docs.js) escapes out curly braces for single code elements but it doesn't work for codeblocks.

PR #273, opened 2016-12-16
Fix mustache syntax rendering - issue 272

bpaduch:
Added requisite of whitespace before mustache code elements so that mustache syntax code in codeblocks will render correctly. Tested and diff'd results.

Fixes issue #272 

to/ @pbakaus


pbakaus:
Nice one, thanks @bpaduch!

PR #275, opened 2016-12-18
fix vertical-align on header navigation

tgfjt:
Hi! 😸 
Since there is together alphabet font and Japanese characters(`ドキュメント`, `サポート`) font,  these height are not aligned.

### Screenshot

Here is a screenshot: on the top is current situation, on the bottom is with `vertical-align: top;`.


pbakaus:
Nice fix, thanks!

PR #280, opened 2017-01-03
html <head> should not translated to 其标头.All html element should be keep in original.

alphayang:
…eep in orignal


alphayang:
I signed it!

PR #288, opened 2017-01-11
Fix url so that login-requiring does not 404

ejcx:
The link[1] on the guide to login-paywalls is 404ing

[1] https://www.ampproject.org/docs/reference/content/docs/get_started/login_requiring.html


pbakaus:
Good catch, thanks!

Issue #98, opened 2016-05-19
Incorrect amp-img examples in ampproject.org


jenniferchui:
On https://www.ampproject.org/docs/guides/responsive/control_layout.html#control-page-layout-using-media-and-media

the example on the page:

<amp-img src="wide.jpg" 
                 **srcset="wide.jpg" 640w, "narrow.jpg" 320w**
                 sizes="(min-width: 650px) 50vw, 100vw" > 
</amp-img>

This example does not validate it with the chrome validator.
(error.js:62AMP-IMG# Layout not supported for: container)

The correct code should be something like this:

<amp-img src="wide.jpg" 
                 **srcset="wide.jpg 640w, narrow.jpg 320w"** 
                 width="320" 
                 height="###" 
                 sizes="(min-width: 650px) 50vw, 100vw"      
                 layout="responsive">
</amp-img>

pbakaus:
Ugh, bad bug. Thanks for discovering! We'll fix soon.

bpaduch:
Fixed

PR #297, opened 2017-01-20
Add height + width to examples for amp-img

bpaduch:
These snippets wouldn't validate without having width and height.  

to: @pbakaus


bpaduch:
Fixes #98

PR #300, opened 2017-01-24
Fixing broken link for amp-mustache component

bpaduch:
Correcting link for amp-mustache per issue - https://github.com/ampproject/amphtml/issues/7149

to: @pbakaus


bhansa:
Well, I don't have any idea how this works, but still I am not able to access the link on the website. When it will be updated, just curious ?

bpaduch:
It was only merged to Master but it's now in Production and the link is fixed.  Thanks for finding it!

PR #303, opened 2017-01-26
Fix broken link to add styles to a page

dandv:
The link to "Add styles to a page" at https://www.ampproject.org/docs/guides/responsive/style_pages is broken - it points to a page about validation.

BTW, why do we have both "responsive" and "responsive_amp" as a URL path element?

https://www.ampproject.org/docs/guides/responsive_amp
vs.
https://www.ampproject.org/docs/guides/responsive/style_pages

### DIFF
--- a/content/docs/guides/responsive_amp/style_pages.md
+++ b/content/docs/guides/responsive_amp/style_pages.md
@@ -11,7 +11,7 @@ Also certain styles are disallowed due to performance implications;
 inline style attributes aren't allowed.
 
 All styles must live in the head of the document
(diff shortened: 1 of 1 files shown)


bpaduch:
https://www.ampproject.org/docs/guides/responsive_amp
-> This guide used to have the title "Create Responsive AMP Pages" but was changed.  The filename remains the same, so as does the URL.

https://www.ampproject.org/docs/guides/responsive/style_pages
-> This was a subpage of the Responsive guide, and that's why "responsive" is in the URL.

PR #306, opened 2017-01-30
Additional French and Spanish translations

yunelee:


bpaduch:
Note that the contribute.md file will be out-of-date quickly as it's being changed significantly in this soon-tobe-merged pull request [305](https://github.com/ampproject/docs/pull/305).

yunelee:
Agreed that the content in contribute@*.md is already outdated, but I took a look at the EN changes and although we'd definitely want those translated as well, I think the old version will still be useful in the meantime -- same links, less detail.

PR #313, opened 2017-01-31
Update discovery guidance to establish "paired" terminology

rudygalfi:
Using the keyword "paired" gives us a helpful way to reference one of the main AMP publishing models. This incorporates that terminology into our getting started guide.


rudygalfi:
Note: I'm not sure how it works to update the translated versions, so I'm just touching the EN version for now.

PR #319, opened 2017-02-07
Add info on noscript fallback for images

bpaduch:
Including information on providing fallback image if user disables scripts. Fixes issue https://github.com/ampproject/amphtml/issues/2327.

to: @cramforce, @jmadler
cc: @thebigbad, @pbakaus


content/docs/guides/amp_replacements.md
```diff
@@ -35,6 +35,16 @@ using the [`amp-img`](/docs/reference/components/amp-img.html) element, like so:
 In this most basic example, the image will display with the specified fixed
 height and width. At minimum, an explicit width and height needs to be set.
 
+As `<amp-img>` relies on JavaScript, if the user chooses to disable scripts, images won't display.  In this case, you should provide a fallback to the image using `<img>` and `<noscript>`, like so:
```
cramforce: I think this deserves a headline.

pbakaus:
If the user chooses to disable scripts, they are also intentionally
crippling their browser, so we should reword this to say "if for some
reason JS doesn't execute". I have no sympathy for users who disable JS.

cramforce:
`<noscript>` is really only for people who intentionally deactivate JS. In cases where for example all JS on the page is blocked by CSP or a content/ad blocker the noscript stuff still doesn't run.

bpaduch:
I find it strange that people disable JS but there's probably use cases for that, especially for the reporter of the issue.  If JS doesn't execute, then that would imply we are recommending this for all cases, and it's not the case - this applies to users who intentionally disable JS.  I'll add the following heading "Displaying images when JavaScript is disabled".

bpaduch:
@cramforce  - ping

PR #344, opened 2017-03-06
Add case studies for Milestone and Terra

bpaduch:
Elena/Paul - please take a look.


bpaduch:
@pbakaus - I just compressed them using a diff tool.  Hope that works.  Will merge to master to get preview on staging. Thx!

pbakaus:
Thanks, lot great but better than before :) Can you merge it onto amp-conf-live when you get a  chance, after it looks good on master? I'm manually deploying to prod from that branch right now.

bpaduch:
Yep, once Elena reviews.

PR #348, opened 2017-03-08
Update style_pages@es.md

SidVal:
There is an error linking: https://www.ampproject.org/es/docs/guides/author-develop/responsive/style_pages#the-custom-fonts-exception


SidVal:
I signed it!

bpaduch:
Thanks @SidVal! After looking at this, it looks like it's a bigger issue with many localized pages. I've logged #352 for this issue.

PR #350, opened 2017-03-09
Make variables translatable + minor css

bpaduch:
1.  Corrected elements in markdown and yaml files to make certain elements (mostly titles) pulled out for translation (add `@`).
1.  Fix CSS so that red highlighting doesn't display for syntax highlighting (like we had before in [PR 286](https://github.com/ampproject/docs/pull/286))


bpaduch:
@leeee - thanks!  Should I extract translations as well?

PR #361, opened 2017-03-13
Fix KO translations to fix broken build

bpaduch:
Docs build broken.  Fixed translations:

1) Removed variables (not needed in translated files)
2) Fixed indentations and extra quotes
3) Fixed msgstr


bpaduch:
cc: @techhtml

PR #370, opened 2017-03-14
Make breadcrumbs more robust

derekcecillewis:
Adds support for infinite chaining as opposed to being restricted to three levels.


derekcecillewis:
cc @lilies

PR #396, opened 2017-03-23
Update messages.po

SidVal:
Translate to spanish all messages.po v1 :)

### DIFF
--- a/translations/es/LC_MESSAGES/messages.po
+++ b/translations/es/LC_MESSAGES/messages.po
@@ -26,14 +26,19 @@ msgid ""
 "business is positioned for the future. Integrate your technology with AMP "
 "today."
 msgstr ""
+"Un creciente número de publicistas y anunciantes están adoptando AMP para entregar "
+"contenido rápido y atractivo a sus usuarios. Como sus clientes desean adaptar la "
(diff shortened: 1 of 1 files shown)


translations/es/LC_MESSAGES/messages.po
```diff
@@ -26,73 +26,85 @@ msgid ""
...
 msgid "AMP HTML Specification"
-msgstr ""
+msgstr "Especificación AMP HTML"
 
 #: /content/learn/overview.yaml
 msgid ""
 "AMP HTML is basically HTML extended with custom AMP properties.\n"
 "The simplest AMP HTML file looks like this:\n"
 msgstr ""
+"AMP HTML is basically HTML extended with custom AMP properties.\n"
```
bpaduch: If there isn't a Spanish translation, you don't need to include it.  Please remove the non-translated text that was added here and down below. Thanks.

SidVal: I putted that, because when I'll translate it, it's more easy replace the text, that add new one.

bpaduch: Hmm, I don't understand because the English text is already right above.  Can you just remove the extra English parts or translate?

SidVal: Done.

PR #405, opened 2017-03-28
Remove <h2> border to give the text room to breathe

pbakaus:
IMHO the border on top of the headings in the docs is unnecessary. The visual separation between blocks is already achieved with the top margin, and the border simply makes it seem more busy. Compare:

## with borders:
image

## without borders:
image

WDYT @bpaduch @leeee ?


SidVal:
Nice move ;)

PR #441, opened 2017-04-19
Fix homepage title

### LINKED ISSUE #424: Homepage has bad title
"Accelerated Mobile Pages Project – Accelerated Mobile Pages Project"

Lets fix the redundancy and maybe add AMP (if we always append the string, lets use AMP there).

bpaduch:
Fixes #424 

to: @cramforce

### DIFF
--- a/content/pages/home.html
+++ b/content/pages/home.html
@@ -1,5 +1,5 @@
 ---
-$title: AMP Project
+$title: Accelerated Mobile Pages Project
 $path: /
 
 $localization:
--- a/content/pages/home.html
+++ b/content/pages/home.html
@@ -1,5 +1,5 @@
 ---
-$title: Accelerated Mobile Pages Project
+$title: AMP Project
 $path: /
 
 $localization:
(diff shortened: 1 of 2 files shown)


views/partials/head.html
```diff
@@ -7,7 +7,7 @@
   <meta name="description" content="{% if doc.description %}{{doc.description}} – {% endif %}Accelerated Mobile Pages Project">
```
cramforce: Also fix this.

content/pages/home.html
```diff
@@ -1,5 +1,5 @@
 ---
-$title: Accelerated Mobile Pages Project
```
cramforce: Lets keep this for now. Works well with the `- AMP` that is added automatically

views/partials/head.html
cramforce: Also wrap the whole tag in `if doc.description`. Much better to have none, than a generic one.

content/pages/home.html
bpaduch: Done

views/partials/head.html
bpaduch: Done and tested locally.

bpaduch: Done

content/pages/home.html
bpaduch: Done

PR #448, opened 2017-04-24
Building the example snippets and host them - part 1 of 2 for resp guide

bpaduch:
Add process to create the sample embed files for sample source file `responsive.html`and host the snippets on https://ampproject-b5f4c.firebaseapp.com/.

After example snippets are built and hosted, I'll publish the guide that references the snippets and embed them as samples.

to: @pbakaus, @sebastianbenz


gulpfile.js
```diff
@@ -6,12 +6,28 @@ const sass = require('gulp-sass');
 const svgSprite = require('gulp-svg-sprite');
 const swBuild = require('sw-build');
 const fs = require('fs');
+const path = require('path');
```
pbakaus: We now have a lower case "path" and a "Path" variable in the same file. Can you rename them so they make more sense in context?

gulpfile.js
```diff
@@ -6,12 +6,28 @@ const sass = require('gulp-sass');
...
+const abe = require('amp-by-example');
+
+const config = {
```
pbakaus: If this is just the config for the ABE task, can you move it into the task?

gulpfile.js
```diff
@@ -6,12 +6,28 @@ const sass = require('gulp-sass');
...
+}
 
 const Path = {
   CSS_SOURCES: './assets/sass/**/*.scss',
   CSS_OUT_DIR: './assets/css/'
 };
 
+// build examples used as sample embeds in docs
+gulp.task('build-examples', function() {
+  abe.generatePreview(config);
```
pbakaus: Is this a synchronous call? If yes, then all good from my side.

sebastianbenz: The call is synchronous, but runs other async calls. I need to properly promisify the API.

gulpfile.js
bpaduch: Fixed

gulpfile.js
bpaduch: Done

PR #449, opened 2017-04-25
Sidebar sanity

pbakaus:
This restores much of the original sidebar, but keeps the new looks intact. Some random cool features:

* Guide sub navigation is flattened again, but still categorized in sidebar
* One menu to rule them all: All docs categories appear in the sidebar and are accessible
* Next/Prev link at bottom that used lots of space with questionable usefulness got removed
* Slightly saner margins and paddings throughout
* Elastic responsive design means quite a bit more available space for the content on smaller resolutions

Enjoy!

image

PS: All redirects should be in place.


bpaduch:
This is good; I'd like to improve the styling so that the categories aren't so washed away, that is, they blend it and almost seem clickable, which they aren't.  Also, it'd be nice to indent the content that's related to the parent category.

pbakaus:
Yeah I agree re: categories. Have fiddled a little but haven't found a solution yet I liked. Will try a bit more today. Thanks!

PR #453, opened 2017-04-25
Rename title to an actionable title

bpaduch:
Didn't apply to other lang files as they are outdated and need full translations.

### DIFF
--- a/content/docs/guides/responsive_amp.md
+++ b/content/docs/guides/responsive_amp.md
@@ -1,5 +1,5 @@
 ---
-$title: Style & Layout
+$title: Apply Styles and Layouts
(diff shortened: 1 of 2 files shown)


pbakaus:
I'm not a fan of the proposed title. Apply doesn't really speak to me here. I changed it to "Style & Layout" because both can be used as verbs, and that's actually what you're doing: You're styling and your layouting (I know, Oxford might disagree on its use as verb but it's common amongst web developers).

bpaduch:
In our discussion, you'd suggested revising it to an actionable title, so I'm confused by this about-face.  We'll leave as is.

PR #454, opened 2017-04-25
Flatten component URLs to bring them back to a predictable URL schema

### LINKED ISSUE #429: Settle on one predictable URL for every amp extended component
The final URL for the various amp component documents has been moving around a bit over time, for example:

1. `https://www.ampproject.org/docs/reference/components/layout/amp-iframe`
2. `https://www.ampproject.org/docs/reference/components/amp-iframe`
3. `https://www.ampproject.org/docs/reference/extended/amp-iframe.html`

The AMP Validator emits these URLs. We have been trying to change the output over time to match changes on ampproject.org. The latest change is problematic, because it is no longer computable from the name of the extension, due to the `/layout/`, `/media/`, etc. category.

Let's decide on a single URL that can be used permanently going forward to reference an amp extension and use this in the validator output. This can simply redirect, but let's also commit to making sure this redirect is not removed in the future.

It looks like we could use `https://www.ampproject.org/docs/reference/components/<amp-component>`, but I'm not sure if this is the recommended URL or not.

pbakaus:
This PR restores the old component URLs (while keeping the new navigation), which in return fixes #429.

/cc @Gregable


Gregable:
Thanks!

Issue #327, opened 2017-02-17
Improve validation docs to discuss tools and approaches for continuous testing


rudygalfi:
Looking at https://www.ampproject.org/docs/guides/validate I think we could definitely do a better job at documenting how these tools could be used in more continuous testing.

It should definitely be possible to use the NPM module (https://www.npmjs.com/package/amphtml-validator) from a karma test, but I agree that some sample code would go a long way!

cc @cramforce

Issue #213, opened 2016-10-04
Using snapshots with old AMP logo


kul3r4:
Go to [validate](https://www.ampproject.org/docs/guides/validate), scroll to "Web interface", the snapshot contains the old logo.

PR #459, opened 2017-04-28
Update Validation Guide with info on npm packages

bpaduch:
Updated validation guide with info on NPM packages that can be used in CI.  Also updated validator image (was outdated).

Fixes: #327, #213

to: @rudygalfi, @cramforce 
cc: @pbakaus


pbakaus:
ah just realized I wasn't the actual assigned reviewer ;) @cramforce @rudygalfi have a look, we can open another PR for fixes.

PR #473, opened 2017-05-09
Create file-a-bug@es.md

SidVal:
This is for translate https://www.ampproject.org/es/docs/contribute/
The part of "file a bug"

### DIFF
--- a/content/docs/contribute/file-a-bug@es.md
+++ b/content/docs/contribute/file-a-bug@es.md
@@ -1,0 +1,5 @@
+---
+$title@: Reportar un error
+$order: 2
+goto: https://github.com/ampproject/amphtml/issues/new
+---


bpaduch:
Hi Sid, this isn't needed as you are already translating the title in [messages.po](https://github.com/ampproject/docs/blob/master/translations/es/LC_MESSAGES/messages.po#L382).

The title in this file is a translation variable (indicated by the `@`), so it's pulled into the messages file.

bpaduch:
Hi Sid, this isn't needed as you are already translating the title in [messages.po](https://github.com/ampproject/docs/blob/master/translations/es/LC_MESSAGES/messages.po#L382).

The title in this file is a translation variable (indicated by the `@`), so it's pulled into the messages file.

SidVal:
Ok. But [here](https://github.com/ampproject/docs/blob/master/translations/es/LC_MESSAGES/messages.po#L382) said that it's translated. Why does not show  in spanish?

bpaduch:
Ah, that is a good find.  I'll submit a fix for that so it picks up the translation variable!

SidVal:
Thanks @bpaduch ... one more doubt.
In the same URL, you will see: `Integrate your technology with AMP` and `Open Source Governance`, where is or should be that translations ? (I'm lost :-1:  )

bpaduch:
Hi Sid, 
Those two are the titles for two docs, that aren't translated, and so they appear as English (same as in the left-hand navigation).  Once translated, they will appear in Spanish.

If you have further questions, join the Slack channel and the "docs" thread for further docs discussions. :)

SidVal:
Oh, sure Barb. 
Thank you again !

Issue #483, opened 2017-05-12
Pull in new components: amp-3q-player, amp-facebook-like, amp-ima-video & amp-timeago


honeybadgerdontcare:
These were recently added.

Now allowed in production:
 - [x] amp-facebook-like https://github.com/ampproject/amphtml/pull/8989

Coming soon in production (~2weeks):
 - [x]  amp-3q-player https://github.com/ampproject/amphtml/pull/7999
 - [x ] amp-ima-video https://github.com/ampproject/amphtml/pull/8782
 - [x]  amp-timeago https://github.com/ampproject/amphtml/pull/8557

bpaduch:
Hi @honeybadgerdontcare  - I've been keeping any eye on these.  Is there a place/link to verify the component is now in Production (without having to run a test AMP page to see if component fails/passes validation).  For example, does the build system now list the component? Does the validator now list the component.

honeybadgerdontcare:
@bpaduch It'd be nice to have such a `system` but I don't think one exists. For validator changes we watch them go through production release and attempt to update the issues when they've gone into prod. Internal tracking is done via go/ampversion.

bpaduch:
amp-ima-video - needs a markdown file (e.g., amp-ima-video.md) for it to be included in docs set.  I've commented on https://github.com/ampproject/amphtml/pull/8557.

honeybadgerdontcare:
@Gregable for ask around verifying documentation as a potential improvement over the current status quo.

I've also asked for a MD file on the original request and @aghassemi maybe that should be a requirement for new component PRs....

PR #484, opened 2017-05-12
Include amp-facebook-like component into docs

bpaduch:
Now that amp-facebook-live component is in production, include it in the docs. Related to: #483 
Also, adjusted side-nav so "1st" component listed is selected when user selects "Social"
Also, removed "amp-share-tracking" from docs because a) there are no docs for the component.


honeybadgerdontcare:
@bpaduch looks like I don't have write access to ampproject/docs, so you'll need another approver to merger.

Issue #481, opened 2017-05-11
Breadcrumb "Who is AMP for" is broken


bpaduch:
1. Go here: https://www.ampproject.org/learn/who-uses-amp/advertisers/
2. In the breadcrumbs, click "Who is AMP for".
3. Get a 404, it's looking for https://www.ampproject.org/learn/who-uses-amp/who-uses-amp/

bpaduch:
cc: @derekcecillewis

derekcecillewis:
Good catch! Should be taken care of with my PR.

SidVal:
I've got a doubt here, did not know if I could talk here or should I open an other issue.

Check this:

1. https://www.ampproject.org/es/support/developer/
2. Breadcrump: Support >> Developer Support
3. If you click on "Support" link, goes to: https://www.ampproject.org/support/ (in English).

How do I edit the file to redirect to: https://www.ampproject.org/es/support/ ?

bpaduch:
@SidVal - Localization breadcrumb fixes will be fixed in Derek's revised version of https://github.com/ampproject/docs/pull/486.

PR #485, opened 2017-05-12
Fix "Who is AMP for?" breadrcrumb path

derekcecillewis:
Fixes "Who is AMP for?" breadrcrumb path (#481) by adding correct path to `who-uses-amp.yaml`.


content/learn/who-uses-amp.yaml
```diff
@@ -3,6 +3,7 @@ $order: 1
 class: about-who
 $view: /views/about-who.html
 $parent: /content/learn/overview.yaml
+$path: /learn/who-uses-amp/
```
pbakaus: Curious - why doesn't the implicit path work here?

derekcecillewis: Grow uses collections to figure out paths. Since the collection's name (`learn/who`) does not match the path (`learn/who-uses-amp`), it is giving incorrect output. So the only options available are to either explicitly state the path as this PR does or change the name of the collection's folder from `who` to `who-uses-amp`. If you decide on the latter option, which might be preferable, it will break a few other places that the old collection path is referenced (`content/includes/menu.yaml`, `views/about-who.html`). Let me know what you decide, I'd be happy to rebase.

pbakaus: ah got it! all good.

bpaduch:
This PR has broken the [Master build](https://travis-ci.org/ampproject/docs/builds/232900266).  Can @derekcecillewis debug and correct it?

Error:
`grow.pods.documents.PathFormatError: No path format found for <Document(/content/latest-ads/list-blog-ads.html, locale='ru')>. You must specify a path format in either the blueprint or the document.`

derekcecillewis:
Yes, Barb. Paul informed me of this yesterday. Sorry for the delay, I've been traveling and my vacation started today. I'll take a look after breakfast.

bpaduch:
I commented on the wrong PR, it's PR #486 that broke the build.

PR #486, opened 2017-05-16
Add localization support for breadcrumbs and other minor improvements

derekcecillewis:
This PR adds localization support to the breadcrumbs as well as the following minor improvements:

- Add comments for improved readability
- Normalize variable names to improve others' ability to more easily identify the pattern
- Prefer `do` instead of `if` hack
- Prefer underscore naming convention
- Properly indent `set` blocks by using [whitespace control](http://jinja.pocoo.org/docs/dev/templates/#whitespace-control)


pbakaus:
LGTM, thanks!

PR #501, opened 2017-05-18
Add partial localization support for breadcrumbs and other minor improvements

derekcecillewis:
In addition to the changes mentioned in #486, this PR accounts for path errors as seen in #481, which were also causing the build to break with the previous rendition of this PR. This seems like a Grow bug to me, but I will have to investigate further to see why paths that are only defined in the blueprint are based on collection names. This PR builds without errors and I've tested it to ensure that it is working properly. The reason why several of the breadcrumbs are not translated is because their respective titles are not translated, which may be something worth working on.

### DIFF
--- a/scripts/update_blog_links.js
+++ b/scripts/update_blog_links.js
@@ -90,7 +90,9 @@ avatar: ${ item.avatar }
 date_data: ${ moment(item.date).format() }
 $date: ${ moment(item.date).format("MMMM D, YYYY") }
 $parent: ${parent}
-
+$path: /latest/blog/{base}/
+$localization:
+  path: /{locale}/latest/blog/{base}/
 components:
   - social-share
 ---
@@ -215,6 +217,7 @@ const fetchLatestAdsContent = fetchFeedAsync({
   // write the posts into their own files
   for (var post of posts) {
     writeBlogPage(post, '../content/latest/blog-ads/', '/content/latest-ads/list-blog-ads.html');
+    writeBlogPage(post, '../content/latest/blog-ads/', '/content/latest/list-blog.html');
   }
 
   // combine posts  and save into list-blog-ads.yml
--- a/views/partials/breadcrumb-nav.html
+++ b/views/partials/breadcrumb-nav.html
@@ -1,26 +1,37 @@
-{% set currentDocs = [] %}
-{% do currentDocs.append(g.doc(doc.pod_path)) %}
-{% set listItems = [] %}
-{% set currentDoc = currentDocs|last %}
-{% set firstListItem %}
-<li>{{currentDoc.titles('breadcrumb')}}</li>
-{% endset %}
-{% if listItems.append(firstListItem) %}{% endif %}
-{% if currentDocs.append(g.doc(currentDoc.pod_path).parent) %}{% endif %}
-{% if not currentDocs|last == None %}
-  {% for _ in range(10) %}
-    {% set currentDoc = currentDocs|last %}
-{% set nextListItem %}
-<li><a href="{{currentDoc.url.path}}">{{currentDoc.titles('breadcrumb')}}</a></li>
-{% endset %}
-    {% if listItems.append(nextListItem|safe) %}{% endif %}
-    {% set currentDoc = currentDocs|last %}
-    {% if currentDocs.append(g.doc(currentDoc.pod_path).parent) %}{% endif %}
-    {% if currentDocs|last == None %}{% break %}{% endif %}
+{#
+ # Breadcrumb Nav
+ #}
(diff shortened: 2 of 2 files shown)


bpaduch:
I've tested this PR locally, and the breadcrumbs function, and the build doesn't break, however the part about why the respective titles aren't translated isn't quite what I expected, that is, I expected the breadcrumbs to use translated content from the doc title (if present, see option1), or the variable (option2).  

There's a couple ways translations are done in AMP:

1) Separate translated files for each locale (e.g., `overview@es.yaml`, `overview@ko.yaml`, etc.)
2) Using translation variables, that is, variables postfixed with @ (e.g., `$title@`, `cta@`, etc.), and the translations for those extracted variables are in the locale's associated messages.po file.  For example the variable [`$title@: Who is AMP for?`](https://github.com/ampproject/docs/blob/master/content/learn/who-uses-amp.yaml#L1) has an associated [Spanish translation](https://github.com/ampproject/docs/blob/master/translations/es/LC_MESSAGES/messages.po#L1120).

So, in the breadcrumbs, if the page has a separate translated file, the breadcrumb pulls the title from the translated document, like in the case of the  [Korean page "Who is AMP for"](https://www.ampproject.org/ko/learn/who-uses-amp/). However, note that the first breadcrumb "What is AMP" is not translated even though that page is fully translated.

And in the case of [Spanish (ES), the "Who is AMP for" page](https://www.ampproject.org/es/learn/who-uses-amp/) and its content is translated (not via a separate file but through extracted translated variables).  

It appears that the breadcrumb for "What is AMP" is a translated variable: [` breadcrumb@: What is AMP`](https://github.com/ampproject/docs/blob/master/content/learn/overview.yaml#L3) but for some reason the breadcrumb view/template isn't picking up the translated variable's translations.

derekcecillewis:
Barb, you're absolutely right! In order for the translations done with extracted variables to take effect, the [`gettext` (`_`)](http://grow.io/docs/templates/#_) function needs to be applied to each breadcrumb. Unfortunately, this breaks certain pages. I haven't yet had the time to determine the cause of this, but I'll push the commit that incorporates this function for all to see. Please don't merge until we get this issue squared away.

pbakaus:
What's the status of this?

derekcecillewis:
Pending writing a reduction and submitting a bug to Grow repo.

derekcecillewis:
I've updated this PR to reflect the conversation we had on Slack regarding partial breadcrumb translation support until the bug is resolved. If merged, the current page's breadcrumb title will be translated as well as any as any ancestors that do not have their `title` key overridden by `titles.breadcrumb`. Pending approval from @bpaduch, it should be ready to merge.

bpaduch:
Performed before/after testing:

**Test 1: Page: ["AMP for Ad Tech Platforms"](https://www.ampproject.org/es/learn/who-uses-amp/ad-tech-platforms/)**

BEFORE
1) :x: Breadcrumbs appear as: What is AMP >> Para quien es AMP >> AMP for Ad Tech Platforms
2)  :white_check_mark: Page title is actually "AMP para Plataformas Tecnológicas de Anuncios"
3) :x: When I click the ancestor "Para quiens es AMP", get a 404 - https://www.ampproject.org/es/learn/who-uses-amp/who-uses-amp/
4) :x: When I click "What is AMP", takes you to https://www.ampproject.org/es/learn/overview/, where content is ES, but breadcrumb is "What is AMP".

AFTER 
1) :white_check_mark: +  :question: What is AMP >> Para quien es AMP >> AMP para Plataformas Tecnológicas de Anuncios. Better, but expected all breadcrumbs to be ES.
2) :white_check_mark: Page title remains correct as "AMP para Plataformas Tecnológicas de Anuncios"
3) :white_check_mark: When I click the ancestor "Para quiens es AMP", I land on the correct page.
4)  :white_check_mark: When I click "What is AMP", takes you to https://www.ampproject.org/es/learn/overview/, where content is ES, and breadcrumb is "Que is AMP"

**Test 2: Page ["FAQs"](https://www.ampproject.org/es/support/faqs/)**
- Similar results as Test 1

**Summary**
This PR is an improvement, and I'll approve.  However, is it possible to dig into the remaining issue -- if there are children in the breadcrumb, the primary ancestor (the 1st breadcrumb) appears in English, no matter the selected language.  Even if it's a `breadcrumb` variable, like in [overview.yaml](https://github.com/ampproject/docs/blob/master/content/learn/overview.yaml#L3) or not, like in [support.md](https://raw.githubusercontent.com/ampproject/docs/master/content/support/support.md).

derekcecillewis:
Sure, would you mind creating a separate issue for it please?

PR #505, opened 2017-05-18
Update advertisers.yaml: Fixed grammar for 'its'

coryphwang:
Updated 'it's' to 'its' for "Even the most memorable creative won't serve its purpose if..."


bpaduch:
Thanks for the fix!

PR #512, opened 2017-05-24
Removed extra 'hotels' in the first paragraph

coryphwang:
This is also a problem in the attached PDF: https://www.ampproject.org/static/img/case-studies/wego.pdf


bpaduch:
@mattludwig - Can you provide an updated PDF for Wego? I'll approve this current PR in the actual text.

PR #515, opened 2017-05-24
Create who-uses-amp@es.yaml

SidVal:
More :es:


bpaduch:
Sid, I've corrected some things in the file, and logged two bugs in relation to problems with translations of this file.

Issues logged: #520, #519

PR #524, opened 2017-05-26
Incorporate `gettext` function for extracted title translations

derekcecillewis:
This PR applies the `gettext` function to page titles. The translations have already been done, this just activates them on non-default locales. This also prepares #501 to utilize an alternative method for breadcrumb translations.


bpaduch:
Tested locally and it's picking up the translated text.

PR #528, opened 2017-05-29
Correct vendor sort order

bpaduch:
Removed "sort" from page view because content for vendors comes in already sorted but was displaying in an unsorted way.  It should be listing them alphabetically.  This does that.


pbakaus:
Without explicit sorting, it sometimes breaks based on the OS. I'll merge, but let's double check that it works on staging/prod.

bpaduch:
Viewed staging on: Chrome on OS, Chrome on Linux, Chrome on Android, Firefox on Linux - all appear correctly.

PR #582, opened 2017-06-27
Update responsive_design.md

SidVal:
Better understanding for the css class.
Related: https://github.com/ampproject/docs/issues/581#issuecomment-311453446


bpaduch:
Rather than linking out to that file, I incorporated it into the doc, see my response in https://github.com/ampproject/docs/issues/581#issuecomment-311465053.  If you approve, I'll integrate my PR.

SidVal:
Fixed with the PR #583

PR #603, opened 2017-07-07
Translate menu items into Japanese

oikwsat:
Dear AMP Project.

I've complete translate menu items into Japanese.
Please take a look.

Thanks.

### DIFF
--- a/translations/ja/LC_MESSAGES/messages.po
+++ b/translations/ja/LC_MESSAGES/messages.po
@@ -83,7 +83,7 @@ msgstr ""
 
 #: /content/docs/reference/spec/_blueprint.yaml
 msgid "AMP HTML Specification"
-msgstr ""
+msgstr "AMP HTML仕様"
 
 #: /views/amp-ads.html:252
 msgid "AMP Landing Page"
@@ -112,15 +112,15 @@ msgstr ""
 
 #: /content/learn/who/ad-tech-platforms.yaml
 msgid "AMP for Ad Tech Platforms"
-msgstr ""
+msgstr "広告プラットフォーム向けのAMP"
 
 #: /content/learn/who/advertisers.yaml
 msgid "AMP for Advertisers"
-msgstr ""
+msgstr "広告主向けのAMP"
 
 #: /content/learn/who/publishers.yaml
 msgid "AMP for Publishers"
-msgstr ""
+msgstr "サイト運営者向けのAMP"
 
 #: /content/includes/home.yaml
 msgid ""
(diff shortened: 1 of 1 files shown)


translations/ja/LC_MESSAGES/messages.po
```diff
@@ -654,7 +654,7 @@ msgstr ""
 #: /content/includes/menu.yaml /content/pages/home.html:37
 #: /content/pages/home.html:77 /views/amp-ads.html:409 /views/amp-ads.html:444
 msgid "News"
-msgstr ""
+msgstr "ブログ"
```
bpaduch: Might this be better: ニュース  (via Google Translate)

translations/ja/LC_MESSAGES/messages.po
```diff
@@ -161,7 +161,7 @@ msgstr ""
 
 #: /content/includes/menu.yaml
 msgid "About"
-msgstr ""
+msgstr "AMP とは"
```
bpaduch: This seems to translate to "What is AMP".  Is there a better translation for just "About"?

techhtml:
I'll review this translations.

bpaduch:
Thanks @techhtml!

translations/ja/LC_MESSAGES/messages.po
techhtml: This translation are not bad translation. because `news` menu will move in `/blog/`.

translations/ja/LC_MESSAGES/messages.po
techhtml: Korean translation are also use different translation for batter word.
For example, if translate `about` to directly in Korean that is `대하여`, but it's too difficult to know about what for that menu.
So i'll add nouns so that is `AMP에 대하여`, Japanese also same for this problem.

But I think you can change this one to `AMPについて`.

techhtml:
Thank you for your contribution @oikwsat.
I think your translation is good, but change some sentence maybe it will more batter word.
PTAL.

Thanks.

translations/ja/LC_MESSAGES/messages.po
oikwsat: > news menu will move in /blog/.

Yes, the reason why I chose `ブログ` is this.
In Japanese,  News is `ニュース`, Blog is `ブログ `.
I wonder if I should chose `ニュース` or `ブログ`?
Okay, let's fix to `ニュース`. If you change English, I will also change Japanese too.

translations/ja/LC_MESSAGES/messages.po
oikwsat: Thank you for comments.

> But I think you can change this one to AMPについて.

It's better!  I will do.

Issue #618, opened 2017-07-14
Specify placeholder/fallback interaction


choumx:
```
1.  Show placeholder till load
2.  If load success, hide placeholder and display content
3.  Else if load failure,
    1.  If fallback, display fallback
    2.  Else if placeholder, keep placeholder
```

Context: https://github.com/ampproject/amphtml/pull/10281#discussion_r125968015

PR #626, opened 2017-07-18
Add interaction behav for placeholders & fallbacks

bpaduch:
Related to #618 - Added info on interaction of placeholders & fallbacks that will be common text for amp-twitter and amp-list.  Per https://github.com/ampproject/amphtml/pull/10488#issuecomment-316133089.

### DIFF
--- a/content/docs/guides/responsive_amp/placeholders.md
+++ b/content/docs/guides/responsive_amp/placeholders.md
@@ -72,7 +72,7 @@ For AMP components that rely on dynamic content (e.g., `amp-twitter`, `amp-list`
   <li>If the content fails to load:
     <ol>
       <li>If there's a fallback element, display the fallback.</li>
-      <li>Otherwise, if there's a placeholder, display the placeholder.</li>
+      <li>Otherwise, continue displaying the placeholder.</li>
     </ol>
   </li>
 </ol>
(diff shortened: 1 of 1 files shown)


content/docs/guides/responsive_amp/placeholders.md
```diff
@@ -62,6 +62,21 @@ that the browser doesn’t support a particular feature:
...
+
+For AMP components that rely on dynamic content (e.g., `amp-twitter`, `amp-list`), the interaction of fallbacks and placeholders operates as follows:
+ 
+<ol>
+  <li>Display the placeholder while the content is loading.</li>
+  <li>If the content loads successfully, hide the placeholder and display the content.</li>
+  <li>If the content fails to load:
+    <ol>
+      <li>If there's a fallback element, display the fallback.</li>
+      <li>Otherwise, if there's a placeholder, display the placeholder.</li>
```
choumx: How about "Otherwise, continue displaying the placeholder."

We don't re-display the placeholder in this case, we just keep showing it.

bpaduch: Only, if placeholder is present, right?  If there's no placeholder, no fallback, and the content loads, display nothing. So, "Otherwise, continue displaying the placeholder (if present)."

PR #641, opened 2017-07-31
Update messages.po

MattiaDes:


translations/it/LC_MESSAGES/messages.po
```diff
@@ -26,7 +26,7 @@ msgstr "Un percorso verso il successo"
 
 #: /views/amp-ads.html:249
 msgid "AMP Ad"
-msgstr ""
+msgstr "AMP Ad"
```
bpaduch: Just to confirm, is that the Italian translation for "AMP Ad"? If it's not, please don't fill in it.  By default, if there's no string, the English is used.

translations/it/LC_MESSAGES/messages.po
```diff
@@ -92,7 +110,7 @@ msgstr "Specifiche HTML AMP"
 
 #: /views/amp-ads.html:252
 msgid "AMP Landing Page"
-msgstr ""
+msgstr "AMP Landing Page"
```
bpaduch: Please insert Italian translation instead of English translation.

MattiaDes:
Ok done

PR #648, opened 2017-08-03
Add icon and CSS for external links in ref docs

bpaduch:
To make it apparent that links are "external" to the site, adding CSS and icon that will be used in ref docs, like in the following example:

external-links-ref-docs

Now, I'll just need to change all the ref docs that have examples, and add class="external" + consistent wording.

cc: @mattludwig

### DIFF
--- a/assets/sass/_content-post.scss
+++ b/assets/sass/_content-post.scss
@@ -17,7 +17,7 @@
     text-transform: none;
   }
 
-  a.external::after {
+  a[href*="//"]:not([href*="www.ampproject.org"]):not([href*="github.com/ampproject"])::after {
     content: "";
     display: inline-block;
     background-image: url('/static/img/ic_open_in_new_blue.svg');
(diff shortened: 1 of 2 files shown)


assets/sass/_content-post.scss
```diff
@@ -17,6 +17,17 @@
     text-transform: none;
   }
 
+  a.external::after {
```
pbakaus: Much nicer way of doing this that doesn't require an additional class: https://davidwalsh.name/external-links-css

bpaduch: So, you'd like to see this for *every* external link (even the ones to github  docs)?  I was thinking of being selective about it, say for only ABE and AMP Start.

bpaduch: Oooo, nevermind, just doing two `:not`s.  This was cool.  Thanks for the tip and for the learning experience!

pbakaus: yep, great! :)

Issue #650, opened 2017-08-05
amp-bind page - table is not readable on 13" mac pro


camelburrito:
Look at 
https://www.ampproject.org/docs/reference/components/amp-bind#white-listed-functions
<img width="1201" alt="screen shot 2017-08-05 at 1 01 17 pm" src="https://user-images.githubusercontent.com/1037311/28998383-77829bec-79de-11e7-80f0-6449a82e14cc.png">

bpaduch:
@pbakaus - I can't figure out a "great" solution.  I worked with this for a couple hours, and it's the `<pre>` tags that are causing the grief. I came up with:

1. Remove the `[toc]`
2. Add the class `"nowrap"` to the particular troublesome table.
3. Add the following  css to `_tables.scss`:

```css
// Wrap the pre code
td > pre {
  white-space: pre-line;
}

// don't wrap code elements
table.nowrap {
  td > code {
    white-space: nowrap;
  }
}
```

Thoughts?

Oh, and there's a `rowspan` tag that was a straggler that s/b removed/corrected.

PR #652, opened 2017-08-09
Fix table rendering for narrow widths

bpaduch:
Fixes: #650 

For [amp-bind](https://www.ampproject.org/docs/reference/components/amp-bind#white-listed-functions), the tables contain `<pre>` blocks which affect the readability of content in the other columns.  

- I've changed the TOC breakpoint so that the TOC renders at the top instead of the side for 1350 or less. 
- For TOC breakpoint width, `<pre>` codeblocks in tables are wrapped, and I've reduced the padding.

Thoroughly tested locally, reviewing all component docs.


pbakaus:
sorry I've been swamped. Will take a look at a later date to see if there's a more universal solution, but this seems good for now.

bpaduch:
np, thanks.

PR #703, opened 2017-09-25
Add clarifications to "Add advanced AMP features" codelab

mrjoro:
I made some clarifications for things I found a little bit confusing when walking through this codelab.

/cc @pbakaus


mrjoro:
When walking through the codelab, my page also ended up looking different at the end than the screenshots shown (starting from the sidebar/home link).

I tracked this down to the final screenshots showing the <article> tag from the initial document just below the <header>, but the instructions says to add the first component we add (amp-ad) just under the <header>.  This puts it above the <article>, which is what the screenshot on that step suggests.

Some other steps don't specify where to put the new elements; we could just be consistent and always say "put these just under <header>" or "put these just under <article>" and then adjust all screenshots to match.

WDYT?

PR #706, opened 2017-09-28
Importing support.md file and adding to dev support section

bpaduch:
Importing [AMPHTML's support.md](https://github.com/ampproject/amphtml/blob/master/SUPPORT.md) file into ampproject.org docs.  Adding the doc to the "Developer Support" section.

cc: @rudygalfi, @cramforce 

getting-support

Note: Needed to amend format in support.md - see https://github.com/ampproject/amphtml/pull/11474

### DIFF
--- a/assets/sass/_content-section.scss
+++ b/assets/sass/_content-section.scss
@@ -13,6 +13,13 @@
   }
 }
 
+.md {
+  ul,
+  ol {
+      margin-left: 40px;
+  }
+}
+
 @include max-screen($mobile-breakpoint) {
   .post-title {
     text-align: left;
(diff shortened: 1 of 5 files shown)


assets/sass/_content-section.scss
```diff
@@ -13,6 +13,13 @@
   }
 }
 
+.md {
+  ul,
+  ol {
+      margin-left: 40px;
```
pbakaus: this seems like a change that could affect many other pages, no? If not, ignore my comment :)

bpaduch: Thanks for the check.  I triple checked all the views/pages that use class "md" and "ul,ol" and they have their own styles, that aren't affected by this change.

PR #721, opened 2017-10-16
Added Kentico as an AMP-enabled platform

petrsvihlik:
As per https://github.com/ampproject/docs/issues/686 I've added Kentico as a supported platform (CMS).


petrsvihlik:
I signed it!

PR #724, opened 2017-10-18
Fix tons of blog issues by using WP converted AMP

pbakaus:
This fixes tons of layout issues and AMP errors by making the importer that imports blog posts use the AMP-converted versions from WordPress, instead of converting the RSS source.


bpaduch:
Fixes: https://github.com/ampproject/docs/issues/523

PR #729, opened 2017-10-20
Remove hamburger menu random gray box

morsssss:
Remove box-shadow so that mysterious gray box won't appear on hamburger menu hover.

I don't believe this rule applies anywhere else on the site.  I did look around!

Fixes: https://github.com/ampproject/docs/issues/472

### DIFF
--- a/assets/sass/_header.scss
+++ b/assets/sass/_header.scss
@@ -128,11 +128,12 @@
     }
   }
 
-  .tab {
+  div.tab {
     padding: 6px 20px;
 
     &::before {
       bottom: 0;
+      box-shadow: $header-shadow;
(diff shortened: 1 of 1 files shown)


bpaduch:
@morsssss - I tested this and it turns out that box-shadow is used for tabs, which are used in desktop.  By removing it, tabs are not easily distinguishable: 

drop-shadow

morsssss:
I'm glad you tested and found that!  I'm sure I can scope the rule less broadly. Will take a look.

morsssss:
Think I did it the right way this time.  Please review!

bpaduch:
Nice!

PR #734, opened 2017-11-02
Added translations for ID, JA, PT_BR, ZH_CN for supported-browsers.md

coryphwang:
Continuation of https://github.com/ampproject/docs/pull/719#pullrequestreview-69052258

Made changes to the id, ja, pt_BR, zh_CN translations to change the file names from "supported-browser" to "supported_browsers" (with an added "s" at the end).

### DIFF
--- a/content/support/faqs/supported-browsers@id.md
+++ b/content/support/faqs/supported-browsers@id.md
@@ -1,0 +1,31 @@
+---
+$title@: Browser yang Didukung
+$order: 4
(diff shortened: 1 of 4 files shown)


bpaduch:
Shouldn't this pull request also include the deletions of the non-s ending files?  When you do a rename, typically there's an add and delete.

coryphwang:
I'm going to close this pull request, and do another one that deletes the old files, and adds the new files with the updated file names.

PR #749, opened 2017-11-15
Fix roadmap status of client-side sorting and filtering extensions

dereklewis:
This PR reflects the actual status of the progress made. It seems as though these components were not bumped forward on the roadmap update.


dereklewis:
I signed it!

PR #767, opened 2017-11-30
add instructions to import amp-fit-text

calebcordry:
We are not currently explicitly telling the tutorial user to import this component. 

I am not sure if we are assuming they know how to do this on their own by this point in the tutorial. If not, we need something like this added to the docs, or included in the initial repo.


bpaduch:
Turns out it's actually the order of the topics that were the issue.  The "Adding Extended AMP Components" topic had amp-fit-text added (along with the script) but when I added the tutorial I didn't put the topics in the correct order.  Will be fixing that with https://github.com/ampproject/docs/pull/768.

calebcordry:
Perfect. Thanks.

PR #768, opened 2017-12-01
Fix order of topics in tutorial

bpaduch:
Thanks to @calebcordry - amp-fit-text script should be added before component used; turns out the topics weren't in correct order, that is, "Adding carousels" should come *after* "Adding extended components".


calebcordry:
Would you also want to change the order in the sidebar?

image

Let me know if you want me to figure out how to do this. Happy to do so.

bpaduch:
Yep, that will happen when merged to Production. :)

PR #783, opened 2017-12-13
Added translations for es, id, ja, ko, pt_BR, zh_CN

bt-amp:
6 translations for "Common Attributes" [es, id, ja, ko, pt_BR, zh_CN]


bpaduch:
In the future, please remember to use `pt_br` vs `pt_BR` and `zh_cn` vs `zh_CN` - otherwise you receive 404s. 

cc: @coryphwang

PR #810, opened 2018-01-19
Adding configuration for the new code highlight to use classes.

### LINKED ISSUE #809: AMP pages with codeblocks are invalid
See https://www.ampproject.org/docs/getting-started/, which was previously valid but now has 205 AMP errors due to inline styles!

I suspect this is related to some changes in Grow's styling.

cc: @Zoramite, @rsimha-amp 

grow-inline-styles-issue

Zoramite:
Fixes #809

### DIFF
--- a/assets/sass/_syntax_highlighting.scss
+++ b/assets/sass/_syntax_highlighting.scss
@@ -87,7 +87,7 @@ pre {
   }
 }
 
-.highlight {
+.code {
   .hll { background-color: #ffffcc }
   .c { color: #0099FF; font-style: italic } /* Comment */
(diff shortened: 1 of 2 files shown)


bpaduch:
@Zoramite - I should probably switch Grow back to 4.1 to see if this change works, right?

Zoramite:
@bpaduch yes, switching back to 4.1 will make use of the updated config and fix the code highlight coloring.

bpaduch:
Per https://github.com/ampproject/docs/pull/810#issuecomment-359041524

PR #823, opened 2018-01-30
Opt in url-replacement-v2 experiment

lannka:
@pbakaus

/cc @calebcordry


lannka:
Thanks @bpaduch !

Would you be able to push the site, so the change can be picked up?

bpaduch:
I did.  I kicked off production build right after master build.

PR #885, opened 2018-03-20
Update amp_story_best_practices.md

ithinkihaveacat:
If auto advancing on video completion, use video id rather than expected video length.

/cc @newmuis


newmuis:
+1000, great to have this recommendation

PR #893, opened 2018-03-26
Update analytics vendor copy for New Relic.

blainesch:
This is the updated copy from the product manager of the Browser team, which is responsible for the AMP feature.

Thank you!


blainesch:
I signed it!

PR #898, opened 2018-03-27
Add instruction in tutorial to add `amp-ad` script

nonproftechie:
Because using `amp-ad` will throw an error soon if it is not imported as a custom component, I'm suggesting an early update to the tutorial.

Fixes #898


nonproftechie:
(Signed CLA through my gmail: nesher181@gmail.com)

PR #902, opened 2018-03-29
Add text about valueless attributes workaround

SidVal:
Ref: https://github.com/ampproject/docs/commit/07f3d83f675f0e9297da5291b95e46733bb7d600


bpaduch:
Thanks!

PR #911, opened 2018-04-05
Add exception case for click tracking

jasti:
Add cautious note about not using amp-analytics for click tracking for AMPHTML ads


bpaduch:
LGMT

PR #972, opened 2018-05-30
Update and lock down package versions used by `ampproject/docs`

### LINKED ISSUE #894: Build infra: Lock dep versions and keep up to date automagically
We should lock versions of dependencies to make our builds less fragile and our caching in Travis more dependable, and find a way to semi-automatically/painlessly keep them updated.

rsimha:
This PR does the following:
- Locks down the versions in `package.json` so that upgrades to dependencies cannot break the build
- Updates `yarn.lock` with the result of running `yarn`
    - Since caching is already enabled for `yarn` on Travis, the install step during Travis builds will be much faster
- Adds a new script `build_scripts/check_package_manager.js`, which is run prior to the installation of packages.
    - Detects and prints the versions of `node`, `yarn`, and `grow`
    - Suggests upgrades if `node` is not LTS, or if `yarn` is not stable, or if `grow` is not latest
    - Makes sure that `yarn` is being used to add / install packages during local development
    - Fails the install and prints a useful message if `npm` is run instead of `yarn`

Fixes #894

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,7 +34,7 @@ jobs:
         - ./build_scripts/build.sh tr zh_CN ru
     - stage: deploy
       install:
-        - npm install -g firebase-tools
+        - yarn global add firebase-tools
       script:
         - ./build_scripts/deploy.sh
 branches:
--- a/build_scripts/check-package-manager.js
+++ b/build_scripts/check-package-manager.js
@@ -1,0 +1,218 @@
+/**
+ * Copyright 2018 The AMP HTML Authors. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS-IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+'use strict';
+
+const childProcess = require('child_process');
+const https = require('https');
+
+const setupInstructionsUrl = 'https://github.com/ampproject/amphtml/blob/master/contributing/getting-started-quick.md#one-time-setup';
+const nodeDistributionsUrl = 'https://nodejs.org/dist/index.json';
+const growReleasesUrl = 'https://api.github.com/repos/grow/grow/releases';
+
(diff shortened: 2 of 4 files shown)


bpaduch:
Are we able to lock down "Grow" version and its dependencies?

bpaduch:
Does this include locking down "Grow" version and its dependencies?

rsimha:
I'm still trying to figure out a way to reliably install a particular version of `grow`. Once I figure it out, I'll send out a separate PR. Meanwhile, it's good to expose version info whenever someone does an install.

bpaduch:
Would like @pbakaus to review too.

rsimha:
This PR now checks that the version of `grow` matches the latest stable version according to https://api.github.com/repos/grow/grow/releases

rsimha:
And here's what a package update with old versions will look like:
```
~/src/docs$ yarn
yarn install v1.3.2
$ node build_scripts/check-package-manager.js
WARNING: Detected node version v8.9.4. Recommended (latest LTS) version is v8.11.2.
⤷ To fix this, run "nvm install --lts" or see https://nodejs.org/en/download/package-manager for instructions.
Detected grow version 0.5.0 (latest).
WARNING: Detected yarn version 1.3.2. Recommended (stable) version is 1.7.0.
⤷ To fix this, run "curl -o- -L https://yarnpkg.com/install.sh | bash" or see https://yarnpkg.com/docs/install for instructions.
Attempting to install packages...
[1/4] Resolving packages...
success Already up-to-date.
Done in 3.77s.
```

rsimha:
@pbakaus @bpaduch Could I bother you folks to check out this branch and run `yarn`? I'd expect that you'll see version warnings, but they should be self explanatory.

Let me know how it goes.

bpaduch:
Got this

```
yarn install v1.6.0
$ node build_scripts/check-package-manager.js
WARNING: Detected node version v9.9.0. Recommended (latest LTS) version is v8.11.2.
⤷ To fix this, run "nvm install --lts" or see https://nodejs.org/en/download/package-manager for instructions.
Detected grow version 0.5.0 (latest).
WARNING: Detected yarn version 1.6.0. Recommended (stable) version is 1.7.0.
⤷ To fix this, run "curl -o- -L https://yarnpkg.com/install.sh | bash" or see https://yarnpkg.com/docs/install for instructions.
Attempting to install packages...
[1/4] Resolving packages...
[2/4] Fetching packages...
info fsevents@1.2.4: The platform "linux" is incompatible with this module.
info "fsevents@1.2.4" is an optional dependency and failed compatibility check. Excluding it from installation.
[3/4] Linking dependencies...
[4/4] Building fresh packages...
Done in 9.21s.
```

rsimha:
Yep, that's exactly what you should see.

PR #996, opened 2018-06-14
New doc demonstrating layouts

bpaduch:
I finally got some time for this - a page that visually demonstrates the layouts in AMP.
Demo page on staging at: https://ampproject-custom.firebaseapp.com/docs/design/amp-html-layout/layouts_demonstrated


assets/sass/_content-post.scss
```diff
@@ -83,6 +83,62 @@
       border: none;
     }
   }
+
+  /* styles for layouts demonstrated doc */
+  .main-layouts {
```
pbakaus: can we move this into another custom file? Loading this on dozens of pages might be excessive.

bpaduch: Ah, good point.  I'll work on that.

content/docs/design/amp-html-layout/layouts_demonstrated.md
```diff
@@ -0,0 +1,138 @@
...
+              </amp-img>
+            </div>
+          </div>
+        </div>
+      </div>
+
+      <!-- flex-item -->
+      <div class="layout-item">
+        <h3>flex-item</h3> 
+        <div class="description">With a <code>fixed-height</code> layout, the element and other elements in its parent take the parent container's remaining space when the parent is a flexible container (i.e., display: <code>flex</code>).</div>
```
sebastianbenz: s/fixed-height/flex-item

I don't quite understand the flex-item behaviour from the diagram and the explanation. I always thought it just makes the element behave as defined by the flex container (which usually means resizing all elements together).

bpaduch: Thanks for the comments.

Based on the AMP's team's own [sample](https://github.com/ampproject/amphtml/blob/master/examples/layout-flex-item.amp.html), it doesn't look like the child items resize to "fit" the container.  Am I misunderstanding?

bpaduch:
@pbakaus - I moved the styles into a custom file, but the styles still need to be pulled in via the docs stylesheet.  Is there a way to include only that custom .scss file for one page without having to make a separate view file (overboard imho).

bpaduch:
@aghassemi, @jpettitt  - would be good to have your eyes on this so I am interpreting layouts correctly.

content/docs/design/amp-html-layout/layouts_demonstrated.md
sebastianbenz: You're right, it's just what I mostly used flex layout for. 

Suggestion: "With a <code>flex-item</code> layout, the element will take the available space based on the flexbox model."

via https://github.com/ampproject/amphtml/issues/2778

bpaduch: Good wording! I've rewritten.

PR #1019, opened 2018-07-03
Allow inline styles

choumx:
Inline styles are now allowed in the AMP validator. They share the same 50KB limit as `<style amp-custom>`. 

We probably should also add some language to https://www.ampproject.org/docs/design/responsive_amp#add-styles-to-a-page.

I'm not sure if there are other places in the docs that reference this. Please update those too! 

/to @bpaduch

### DIFF
--- a/content/docs/design/responsive_amp/style_pages.md
+++ b/content/docs/design/responsive_amp/style_pages.md
@@ -10,7 +10,7 @@ but you can’t reference external stylesheets
 (with the exception of [custom fonts](#the-custom-fonts-exception)).
 Also certain styles are disallowed due to performance implications.
 
-All styles must live in the head of the document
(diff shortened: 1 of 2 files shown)


pbakaus:
Should we tweet about this?

choumx:
I think it's tweet-worthy. 😄 /cc @honeybadgerdontcare

honeybadgerdontcare:
We'll need to also update https://www.ampproject.org/docs/guides/author-develop/responsive/style_pages. That is where the error message will direct those who go over 50kb or 1kb.

choumx:
@pbakaus There's some internal cleanup that should land in a couple weeks. Might be worth holding off until then.

bpaduch:
@CrystalFaith - can you review existing docs for reference to styles?

PR #1036, opened 2018-07-24
Clarify that AMP JS doesn't need to be second tag in <head>

fstanis:


cramforce:
This is correct. But having it early is best practice.

Issue #1027, opened 2018-07-18
Clean up docs to include ability for inline styles (not restricted anymore)


bpaduch:
In reference to https://github.com/ampproject/docs/pull/1019 and https://github.com/ampproject/amphtml/issues/14072, make sure references to styling isn't exclusive to just using the single inline stylesheet, now users can use inline styles.

Docs to check:

- Tutorials
- Style guides

CrystalFaith:
Also added changes in [pull request #17226](https://github.com/ampproject/amphtml/pull/17226) for documentation in amphtml. I wasn't able to assign a reviewer.

PR #1067, opened 2018-07-31
Updated documents to include inline styling 

CrystalFaith:
AMP now allows inline style attributes in HTML. This request adds this as an option to documentation. 
Contributes to issue #1027

### DIFF
--- a/content/docs/design/responsive_amp.md
+++ b/content/docs/design/responsive_amp.md
@@ -22,7 +22,7 @@ It is super easy to make elements responsive in AMP. Just put `layout="responsiv
 
 ## Add styles to a page
 
-Add all CSS inside a `<style amp-custom>` tag in the head of the document.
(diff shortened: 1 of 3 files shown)


bpaduch:
@CrystalFaith - does this fix #1027?  If so, make sure you put in your PR description "fixes #<issue#>"

content/docs/design/responsive_amp.md
```diff
@@ -22,7 +22,7 @@ It is super easy to make elements responsive in AMP. Just put `layout="responsiv
 
 ## Add styles to a page
 
-Add all CSS inside a `<style amp-custom>` tag in the head of the document.
+Migrate CSS inside a `<style amp-custom>` tag in the head of the document.
```
bpaduch: The word "migrate" assumes they are converting a page, which is always the case.  How about jus "Add your"?

CrystalFaith: Ah, good point - done.

PR #1069, opened 2018-08-01
Improve KO translation for "reference".

captain-lee:
The current translation "참조" can be used as reference, but "참고" is
better. "참조" means a reference for comparison, while "참고" means just
a reference.

A useful example for "참조" is like this. Refer the original document
(to compare). As for "참고", refer the chapter 14 for more detail. You
see the difference.


bpaduch:
@techhtml - can you confirm?
fyi: @coryphwang

PR #1073, opened 2018-08-03
Create benefits@es.yaml

SidVal:


bpaduch:
Hi SidVal, thanks for your contribution.  Instead of creating a separate .yaml file, I would encourage you to instead add the translated terms for "benefits.yaml" in the [messages.po](https://github.com/ampproject/docs/blob/master/translations/es/LC_MESSAGES/messages.po#L276) file.

bpaduch:
cc: @coryphwang

SidVal:
mmmm Why? I can't understand the criteria.
I mean, others files slit with language file, and this file not? :thinking:

bpaduch:
It's tricky but for a few reasons:

1) For the "benefits", we already have some of the text translated in the messages.po file, so it's just a matter of translating the rest.
2) It's a small .yaml file and messages.po are easier to maintain for our translation process (we are in the process of correcting existing .yaml files so they are just in the messages.po file).  For *large* .yaml files, we'll maintain separate files per locale.
3) For other locales, we are using the messages.po file, and we are actively translating them now.  As the localization team is already doing this, it would create more work for them to create separate .yaml files.

If you'd like, I can take what you have in your .yaml file, and move the text to messages.po file.  Let me know.

SidVal:
Now @bpaduch  ?

bpaduch:
Thanks @SidVal!  Looks great.

PR #1088, opened 2018-08-10
Improve KO translation for a menu text.

captain-lee:
The menu text for "Advertising" is translated to "광고하는" in KO, which doesn't make sense without an additional noun. The current form is an adjective that expects a noun at the end which is missing. So it looks incomplete. 
screenshot

Instead, it should be good to make it as "광고" as a noun. Yes, it's the same meaning with "advertisement", but in this case, it's better and it works.


captain-lee:
@techhtml, can you review this?

Issue #1117, opened 2018-08-23
Links in byline are displayed as HTML


cramforce:
See https://www.ampproject.org/latest/blog/privacy-and-user-choice-in-amps-software-architecture/

<img width="521" alt="screen shot 2018-08-23 at 1 53 47 pm" src="https://user-images.githubusercontent.com/89679/44551419-f8fa4500-a6db-11e8-8c84-02a850b8941e.png">

PR #1120, opened 2018-08-23
removed html from byline

CrystalFaith:
Resolves issue  [#1117](https://github.com/ampproject/docs/issues/1117)


CrystalFaith:
Should be better!

PR #1131, opened 2018-08-29
Adding translation for validate@es.md

rachnakumar:
Adding translation for validate@es.md


rachnakumar:
This is a revision file for validate@es.md

PR #1136, opened 2018-09-06
Added es, id, ko, pt_BR and zh_CN translations for validate.md

rachnakumar:
Added es, id, ko, pt_BR and zh_CN translations for validate.md


CrystalFaith:
@techhtml - Would you mind taking a look at the KO files? :)

PR #1182, opened 2018-09-21
Add KO translations for the entries in the KO glossary.

captain-lee:
Note that I chose to give multiple translations for a term. I acknowledge there can be slightly different translations depending on the context. (this glossary is a guide for better translation and consistency, but it is not a law).


captain-lee:
@techhtml Please review this PR.

FYI, @JeongminFelixShin, @janghanboram (please feel free to comment if you want).

PR #1204, opened 2018-09-27
Updated messages.po for JA translation

coryphwang:


CrystalFaith:
@coryphwang - it was just extra space in the front of the string - easy fix, but scary line!

PR #1213, opened 2018-10-01
Complete translation demostrating AMP layouts into Korean

techhtml:
Complete translation demostrating AMP layouts into Korean
Please Take a Look


CrystalFaith:
Awesome! Thank you!

PR #1231, opened 2018-10-11
added light styling to first demo of amp-list

CrystalFaith:
Part 1 of 2 resolves [issue #18619](https://github.com/ampproject/amphtml/issues/18619)

[Part 2](https://github.com/ampproject/amphtml/pull/18693)


examples/src/amplist.html
```diff
@@ -34,11 +34,15 @@
...
+    amp-list div[role="list"] {
+      display: grid;
+      grid-gap: 0.5em;
+    } 
   </style>
 </head>
 <body>
   <!-- basic -->
-  <amp-list width="auto" height="100" layout="fixed-height"
+  <amp-list id="testclass" width="auto" height="100" layout="fixed-height"
```
pbakaus: why the id? also, why does it say "testclass" when it is an id,not a class ;)

CrystalFaith: lol, for reasons that were real... thanks for the catch!

PR #1235, opened 2018-10-12
removed amp recaptcha

CrystalFaith:
amp-recaptcha-input is not ready and has no documentation - should not be imported yet.

### DIFF
--- a/scripts/component_categories.json
+++ b/scripts/component_categories.json
@@ -24,7 +24,6 @@
     "amp-live-list": "dynamic-content",
     "amp-mustache": "dynamic-content",
     "amp-next-page": "dynamic-content",
-    "amp-recaptcha-input": "dynamic-content",
     "amp-selector": "dynamic-content",
     "amp-subscriptions": "dynamic-content",
     "amp-subscriptions-google": "dynamic-content",


pbakaus:
This is OK for now, but the real problem here is that that doc file should probably never have landed in the amphtml repo in that state (cc @cramforce @rsimha).

Malte/Raghu: Is there a tag or branch that is safer to point at to pull docs in? Seems like master is too dangerous. Unfortunately though, I think we probably have production releases with empty docs right now as well..

rsimha:
I wonder if we can look at https://github.com/ampproject/amphtml/releases and use the SHA of the `latest release` instead of `HEAD`. This way, we're still working off of `master`, but won't have to include code that hasn't been released yet.

As an aside, @erwinmombay is working on re-enabling our GH hook that adds labels like "In canary" and "In production" to PRs based on their release status. I wonder if this can be useful here.

rsimha:
Specifically, https://github.com/ampproject/amphtml/releases/latest will redirect to the latest AMP release each week.

pbakaus:
...this was nagging me enough that I just implemented it, PTAL ;)

PR #1244, opened 2018-10-15
Pin imported docs to a current AMP release

pbakaus:
As per discussion in https://github.com/ampproject/docs/pull/1235.

### DIFF
--- a/scripts/import_docs.js
+++ b/scripts/import_docs.js
@@ -3,6 +3,7 @@
 const github = require('octonode');
 const fs = require('fs');
 const path = require('path');
+const request = require('request');
 
 const clientSecret = process.argv[2] || process.env.AMP_DOC_SECRET;
 const clientId = process.argv[3] || process.env.AMP_DOC_ID;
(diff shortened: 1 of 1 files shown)


scripts/import_docs.js
```diff
@@ -25,6 +26,7 @@ const client = github.client(
   });
 
 const ghrepo = client.repo('ampproject/amphtml');
+var latestReleaseTag = null;
```
rsimha: nit: Perhaps use `let` instead of `var`?

scripts/import_docs.js
```diff
@@ -255,7 +257,22 @@ async function downloadExtensions() {
...
+// to make sure our docs always match the latest released AMP version.
+request({
+  url: 'https://api.github.com/repos/ampproject/amphtml/releases/latest',
+  headers: {
+    'User-Agent': 'request'
+  },
+  json: true
+}, function (error, response, body) {
+
+  latestReleaseTag = body.tag_name; // updates global var, used in the other functions
```
rsimha: Would it make sense to verify that `body.tag_name` is a non-empty string?

PR #1276, opened 2018-10-20
[KO] Improve responsive_amp/style_pages@ko.md.

captain-lee:
Current EN page: https://www.ampproject.org/docs/design/responsive/style_pages
Current KO page: https://www.ampproject.org/ko/docs/design/responsive/style_pages

Note that EN page has an update which is coming: https://github.com/ampproject/docs/pull/1234/files
And this KO adopted the update as well.


captain-lee:
@techhtml, PTAL.

PR #1294, opened 2018-11-15
Adding missing redirects from /docs/design/* to /docs/design/responsive/*

sandersk:
Adding missing redirects from /docs/design/* to /docs/design/responsive/* to fix a handful of 404s I spotted, e.g. the first few links on the [Style & Layout page](https://www.ampproject.org/docs/design/responsive_amp).

### DIFF
--- a/firebase.json
+++ b/firebase.json
@@ -954,6 +954,31 @@
         "type": 301
       },
       {
+        "source": "/docs/design/art_directions?(.html)",
+        "destination": "/docs/design/responsive/art_direction.html",
+        "type": 301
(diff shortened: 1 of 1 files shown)


pbakaus:
Hi thanks for the contribution! I think we broke these links in a recent PR, and I think the proper way to fix those is in the actual page, not add more redirects. Feel free to close this one, or use it to do what I suggested. Up to you :) Thanks again.

pbakaus:
Closed in favor of bigger fix in #1300.

PR #1338, opened 2019-01-08
Added new analytics vendor: Retargetly

fdnieves:
Added Retargetly vendor description inside docs/content/docs/analytics/analytics-vendors.md files


fdnieves:
I signed it! Cla has been signed nef

PR #1347, opened 2019-01-14
Add seatmap guide

kul3r4:
This adds an AMP seatmap guide implementation.
I wasn't able to get the code snippet to proper render, it contains some server side code such as {{#name of the variable}} which I don't know how to escape.
I tried with 
`{% raw %}{{#name of the variable}}{% endraw %}` but it still doesn't work, I think because it's not inside an html tag. (you can take a look at the code to see what I exactly mean)

@pbakaus @CrystalFaith any idea how to fix this in grow.io?


kul3r4:
@matthiasrohmer I think we are ready to merge this PR if it looks ok for you.

PR #1348, opened 2019-01-14
Migrate playground

sebastianbenz:
* adds playground from ampbyexample
* introduces embed mode, test via: http://localhost:8080/playground/embed#mode=Responsive&url=https://ampbyexample.com/introduction/hello_world/source/
* adds playground backend to platform (not sure if this a good way to do
it)
* adds playground build script to platform, which needs to be invoked
manually for now.
* implements fetch playground api

Note: component list api is not yet migrated from Go.


sebastianbenz:
Note for review, the following files are new, the rest is simply copied from ampbyexample:

* development.js (which still needs a webpack integration for local dev)
* backend/**/*
* src/embed-mode/ (which is used in src/document/controller)

Sorry, I should've split this into two PRs.

sebastianbenz:
One more thing: the normal playground looks broken now due to the dark theme.

sebastianbenz:
I think that we don't need to serve the playground from a different (sub)domain for the format sampler. It's only needed if we're using components that require local storage / cookies to work (amp-user-notification,...). 

However, it think it'd be best to serve the playground from a different subdomain: `playground.amp.dev`. //cc @pbakaus

PR #1357, opened 2019-01-21
Locally host custom fonts

sebastianbenz:
...to improve page load performance.

- adds custom font imports to base.scss
- adds preloads to playground
- adds preloads to default layout template
- adds downloaded versions of Google Fonts.

I'm not aware of a simple way to avoid duplicating the preload
declarations. Maybe we should switch the playground to j2 templates as
well.

Other future optimization would be using amp-font, we'd need to
configure a good default fallback for that case though.


matthiasrohmer:
Took care of the needed adjustments, works fine for the Grow pod and also seems to for the playground. If you have trouble getting Grow to run try `npm run develop -- --clean-samples` - it might have problems with updated routes.

PR #1364, opened 2019-01-30
add google analytics

sebastianbenz:


sebastianbenz:
@jeffjose can you please take a look whether I've setup things correctly
@matthiasrohmer please check if I wired things up correctly and if all templates now have an analytics integration.

boilerplate/templates/index.html
```diff
@@ -272,23 +272,19 @@ <h3>Bookend Sample</h3>
          role="button"
          on="tap:boilerplate.scrollTo(duration=200)"
          class="fab">View Code</div>
-    <amp-analytics type="googleanalytics" id="analytics">
-      <script type="application/json">
-        {
-          "vars": {
-            "account": "UA-73836974-1"
-          },
-          "triggers": {
```
jeffjose: Did we miss porting over `triggers` over to the new setup/

sebastianbenz: I thought that this is not really needed as page views are tracked by default. Or am I wrong about this?

matthiasrohmer:
LGTM at first glance. I'd like to take another look tomorrow with a clear mind. Something minor I thought about immediately is that we could disable tracking for local development and staging at least for all pages generated by Grow, if it helps. Would be something along the lines of

```jinja2
{% if not doc.pod.env.dev and not podspec.env.name == 'staging' %}
<amp-analytics ...>
```

sebastianbenz:
Tracking is disabled for staging and localhost via GA.

boilerplate/templates/index.html
sebastianbenz: We discussed this offline.

PR #1371, opened 2019-02-06
Add serviceworker

sebastianbenz:


sebastianbenz:
@prateekbh thanks for taking a look. Any other features we should add?

PR #1378, opened 2019-02-08
🐛 Fix broken references.

matthiasrohmer:
There were indeed broken references in the docs I migrated last-minute, sorry.


matthiasrohmer:
> Did these cause the build to fail?

Most probably. A local build with `--import` did go fine for me now.

sebastianbenz:
Did it actually run the import? If false references in docs break the import something is seriously wrong...

PR #1381, opened 2019-02-08
more robust error handling

sebastianbenz:


matthiasrohmer:
Waiting for the `close` event on the Grow stream instead of inspecting logs is a good idea. That probably also makes the timeout in `pipeline.js` obsolete. Throwing errors for minifying is basically also fine for me I am just pondering if we should postpone that change to when there is something on staging again.

sebastianbenz:
Simplified the PR. We probably don't need the `when` part at all. `close` and checking if `statusCode === 0` should be sufficient (assuming grow returns status=1 if rendering fails). 

@matthiasrohmer PTAL

matthiasrohmer:
Seems good 🙂At least a purposefully broken local build exits properly.

PR #1387, opened 2019-02-09
Format filter on request instead of statically

matthiasrohmer:


platform/lib/routers/pages.js
```diff
@@ -144,16 +148,40 @@ if (config.environment === 'development') {
...
+
+    // Match / to index.html
+    if (requestPath == '/') {
+      requestPath = '/index.html';
     }
 
-    next();
-  }, express.static('pages'));
+    // Check if the requested file exists
+    if (!fs.existsSync(path.join(STATIC_PAGES_PATH, requestPath))) {
```
sebastianbenz: use async/await 

Tip: promisfy let's you convert readFile into a promised based version http://2ality.com/2017/05/util-promisify.html

platform/lib/routers/pages.js
```diff
@@ -144,16 +148,40 @@ if (config.environment === 'development') {
...
+      requestPath = NOT_FOUND_ROUTE;
+    }
+
+    fs.readFile(path.join(STATIC_PAGES_PATH, requestPath), (error, data) => {
```
sebastianbenz: same as above, use async/await

platform/lib/routers/pages.js
```diff
@@ -144,16 +148,40 @@ if (config.environment === 'development') {
...
+      if (error) throw error;
+      response.set('Content-Type', 'text/html');
```
sebastianbenz: Is this needed?

platform/lib/routers/pages.js
```diff
@@ -144,16 +148,40 @@ if (config.environment === 'development') {
...
+
+      // Check if the page should be format filtered
+      const activeFormat = getFilteredFormat(request);
+      if (activeFormat && isFilterableRoute(requestPath)) {
+        try {
+          const filteredPage = new FilteredPage(activeFormat, data);
```
sebastianbenz: sidenote: this looks a bit weird from an API point of view. Why use a class instead of a function?

Does not need to be fixed in this PR.

platform/lib/routers/pages.js
```diff
@@ -144,16 +148,40 @@ if (config.environment === 'development') {
 }
 
 if (config.environment !== 'development') {
```
sebastianbenz: Doesn't this mean that filtering only work on non dev mode?

platform/lib/routers/pages.js
```diff
@@ -144,16 +148,40 @@ if (config.environment === 'development') {
...
-  pages.use('/', (request, response, next) => {
-    // Check if this request should be filtered
-    const activeFormat = getFilteredFormat(request);
-    if (activeFormat) {
-      // And if it should be filtered rewrite to the correct file
-      request.url = request.url.replace('.html', `.${activeFormat}.html`);
+  const STATIC_PAGES_PATH = utils.project.absolute('platform/pages');
+  const NOT_FOUND_ROUTE = '/404.html';
+
+  pages.use('/', (request, response) => {
```
sebastianbenz: Afaik, I you implement a middleware, you should call `next`. Or am I missing anything?

platform/lib/routers/pages.js
```diff
@@ -144,16 +148,40 @@ if (config.environment === 'development') {
...
+    let requestPath = request.path;
+
+    // Match / to index.html
+    if (requestPath == '/') {
```
sebastianbenz: It'd be good to do this for all paths.

platform/lib/routers/pages.js
sebastianbenz: Another suggestion, as you already have a project helper, why not add a method there: `project.pagePath(requestPath)`.

platform/lib/routers/pages.js
matthiasrohmer: Good hint, refactored it.

platform/lib/routers/pages.js
matthiasrohmer: Adopted!

platform/lib/routers/pages.js
matthiasrohmer: Strangely it is, otherwise the files got downloaded.

platform/lib/routers/pages.js
matthiasrohmer: That's right. Filtering for development is handled a bit differently as it involves rewriting the proxied response from Grow. See line 102 and onwards.

platform/lib/routers/pages.js
matthiasrohmer: I am unsure about this to be honest - the built-in static middleware seems to terminate requests and to directly send the file regardless any following middlewares. But we don't need to adopt this behaviour for ours, then we'd just need to set the response body without directly sending it. Do you know how?

platform/lib/routers/pages.js
matthiasrohmer: Adopted!

matthiasrohmer:
Thanks @sebastianbenz for taking a first look. Let me know if my latest changes work for you 🙂

platform/lib/routers/pages.js
```diff
@@ -144,16 +149,41 @@ if (config.environment === 'development') {
...
+    }
+
+    // Check if the requested file exists
+    if (!fs.existsSync(utils.project.pagePath(requestPath))) {
+      requestPath = NOT_FOUND_ROUTE;
+      response.status(404)
+    }
+
+    let page = await readFileAsync(utils.project.pagePath(requestPath)).catch((e) => {
+      throw e;
```
matthiasrohmer: That's also something I am unsure about: my feeling is that this should somehow generate a 500 response, shouldn't it? I am curios what your thoughts are on this, @sebastianbenz.

platform/lib/routers/pages.js
sebastianbenz: This is caused by not calling next. Otherwise express will add the header (we should also check if content length and etag gets added.

platform/lib/routers/pages.js
sebastianbenz: We should have a generic error handler (with a custom 500 page). See https://expressjs.com/en/guide/error-handling.html#writing-error-handlers

sebastianbenz: `next(err)` instead of throw.

platform/lib/routers/pages.js
sebastianbenz: absolutely

platform/lib/routers/pages.js
sebastianbenz: You could wrap the static middleware:

```
const staticMiddleware = app.static('public');
app.use((req, res, next) => {
   if (hasFilter(...)) {
      // do filter
   } else {
        staticMiddleware(req, res, next);
   }
});
```

I've used a similar pattern to enable the optimizer: https://github.com/ampproject/docs/pull/1385/files#diff-c0ef79f636352ecae379454e61514ef8R63

sebastianbenz:
ping

matthiasrohmer:
Sorry for the delay, @sebastianbenz. PTAL. There is now no explicit error handling in the pages handler (also not for 404s) anymore. The generic error handler is then to come in another PR. 

My guess that calling `res.send` will terminate the request was indeed wrong - calling `next()` sets the required headers. All unfiltered requests are nevertheless now handled by the built-in static middleware.

What I can't get my head around is using a promisified existance check over `fs.existsSync` - if it is really just for consistency reasons I would refrain from going that way since I feel `fs.access` is a lot more verbose. But ultimately it's your call of course.

sebastianbenz:
We should keep everything async, feel free to add a helper method that wraps fs.access.

platform/lib/routers/pages.js
```diff
@@ -144,16 +147,36 @@ if (config.environment === 'development') {
...
+    // eslint-disable-next-line max-len
+    if (activeFormat && isFilterableRoute(requestPath) && fs.existsSync(utils.project.pagePath(requestPath))) {
+      // Match root requests to a possible index.html
+      if (requestPath.endsWith('/')) {
+        requestPath = requestPath + 'index.html';
+      }
+
+      // Read the file to a variable to modify it
+      const page = await readFileAsync(utils.project.pagePath(requestPath))
+          .catch((e) => {
```
sebastianbenz: use try/catch with async/await

platform/lib/routers/pages.js
```diff
@@ -144,16 +147,36 @@ if (config.environment === 'development') {
...
+            throw e;
+          });
+
+      // And then filter the page and send it
+      const filteredPage = new FilteredPage(activeFormat, page, true);
+      response.send(filteredPage.content);
+      next();
+    } else {
+      // ... otherwise pass on to the static middleware
+      staticMiddleware(request, response, next);
```
sebastianbenz: nit: use an early return instead 

```
if (!shouldApplyFormatFilter(req)) {
  return staticMiddleware(request, response, next);
}
...
```

platform/lib/routers/pages.js
matthiasrohmer: Done, but does catching here make sense at all since we are just rethrowing for the express error handler to catch?

matthiasrohmer:
Updated! Please don't hold off with feedback if you have further comments 🙂

platform/lib/routers/pages.js
```diff
@@ -150,32 +150,47 @@ if (config.environment !== 'development') {
...
+    try {
+      const page = await readFileAsync(utils.project.pagePath(requestPath));
+      const filteredPage = new FilteredPage(getFilteredFormat(request), page, true);
       response.send(filteredPage.content);
-      next();
-    } else {
-      // ... otherwise pass on to the static middleware
-      staticMiddleware(request, response, next);
+    } catch (e) {
+      throw e;
```
sebastianbenz: Catch and re-throw is kind of redundant. Correct way is to return the error via next.

matthiasrohmer: Thought so.

PR #1395, opened 2019-02-18
💄 change cookie layer from left to right

timokohlmann:
Cookie layer pops up from the right side now

### DIFF
--- a/frontend/scss/components/organisms/consent.scss
+++ b/frontend/scss/components/organisms/consent.scss
@@ -15,7 +15,8 @@
 @import 'components/atoms/_button.scss';
 
 .#{organism('consent')} {
-  left: 20px;
+  right: 20px;
+  left: auto;
   bottom: 20px;
   width: calc(100% - 40px);
   padding: 15px;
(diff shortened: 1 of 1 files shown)


frontend/scss/components/organisms/consent.scss
```diff
@@ -15,7 +15,8 @@
 @import 'components/atoms/_button.scss';
 
 .#{organism('consent')} {
-  left: 20px;
+  right: 20px;
+  left: auto;
```
matthiasrohmer: `auto` is the [default value](https://developer.mozilla.org/de/docs/Web/CSS/Left) for the left-property. Therefore doesn't need to be set explicitly.

timokohlmann: Left: auto is still in there because there is a left-property in /extensions/amp-consent/0.1/amp-consent.css which has to be overwritten

matthiasrohmer: 🤦‍♂️ then of course it's fine.

PR #1402, opened 2019-02-26
Link documents in Guides & Tutorials

matthiasrohmer:
That's the announced PR to a) link all component names to their actual reference page and unify their formatting to be like `amp-component`. Note that this comes with the downside that all pages linking to a component become unusable for someone who hasn't done a import. And b) fixing a ton of references that have used hardcoded links to either ampproject.org or ampbyexample.com to use the `{{g.doc(...` syntax.

@robinvanopstal did the fixing by using an insane amount of RegExes to cover all edge cases, ran the linking tools then bucket by bucket (Start, Learn, ...) and did a manual check afterwards. So the changes should be fairly safe 🙂 

Additionally this will hopefully fix a lot of the immense amount of 404s that Thomas encountered lately.


matthiasrohmer:
I guarantee that Robin is fine with his changes being contributed. If we require him personally to confirm it, he'll be back from vacation by the 15th of March.

PR #1406, opened 2019-02-27
containerize amp.dev

sebastianbenz:


sebastianbenz:
This is ready to merge. Deploying to staging should work as well.

sebastianbenz:
For a prod release run `cd build_scripts make deploy`.

platform/config/environments/staging.json
```diff
@@ -14,7 +14,7 @@
 		"platform": {
 			"scheme": "https",
 			"host": "amp-dev-staging.appspot.com",
-			"port": ""
+			"port": "8080"
```
matthiasrohmer: This will cause Grow to build URLs with port like `amp-dev-staging.appspot.com:8080/index.html` if I'm not mistaken. Should we add another host object for Grow?

sebastianbenz: good find, this isn't needed anymore.

sebastianbenz: fixed

PR #1407, opened 2019-02-27
Update page title for homepage

### LINKED ISSUE #1389: [amp.dev] Bad title for homepage
The homepage title is currently just "Start".

Also the link on the logo goes to `/index.html`. Please change to `/`.

matthiasrohmer:
This one fixes #1389 while at the same time introducing the functionality to define special titles for specific pages by using 

```yaml
---
$title: Start
$titles:
  title: AMP - a web component framework to easily create user-first web experiences
```

in their frontmatter. At the same time all page titles just defined by `$title` will be postfixed by " - AMP" just like on ampproject.org.

### DIFF
--- a/pages/content/amp-dev/index.html
+++ b/pages/content/amp-dev/index.html
@@ -1,5 +1,7 @@
 ---
 $title: Start
+$titles:
+  title: AMP - a web component framework to easily create user-first web experiences
 $view: /views/custom.j2
 
 news: !g.yaml /content/shared/data/blog.yaml
--- a/frontend/templates/layouts/default.j2
+++ b/frontend/templates/layouts/default.j2
@@ -24,7 +24,12 @@
     {% block head %}
     <meta charset="utf-8"/>
     <script async src="https://cdn.ampproject.org/v0.js"></script>
-    <title>{{ doc.title }}</title>
+    {# If there is a special title use that one without a postfix #}
+    {% if doc.titles('title') != doc.title %}
+    <title>{{ doc.titles('title') }}</title>
+    {% else %}
+    <title>{{ doc.title }} - AMP</title>
+    {% endif %}
     {% if podspec.env.name == 'development' %}
     <link rel="shortcut icon" href="/static/img/favicon-local.png"/>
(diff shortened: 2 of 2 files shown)


sebastianbenz:
What's the difference between `title` and `titles`?

matthiasrohmer:
The titles in the `$titles` dictionary can be called via `doc.titles(*key*)` which falls back to `$title` if no title with that key is defined. That allows to define different titles for various use cases (`<title>`, headline, in navigations). Currently we only made use of it for the header.

pages/content/amp-dev/index.html
```diff
@@ -1,5 +1,7 @@
 ---
 $title: Start
+$titles:
+  title: AMP - a web component framework to easily create user-first web experiences
```
sebastianbenz: Why not replace `start` with the correct title?

matthiasrohmer: Not the case but when the homepage would be met by any of the navigation walkers (sidebar, header) they would print this full-blown title.

We could flip the pattern and use `$titles.nav` instead for navigation titles.

matthiasrohmer:
`$titles` is the required name from Grow. Only the keys inside the list are free to choose.

pages/content/amp-dev/index.html
sebastianbenz: > We could flip the pattern and use `$titles.nav` instead for navigation titles.

SGTM - that would make it more clear

sebastianbenz:
Ah ok - wasn't aware. `titles` it is then. One comment, otherwise LGTM.

pages/content/amp-dev/index.html
matthiasrohmer: Cool, will ping you when done 🙂

sebastianbenz:
@matthiasrohmer are you going to change to `$titles.nav` as part of this PR?

matthiasrohmer:
In this one, I just want to wait for #1416.

matthiasrohmer:
Closed by #1422 and #1424.

PR #1419, opened 2019-03-07
Use new gifs and update templates

kul3r4:
only thing missing here is to decide the alignment of the heading on text template, the PR will be ready to be merged once I finish that


kul3r4:
@calebcordry @stillers FYI

sebastianbenz:
@kul3r4 would it be possible to use videos instead of GIFs due to smaller file sizes?

kul3r4:
videos are not supported a s a preview, when I tried they were not rendering

kul3r4:
@sebastianbenz this is now ready to be merged

calebcordry:
Can we change the preview code so that it will show videos? I agree with @sebastianbenz

sebastianbenz:
That'd be great! I think you'd need to change the previews [here](https://github.com/ampproject/docs/blob/f6f6dc41f34917f1f3601f56f03c80b06b3a2fce/frontend/templates/views/partials/teaser.j2#L60). @matthiasrohmer can tell you exactly what you need to do.

kul3r4:
Would that be ok to do it in another PR? This will take time as I need to
request to generate the videos.

PR #1420, opened 2019-03-07
🐛 Fix xlink error thrown by amp validator

dandzdandz:
Removed all the xmlns:xlink="http://www.w3.org/1999/xlink" attributes in the <use> tags to prevent the amp validator of going haywire


matthiasrohmer:
That doesn't seem to be needed for the live validator, see https://amp-dev-staging.appspot.com/?amp=#development=1. Can you verify?

matthiasrohmer:
`xmlns:xlink` are definitely needed for Safari, the `xmlns:` prefix gets lost somewhere during generation process. Needs to be fixed differently.

PR #1423, opened 2019-03-07
improve playground embed preview

sebastianbenz:
screenshot 2019-03-08 at 00 16 19

### DIFF
--- a/examples/static/samples/standalone/amp-story.html
+++ b/examples/static/samples/standalone/amp-story.html
@@ -3,33 +3,22 @@
   <head>
     <meta charset="utf-8">
     <meta name="viewport" content="width=device-width,minimum-scale=1">
-    <meta name="description" content="This is the AMP Boilerplate.">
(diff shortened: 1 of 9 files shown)


pbakaus:
How do you edit long lines that are under the preview?

matthiasrohmer:
Longos suggestion would be to wrap the text on the preview's border.

pbakaus:
@matthiasrohmer I mean that works, but then it's essentially not "inline", and only looks that way, and could equally well be on the side.. you know what I mean? ;)

sebastianbenz:
My thinking was that the boilerplate works in our favour here. It's so long, that you can easily scroll to any other line (we could even ensure this by adding more padding-right to the source code view).

sebastianbenz:
We can’t wrap due to the boilerplate as would spread over multiple lines.

sebastianbenz:
I'll merge for now, but I agree - we need to find a better solution.

PR #1434, opened 2019-03-08
added amp roadshow to events 

keevkeev:
AMP roadshow and its stylings were created. Due to time constraints this week, some of the roadshow contents are not separated Atomically, this will be handled later.


frontend/scss/components/molecules/event.scss
```diff
@@ -0,0 +1,125 @@
...
+
+*/
+
+@import '_extends.scss';
+@import '_functions.scss';
+@import '_mixins.scss';
+@import '_variables.scss';
+
+@import 'components/atoms/_text.scss';
+@import '../atoms/_color.scss';
```
matthiasrohmer: Please use absolute (`@import 'components/atoms/_text.scss';`) imports consistently.

frontend/scss/components/molecules/faq.scss
```diff
@@ -0,0 +1,43 @@
+/*
+
+#########################
+### MOLECULE: faq ###
+#########################
+
+*/
+
+@import '../../extends';
```
matthiasrohmer: Please use absolute imports.

frontend/templates/views/partials/roadshow-video.j2
```diff
@@ -0,0 +1,7 @@
+{% do doc.styles.addCssFile('css/components/molecules/roadshow-video.css') %}
```
matthiasrohmer: As this is possibly never used on any page except the roadshow page, either directly add it to the template or rename to a more generic partial.

pages/content/shared/data/roadshow.yaml
```diff
@@ -0,0 +1,156 @@
+nextstops:
```
matthiasrohmer: There was an update on the `production` branch for the stations, can you check if it still matches those here?

keevkeev: are the added fields to nextstop unwanted?

PR #1508, opened 2019-03-11
 🔥cut short css-paths

timokohlmann:


matthiasrohmer:
Thanks though we'll postpone this change to later, when things aren't moving that much anymore.

PR #1517, opened 2019-03-11
💬 Add Advance Create success story.

matthiasrohmer:
The sister-PR to #1421. Adds the Advance Create success story to amp.dev with the existing material  - which is fairly limited, speaking of the images 😉


mattludwig:
Approved - I realize we are light on images, but let's go ahead and get it live for now :)

PR #1526, opened 2019-03-12
workaround: point playground to staging until subdomains work

sebastianbenz:


sebastianbenz:
Needed until amp.dev subdomains are available

PR #1539, opened 2019-03-12
Offline optimizer

sebastianbenz:


platform/lib/build/pageMinifier.js
```diff
@@ -87,22 +90,39 @@ class PageMinifier {
...
           callback();
         }))
         .pipe(gulp.dest('./'));
   }
 
+
+  async optimize(html, path) {
+    const ampRuntimeVersion = await runtimeVersionPromise;
+    return ampOptimizer.transformHtml(html, {
+      ampUrl: path.replace(/^pages/, ''),
```
sebastianbenz: @matthiasrohmer I'm sure there's a better way to do this

package.json
```diff
@@ -59,6 +56,9 @@
     "signale": "1.4.0"
   },
   "devDependencies": {
+    "amp-toolbox-optimizer": "0.4.8",
+    "amp-toolbox-optimizer-express": "0.5.1",
```
matthiasrohmer: Is the middleware still a dependency?

platform/lib/build/pageMinifier.js
matthiasrohmer: None that I can think of quickly... 🤔

package.json
sebastianbenz: good catch

sebastianbenz: fixed

PR #1568, opened 2019-03-14
Update amp_to_pwa.md

YaguraStation:
`sw.js`being used in starter code for service worker, not `serviceworker.js`.


matthiasrohmer:
That has been fixed for amp.dev in the mean while. Thanks!

PR #1570, opened 2019-03-14
Update sample routes & preview handling.

matthiasrohmer:
PR goes towards `fix/samples` as building those features on `future` would have caused major merge conflicts, when `fix/samples` will get merged.

- Splits sample routers in one for source files and one for embeds
- Uses Grow for rendering (and delivery) of preview pages
- Loads sources in Iframe for preview
- Adjusts route scheme to use
-- Source: examples/components/amp-accordion
-- Snippet: examples/components/amp-accordion-7
-- Embed: examples/components/amp-accordion/embed
-- Preview: documentation/examples/components/amp-accordion/preview
-- Documentation: -- Preview: documentation/examples/components/amp-accordion

Needs to be done before merge: scope source and embed routers to preview.amp.dev sub-domain, make screen size toggleable on preview page.


frontend/js/story-progress.js
```diff
@@ -0,0 +1,71 @@
...
+  name = name.replace(/[\[\]]/g, "\\$&");
+  var regex = new RegExp("[?&]" + name + "(=([^&#]*)|&|#|$)"),
+  results = regex.exec(url);
+  if (!results) return null;
+  if (!results[2]) return '';
+  return decodeURIComponent(results[2].replace(/\+/g, " "));
+}
+
+// Enable story experiment
+enableExperiment();
```
sebastianbenz: This is no longer needed

sebastianbenz: (only the switch to page part is needed.

platform/lib/build/samplesBuilder.js
```diff
@@ -121,7 +121,7 @@ class SamplesBuilder {
 
       stream = stream.pipe(through.obj(async (sample, encoding, callback) => {
         this._log.await(`Building sample ${sample.relative} ...`);
-        await this._parseSample(sample).then((parsedSample) => {
+        await this._parseSample(sample).then(async (parsedSample) => {
```
sebastianbenz: `await` and `then` in the same line look fishy. Try to only use await.

platform/lib/build/samplesBuilder.js
```diff
@@ -134,7 +134,7 @@ class SamplesBuilder {
             ...this._createDocumentation(sample, parsedSample),
             ...this._buildRawSources(sample, parsedSample),
             ...this._createPreview(sample, parsedSample),
-            ...this._renderEmbed(sample, parsedSample),
+            ...(await this._renderEmbed(sample, parsedSample)),
```
sebastianbenz: is it needed to wait here?

platform/lib/routers/exampleEmbeds.js
```diff
@@ -0,0 +1,33 @@
...
+'use strict';
+
+const express = require('express');
+const utils = require('@lib/utils');
+
+const EMBEDS_DEST = utils.project.absolute('/dist/examples/embeds/');
+
+// eslint-disable-next-line new-cap
+const exampleEmbeds = express.Router();
+const embeds = express.static(EMBEDS_DEST);
```
sebastianbenz: suggestion: rename to `staticEmbeds`

platform/lib/routers/exampleEmbeds.js
```diff
@@ -0,0 +1,33 @@
+/**
```
sebastianbenz: suggestion: move the sample routers into a `example` subdir to avoid duplicating the example prefix in the file name.

platform/lib/routers/exampleSources.js
```diff
@@ -0,0 +1,38 @@
...
+'use strict';
+
+const express = require('express');
+const utils = require('@lib/utils');
+
+const SOURCES_DEST = utils.project.absolute('/dist/examples/sources/');
+
+// eslint-disable-next-line new-cap
+const exampleSources = express.Router();
+const sources = express.static(SOURCES_DEST);
```
sebastianbenz: rename to `staticSources`

matthiasrohmer:
Thanks @sebastianbenz for taking a first look, adopted all your suggestions 🙂 

I wanted to implement the preview subdomain with c1e25fd but that didn't quite work out: http://localhost:8082/preview/examples/components/amp-accordion should give the source but the responsible router doesn't seem to get the request. You've got an idea?

.dockerignore
```diff
@@ -0,0 +1,35 @@
+node_modules
```
sebastianbenz: Why are these changes in here?

sebastianbenz:
@matthiasrohmer you needed to register the proxy server with the subdomain handler. However, this is no longer needed with my latest change from demoes. The only thing you need to change is to pick a different port for preview in development.

.dockerignore
matthiasrohmer: Somehow GitHub can't divide the merged changes from the ones done on the branch itself.

matthiasrohmer:
Cool, seems to work! However I needed to adjust subdomain.js to not start the preview stub server twice. There might be a better way than what I did, see 364f04e.

And I probably used an (soon to be?) outdated playground scheme: http://localhost:8083**/#url**=http://localhost:8084/examples/components/amp-accordion. Can you point me to the correct one?

Last thing: config.hosts.preview needs to be added to VALID_ORIGINS in playground/api.js - shall I hardcode it?

When this is done I think we are good to merge back to `fix/samples`.

sebastianbenz:
> However I needed to adjust subdomain.js to not start the preview stub server twice. There might be a better way than what I did, see 364f04e.

LGTM

> Last thing: config.hosts.preview needs to be added to VALID_ORIGINS in playground/api.js - shall I hardcode it?

SGTM

sebastianbenz:
re: URL scheme. Use the existing scheme, I'll migrate with my still open PR.

PR #1573, opened 2019-03-15
Adds readme for animations and comments

kul3r4:
- add readme.txt to the templates using animations (text, single image, multiple image)
- add additional comments how to view templates when animations are in place
- fix path of an image for single image template

@calebcordry PTAL


kul3r4:
@ithinkihaveacat FYI

calebcordry:
Thanks! I cant actually see anything inside the files

kul3r4:
Which files in particular? Do you mean they are empty?

calebcordry:
Sorry, I just meant that the github diff view doesn't show what is inside the files. I then realized that I could just download the PE :). Anyway LGTM, thanks again!

kul3r4:
@sebastianbenz PTAL

PR #1574, opened 2019-03-15
Add story sample

sebastianbenz:
* add the Mars Explorer story sample
* implement subdomain redirects to document origin for playground
requests (this was needed as the story sample uses relative URLs)
* url is now specified via query param instead of a hash param


sebastianbenz:
@matthiasrohmer PTAL. Had to make a few fixes. Live version is here: https://amp-dev-staging.appspot.com/

platform/lib/middleware/subdomain.js
```diff
@@ -53,6 +63,44 @@ class Subdomain {
       return next();
     };
   }
+
+  async redirectToReferrerOn404_(request, response) {
```
matthiasrohmer: That's really a nit, but I'd love this to have a short comment what problem this actually solves when we or somebody else is coming back to it later 🙂

sebastianbenz: Makes sense. done.

PR #1581, opened 2019-03-19
Feature/amp conf

keevkeev:
new ampconf layout


frontend/scss/components/molecules/agenda-item.scss
```diff
@@ -15,56 +15,122 @@
...
+      border-radius: 50%;
+      align-self: flex-start;
+      margin-right: 10px;
+      margin-left: -60px;
+    }
+
+    &.multiple {
+      position: absolute;
+
+      &:nth-child(1) {
```
matthiasrohmer: Having that many `:nth-child` selectors doesn't seem like the most intuitive solution 😅

frontend/scss/components/molecules/agenda-item.scss
```diff
@@ -15,56 +15,122 @@
...
     width: 80%;
-    margin-right: auto;
-    padding: 7px 0 7px 25px;
     border-left: 2px solid color('blue-ribbon');
-    position: relative;
-    display: flex;
-    align-items: center;
-    color: color('charade');
-
+    padding: 13.5px 0 13.5px 25px;
```
matthiasrohmer: Please don't use half-pixel values.

frontend/scss/components/organisms/agenda.scss
```diff
@@ -33,6 +33,39 @@
...
+      flex-grow: 1;
+      outline: none;
+      color: fade_out(color('limed-spruce'), 0.57);
+      background: color('athens-gray');
+
+      &+.#{organism('agenda')} {
+        display: none;
+      }
+
+      &[option][selected] {
```
matthiasrohmer: Does that selector need to be double specific by using `[option]` and `[selected]` or would the later one be sufficient?

frontend/scss/components/organisms/stage-ads.scss
```diff
@@ -0,0 +1,122 @@
...
+
+  @media (min-width: 768px) {
+    margin-left: 64px;
+    border-bottom-left-radius: 6.250em;
+    -webkit-clip-path: polygon(0 0, 100% 0, 100% calc(100% - 3em), 6.250em 100%, 0% 100%);
+    clip-path: polygon(0 0, 100% 0, 100% calc(100% - 3em), 6.250em 100%, 0% 100%);
+  }
+
+
+  div.#{utility('container-fluid')} {
```
matthiasrohmer: Does that selector need to be that specific by prepending `div`?

frontend/scss/components/organisms/agenda.scss
keevkeev: amp-selector adds some default stylings to selected options using this double selector and without the double selector some of our styling properties are overridden (e.g. border: 'none')

frontend/scss/components/organisms/stage-ads.scss
keevkeev: all stage organisms are wrapped in a <section> tag that utilizes the "ap--container-fluid" class, figured using this selector to style the individual div was less css than creating a whole new utility class with the same properties as "ap--container-fluid" with the exception for a different padding

frontend/scss/components/molecules/agenda-item.scss
keevkeev: sorry about that

PR #1589, opened 2019-03-20
Feature/bug fixes

keevkeev:
Broken layout on support page should now be fixed


frontend/scss/components/templates/policies.scss
```diff
@@ -19,7 +19,7 @@
 @import '../templates/_default.scss';
 
 
-.#{template('code-of-conduct')} {
+.#{template('policies')} {
```
matthiasrohmer: `policies` sounds a bit specific. What do you think of `meta`?

frontend/templates/views/policies.j2
```diff
@@ -0,0 +1,17 @@
...
+{{ super() }}
+
+{% endblock %}
+
+{% block main %}
+{% do doc.styles.addCssFile('css/components/templates/policies.css') %}
+<main class="ap--main ap-t-policies">
+  <section class="ap--container">
+    <div class="ap--content-wide">
+      {{ doc.html|render|safe }}
```
matthiasrohmer: Don't we have a just-content-template already?

keevkeev: none that I could find that isn't loading a lot of unnecessary code.

frontend/scss/components/templates/policies.scss
keevkeev: sounds good to me 👍

frontend/templates/views/partials/footer.j2
```diff
@@ -60,7 +60,35 @@
       </ul>
       <div class="ap-o-footer-copyright">&copy; 2019 AMP Open Source Project</div>
     </div>
+
   </div>
+  <amp-position-observer
```
matthiasrohmer: To be honest I am not a fan of introducing another instance of `amp-position-observer` and `amp-animation` for this, I think I just dealt with this on another branch by fiddling around with the z-indexes and making the sidebar toggle vanish behind the footer. What do you think?

frontend/templates/views/policies.j2
matthiasrohmer: Alright, can we rename to `meta.j2/scss` then? 😉

keevkeev: aren't they already renamed or did I miss a file?

matthiasrohmer: No all good I was just looking not accurately 🙂 Sorry!

PR #1611, opened 2019-03-25
🎨 Improve format filtering.

matthiasrohmer:


matthiasrohmer:
Will extend this PR by something that removes empty level-1 sidebar items.

matthiasrohmer:
Leading tutorial dividers are now also removed, as well as empty first level categories. Additionally I fixed the issues with the dropdown you mentioned, should work a lot better now.

platform/lib/common/filteredPage.js
```diff
@@ -122,9 +121,22 @@ class FilteredPage {
...
+          if (!navItem.has('a').length) {
+            navItem.remove();
+          }
+        });
+
     // Remove eventually unnecessary tutorial dividers left by the
     // previous transformation
-    this._dom('.nav-item-tutorial-divider:last-child').remove();
+    /* eslint-disable max-len */
+    this._dom('.nav-item-tutorial-divider:last-child, .nav-item-tutorial-divider:first-child').remove();
```
sebastianbenz: nit: I'd prefer splitting the string over disabling eslint.

platform/lib/common/filteredPage.js
```diff
@@ -148,13 +160,15 @@ class FilteredPage {
     // Rewrite the active state (which is websites per default) to
     // the current active format
     const activeFormat = this._dom('.ap-m-format-toggle-selected');
-    if (!activeFormat.html()) {
-      console.error('doc did not specify an active format');
+    if (activeFormat.length == 0) {
+      throw new Error('Page has no active format.');
```
sebastianbenz: I don't like throwing an exception here, I it happens in prod the user gets a 502, but it's something we can recover easily from.

matthiasrohmer: Ah you are right. I only had the implementation for local in mind, the error is catched there. Will revert.

platform/lib/common/filteredPage.js
matthiasrohmer: Definitely the better solution 😅

PR #1622, opened 2019-03-26
💄 Rearrange sample preview header.

matthiasrohmer:
Not 100% happy, WDYT?
Bildschirmfoto 2019-03-26 um 13 03 30


sebastianbenz:
I like it, except the massive logo.

How about removing `Example Preview` from the logo and adding `Preview: ` in front of the title?

PR #1632, opened 2019-03-26
Implement statics collector.

matthiasrohmer:
Following the request for smaller PRs I thought I'd already onboard you on this 😉 That's something I started a few days back and love to hear your thoughts on: I started implementing `staticsCollector.js` which I planned to solve two issues with.

It creates WEBPs and image size renditions (later enabling srcsets) while collecting the static files. Additionally it automatically creates ZIP-archives from folders that end on `.zip`. This allows the template ZIPs to be properly version controlled without the need to manually zip them.

What do you think so far? My next step would be to add the statics inventory to modify the static middleware to react on `?width=144` and check in the inventory JSON if an image matching that size is available and then vending it.

You can try it with `node platform/lib/build/staticsCollector.js` - make sure to manually clean `platform/static` before.


sebastianbenz:
I'd pick a slightly different approach here (for the image part). I see this as part of the optimizer. I'd like to have a transformer that analyses all amp-imgs and rewrites them to use webp and srcset. The API could look something like:

```
class ImageTransformer() {
   
  constructor(imageOptimizer) {
     this.imageOptimizer_ = imageOptimizer;
  }

  transform(tree, params) {
     // discover amp images
    const ampImgNode = ...
    const srcSet = [];
    // resize will return the path to resized image
    srcSet.push(this.imageOptimizer.resize(ampImgNode.attribs.src, 'w220'));
    srcSet.push(this.imageOptimizer.resize(ampImgNode.attribs.src, 'w320'));
    ...
   addSrcSet(ampImgNode, srcSet)
   addWebp(ampImgNode, this.imageOptimizer.webp(ampImgNode.attribs.src));
  }
}
```

Sharp is a good candidate for using it as imageOptimizer. 

The zip part is great!

matthiasrohmer:
Core feature of having virtual ZIPs is part of #1881.

PR #1652, opened 2019-03-28
🐛 Fix sidebar scrolling behaviour.

keevkeev:
Component sidebar and sidebar contained a max-height and position sticky that caused the user to have to scroll on desktop devices. For mobile devices the overflow auto property allowed for horizontal  scrolling when overflowed


matthiasrohmer:
Thanks, Andrew. Forwarding to Sebastian to make sure that is what he intended. @sebastianbenz, this would render the sidebar in its full length, even if the content doesn't span the full height. See:

Bildschirmfoto 2019-03-28 um 12 24 05

sebastianbenz:
Interesting side-effect when the sidebar is longer than the content. Can we make it so that it doesn't grow bigger than the content? Otherwise it's weird to have so much empty space before the footer.

keevkeev:
having the height of the sidebar based on the height of the content of component page was a nice solution, however still caused the problem of users having to scroll to reach the sidebar if the component page was larger than the content in the sidebar. The solution I have pushed positions the sidebar to sticky as the user scrolls with a fixed height of 100vh which should prevent the page being larger than it needs to for component pages with less content

matthiasrohmer:
TBH, I think the scrolling behaviour now got worse than it has been before since you have three scrollable somehow nested containers: body, content pane, sidebar. I know it's implemented as you've explained to me, but trying it out for myself I am not really satisfied. 😅I'll cherry pick the images bug, then we need to figure out a different solution for the sidebar.

pbakaus:
what's the status here?

matthiasrohmer:
Closed in favor of #2195.

PR #1660, opened 2019-03-29
Update showcased stories.

robinvanopstal:
This removes the Mic.com and Terra stories and adds SFChronicle, Washington Post and USA Today stories.


sebastianbenz:
Quick note: don't prepend https://cdn.ampproject.org/c/s/ to the URLs in the source code as it'll redirect to the actual cache URL. Either follow the redirect or use [this tool](https://ampbyexample.com/advanced/using_the_google_amp_cache/) here to calculate the correct cache URL.

PR #1670, opened 2019-03-31
Use AMP validator for dependency injection.

matthiasrohmer:
That's a long due change: instead of using unreliable string lookups to determine the used AMP extensions we use the validator to get `MISSING_REQUIRED_EXTENSION` errors that are automatically resolved while building the page. The used extensions are cached on locale-basis.

Gosh... this kills build times. What it currently does is calling the Node validator as a subprocess for each request/page which of course has a lot of overhead. I also experimented with reusing the same process but this breaks as soon as Grow starts multithreading. Another approach I considered was firing up another server (as if we haven't enough already) and using HTTP to communicate between Grow and Node but the overhead might be similar to processes.

Other options I see: use the Protobuffs to build a Python validator or leave the string lookup solution in place and fix-up possible errors in the JS pipeline. While researching on the validator I stumpled over references to a C++ validator but that doesn't seem to be open sourced - that could have been another option then.

Another way to go (that Paul advocated against if I remember correctly) would be to use explicit hints in the templates similar to `{% do doc.amp.use('amp-animation') %}`. That would be by far the speediest solution.


sebastianbenz:
Thanks for looking into this. TBH, I'm in favour of explicitly declaring dependencies via ` {% do doc.amp.use('amp-animation') %}`. Two reasons for this:

1. build times
2. validator doesn't reliably recognize missing extensions in all cases (e.g. amp-lightbox-gallery). 

Adding @pbakaus for thoughts.

pbakaus:
TBH, I'm a fan of automatically figuring things out, but I'm not a fan of an even slower build. Historically it wasn't that big of an issue to insert the dependencies manually.

How about this:

1. We do manual dependency injection
2. We still have a gulp task or something that we can run once in a while to find missing dependencies

I'd love to be able to find and be able to fix missing ones, just not slow down every build.

matthiasrohmer:
Thanks for your thoughts

PR #1672, opened 2019-04-01
Update Ja translation of AMP Conf page to show schedule and speakers

sisidovski:
Hi

Some Japanese developers reached out to me and said there is still no schedule and speakers on the page announcing AMP Conf 2019. 
https://www.ampproject.org/ja/amp-conf/

This change will show the schedule and speakers on the Japanese page. Even though session and speakers info is written in English, it's better than nothing. And, amp.dev doesn't have the Japanese language page so far, it's okay to change only master branch.

### DIFF
--- a/content/pages/amp-conf@ja.html
+++ b/content/pages/amp-conf@ja.html
@@ -44,20 +44,110 @@
   <div class="wrapper">
     <h3 id="schedule">スケジュール</h3>
 
-    <p>詳細なスケジュールはまもなく発表されます。また、CFP の受付を開始しましたので（期日：2019年2月14日）、何か AMP についてコミュニティにシェアすべきテーマがありましたらぜひともお申し込みください。登壇経験は問いません！お待ちしております！</p>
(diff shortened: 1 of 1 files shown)


matthiasrohmer:
In the meantime @pbakaus brought over the japanese version, just hasn't been deployed yet. Thank you though!

sisidovski:
@matthiasrohmer 
Just a confirmation. @pbakaus 's change will be deployed to `ampproject.org` (master branch), right? Or Will `ampproject.org` be redirected to `amp.dev` soon?

A lot of people still see https://www.ampproject.org/ja/amp-conf/ and can't see the schedule at all. It's annoying situation for Japanese developers. We only have two weeks for AMP Conf. We should show the schedule on Japanese page as soon as possible.

pbakaus:
@sisidovski sorry it took us so long to follow up here! The page is now redirected to amp.dev since a couple days ago, so I hope this'll fix it and alleviate your concerns. Let me know otherwise.

sisidovski:
@pbakaus I confirmed it and the problem was fixed! Thank you Paul!

PR #1709, opened 2019-04-02
show a referrer notification

sebastianbenz:
I'm not proud of what I've done here...


sebastianbenz:
Not really. I should extract the template and the css. Although, I don't know a good way to integrate the CSS.

sebastianbenz:
@matthiasrohmer just rebased against future, PTAL before I merge

PR #1733, opened 2019-04-04
Don't call next() after serving a page

sebastianbenz:
This fixes the 404 handler being called after serving a page.


sebastianbenz:
This fixes the bug causing the doc pages not being served.

PR #1739, opened 2019-04-04
🐛 Reintroduce HTML extension for built files.

matthiasrohmer:
HTML extension is needed for gulp to properly get the files in `pageMinifier.js` and express to determine the mime type.


sebastianbenz:
Can you fix the lint errors

PR #1768, opened 2019-04-08
fix get-involved-banner text color

robinvanopstal:


pages/content/amp-dev/documentation/guides-and-tutorials/index.ads.html
```diff
@@ -155,5 +155,5 @@ <h2 class="ap-o-video-slider-text">Featured Videos</h2>
 {% do doc.styles.addCssFile('/css/components/organisms/banner.css') %}
 <section class="ap--banner">
   <h2 class="ap-o-banner-headline">Get Involved</h2>
-  <p class="ap-o-banner-copy">AMP is proudly an open source project with an active community of contributors. We welcome and appreciate documentation and translation contributions! Check out our community section to join our mission to make the web suck less.</p>
+  <span class="ap-o-banner-copy">AMP is proudly an open source project with an active community of contributors. We welcome and appreciate documentation and translation contributions! Check out our community section to join our mission to make the web suck less.</span>
```
sebastianbenz: Why did you switch to `span` here? From a markup perspective, `p` seems to be a better fit.

keevkeev: The css inside our template targets all `p` tags and is overwriting the stylings for `ap-o-banner`. It was my thought that a simple fix for this bug without adding additional css would be to change the tag type, however I agree that keeping the `p` tag is preferable. I'll work Robin to find an alternative

sebastianbenz: Cool, can we get the fix in the next 1 hour?

keevkeev: @sebastianbenz yes the fix should be on it's way shortly

sebastianbenz: 👍

PR #1776, opened 2019-04-08
🐛 Migrate webpush sample files.

matthiasrohmer:
ampproject/ampbyexample#1991.


examples/source/1.components/amp-web-push.html
```diff
@@ -32,7 +32,7 @@
     <!--
     The `amp-web-push` component requires two additional HTML files as well as an amp-web-push compatible Service Worker. These three files form the configuration described below.
     -->
-    <amp-web-push id="amp-web-push" layout="nodisplay" helper-iframe-url="<% hosts.backend %>/webpush/amp-web-push-helper-frame.html" permission-dialog-url="<% hosts.backend %>/webpush/amp-web-push-permission-dialog.html" service-worker-url="<% hosts.backend %>/sw.js">
+    <amp-web-push id="amp-web-push" layout="nodisplay" helper-iframe-url="<% hosts.platform %>/static/samples/webpush/amp-web-push-helper-frame.html" permission-dialog-url="<% hosts.platform %>/static/samples/webpush/amp-web-push-permission-dialog.html" service-worker-url="<% hosts.backend %>/serviceworker.js">
```
sebastianbenz: This won't work: <% hosts.backend %>/serviceworker.js. Instead, this sample should have it's own directory in which it contains it's own serviceworker + the other static files.

matthiasrohmer:
The router in `examples/source/1.components/amp-web-push/api.js` could be automatically registered as soon as there's a `/static/` folder, but I remember you have been called out once on doing things automatically ... 😉

sebastianbenz:
+1 to not doing it automatically for now - we can always add it later if it becomes a common pattern. 

Does the sample work with this approach?

matthiasrohmer:
It should though `amp-web-push` is a bit more picky about HTTPS and I have no local secure setup available adhoc: `<amp-web-push> must have a valid helper-iframe-url attribute. It should begin with the https://`.

PR #1803, opened 2019-04-09
add amp validation task

sebastianbenz:
We'll probably need more lightweight tasks as well...


gulpfile.js/validate.js
```diff
@@ -0,0 +1,38 @@
...
+
+const {src} = require('gulp');
+const {join} = require('path');
+const gulpAmpValidator = require('gulp-amphtml-validator');
+
+/**
+ * Validates all pages build into /platform/pages.
+ */
+function validate() {
+  return src(join(__dirname, '../platform/pages/**/*.amp.html'))
```
matthiasrohmer: Could use `@lib/utils/project.paths.GROW_BUILD_DEST` to be prepared when the build destination sometime gets move to `dist`.

examples/source/1.components/amp-web-push/static/sw.js
```diff
@@ -18,17 +18,9 @@ importScripts('/sw-toolbox/sw-toolbox.js');
 importScripts('https://cdn.rawgit.com/jakearchibald/idb/97e4e878/lib/idb.js');
```
matthiasrohmer: Was linting/fixing this file on purpose?

sebastianbenz: yes - is there a reason why it shouldn't be?

matthiasrohmer: Not really 😉was thinking about browser compatibility but every browser that does support SWs should also support modern syntax.

gulpfile.js/validate.js
sebastianbenz: good point - done

PR #1811, opened 2019-04-09
🐛 Fix local format filtering.

matthiasrohmer:
Local format filtering is broken as `fixCheerio` gets the cheerio instance instead of an HTML string. Additionally due to the URL change the lookup for manual filtered variants didn't work.

### DIFF
--- a/platform/lib/routers/pages.js
+++ b/platform/lib/routers/pages.js
@@ -93,7 +93,7 @@ if (config.isDevMode()) {
    * @return {Boolean}
    */
   async function hasManualFormatVariant(request, format) {
-    const path = request.originalUrl.replace('.html', `.${format}.html`);
+    const path = request.url.replace('.html', `.${format}.html`);
 
     const page = await got(`${growHost}${path}`).catch(() => {
       return {};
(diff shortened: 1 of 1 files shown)


platform/lib/routers/pages.js
```diff
@@ -93,7 +93,7 @@ if (config.isDevMode()) {
    * @return {Boolean}
    */
   async function hasManualFormatVariant(request, format) {
-    const path = request.originalUrl.replace('.html', `.${format}.html`);
+    const path = request.url.replace('.html', `.${format}.html`);
```
sebastianbenz: I'm remember there was a reason why I changed it to originalUrl

platform/lib/routers/pages.js
```diff
@@ -156,6 +143,15 @@ if (config.isDevMode()) {
   });
 
   pages.get('/*', async (request, response, next) => {
+    // Grow has problems delivering the index.html on a root request
```
sebastianbenz: can you extract this into a method as the exact same code is also used below

platform/lib/routers/pages.js
sebastianbenz: Did you test this with `run start:local` and `run develop` ?

platform/lib/routers/pages.js
matthiasrohmer: Good idea: went into `ensureFileExtension`.

platform/lib/routers/pages.js
matthiasrohmer: Now tested both, but the change would actually only have an effect for `if (config.isDevMode())` - checking for a manual variant does work a little bit differently for other environments 🙂

sebastianbenz: didn't notice, fine then

PR #1814, opened 2019-04-09
✨ Make header and footer available standalone.

matthiasrohmer:
This would make header and footer accessible via http://amp.dev/shared/header.html and http://amp.dev/shared/footer.html.

See the attached documents for reference, the final endpoints would be minified though.

Including [`base.scss`](https://github.com/ampproject/docs/blob/future/frontend/scss/base.scss) can possibly be omitted as you might have similar stylings for the blog in place already

[sample.zip](https://github.com/ampproject/docs/files/3061245/sample.zip)

.

### DIFF
--- a/pages/content/shared/footer.html
+++ b/pages/content/shared/footer.html
@@ -1,0 +1,20 @@
+$view: /layouts/blank.j2
+---
+<style amp-custom>
+  {% do doc.styles.addCssFile('css/base.css', -5) %}
+  {% do doc.styles.addCssFile('css/components/atoms/text.css', -3) %}
+  {% do doc.styles.addCssFile('css/components/atoms/anchor.css', -2) %}
+  {% do doc.styles.addCssFile('css/components/atoms/icon.css', -1) %}
+
+  {{ doc.styles.emit() }}
+</style>
+
+{% block icons %}
+<svg style="position: absolute; width: 0; height: 0; overflow: hidden;" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
+  <defs>
+    {{ doc.icons.emit() }}
+  </defs>
+</svg>
+{% endblock %}
+
+{% include 'views/partials/footer.j2' %}
--- a/pages/content/shared/footer.html
+++ b/pages/content/shared/footer.html
@@ -1,20 +1,3 @@
 $view: /layouts/blank.j2
 ---
-<style amp-custom>
-  {% do doc.styles.addCssFile('css/base.css', -5) %}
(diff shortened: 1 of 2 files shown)


sebastianbenz:
//cc @swissspidy does this for you?

pages/content/shared/footer.html
```diff
@@ -0,0 +1,20 @@
+$view: /layouts/blank.j2
+---
+<style amp-custom>
+  {% do doc.styles.addCssFile('css/base.css', -5) %}
+  {% do doc.styles.addCssFile('css/components/atoms/text.css', -3) %}
+  {% do doc.styles.addCssFile('css/components/atoms/anchor.css', -2) %}
+  {% do doc.styles.addCssFile('css/components/atoms/icon.css', -1) %}
+
+  {{ doc.styles.emit() }}
+</style>
```
felixarntz: It would be great if this style tag could be removed.

Eventually it would be good to have a separate CSS file with that content that we could also reference and include as `amp-custom`. This is not required for now though.

pages/content/shared/header.html
```diff
@@ -0,0 +1,21 @@
+$view: /layouts/blank.j2
+---
+<style amp-custom>
+  {% do doc.styles.addCssFile('css/base.css', -5) %}
+  {% do doc.styles.addCssFile('css/components/atoms/text.css', -3) %}
+  {% do doc.styles.addCssFile('css/components/atoms/anchor.css', -2) %}
+  {% do doc.styles.addCssFile('css/components/atoms/icon.css', -1) %}
+
+  {{ doc.styles.emit() }}
+</style>
```
felixarntz: Similarly, could this style tag be removed?

pages/content/shared/footer.html
```diff
@@ -0,0 +1,20 @@
...
+
+{% block icons %}
+<svg style="position: absolute; width: 0; height: 0; overflow: hidden;" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
+  <defs>
+    {{ doc.icons.emit() }}
+  </defs>
+</svg>
+{% endblock %}
```
felixarntz: Can this SVG markup be removed from here? It's the same I believe, so we only need that once (in `header.html`).

matthiasrohmer:
Thanks for the feedback! Stripped it down so that the endpoints will only really contain the required elements.

felixarntz:
@matthiasrohmer Sorry, I might not have clarified that before. We still need the SVG icons, but only once - could you bring that part back into `header.html`?

matthiasrohmer:
@felixarntz, I might have also read this more intent 😉 Readded the icons to `header.html`!

PR #1825, opened 2019-04-10
Add lorem ipsum descriptions to images in api/photo-stream

fstanis:


package.json
```diff
@@ -53,6 +53,7 @@
     "cors": "2.8.5",
     "express": "4.16.4",
     "js-yaml": "3.13.1",
+    "lorem-ipsum": "^2.0.1",
```
sebastianbenz: can you fix the version

examples/api/photo-stream.js
```diff
@@ -17,6 +17,9 @@
 
 const express = require('express');
 const {setNoCache} = require('@lib/utils/cacheHelpers');
+const {LoremIpsum} = require('lorem-ipsum');
```
sebastianbenz: I't be cool to have something that generates random text (which seems to make sense on the first look).

fstanis: I looked into https://www.npmjs.com/package/casual and https://www.npmjs.com/package/faker but it seems no different than https://www.npmjs.com/package/lorem-ipsum when it comes to descriptions.

Happy to replace later if we find a better random text provider. :)

PR #1835, opened 2019-04-10
✨ Rewrite GitHub relative paths when importing docs.

### LINKED ISSUE #1795: Clicking many linkable text led to 404
## Description of the issue found on ampproject.org

Please provide:

- Page link: https://amp.dev/documentation/guides-and-tutorials/learn/spec/amphtml.html
- Description of the documentation issue: clicking clickable texts like "contain the AMP boilerplate code" would end in 404 page

## Root cause:

The src all end with ".md", they should may be changed to ".html". Moreover, some files don't belong to current folder. Correct me if I'm wrong please!

1. Go to https://amp.dev/documentation/guides-and-tutorials/learn/spec/amphtml.html
2. Click any linkable text for another page, like "contain the AMP boilerplate code " 

What browser are you using?
Chrome

What O/S are you using?
Macbook

matthiasrohmer:
Fixes #1795.

### DIFF
--- a/platform/lib/pipeline/markdownDocument.js
+++ b/platform/lib/pipeline/markdownDocument.js
@@ -19,6 +19,7 @@ const yaml = require('js-yaml');
 const {Signale} = require('signale');
 const utils = require('@lib/utils');
 
+// Inline marker used by Grow to determine if there should be TOC
 const TOC_MARKER = '[TOC]';
 // It doesn't make sense to give every MarkdownDocument their own logger instance
 // therefore have one shared one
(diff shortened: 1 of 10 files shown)


platform/lib/pipeline/markdownDocument.js
```diff
@@ -181,6 +182,16 @@ class MarkdownDocument {
     return contents;
   }
 
+  /**
+   * Rewrite relative links and append the given base path to them
+   * @param  {String} contents
+   * @return {String}          The rewritten content
+   */
+  rewriteRelativePaths(base) {
+    this._contents = this._contents.replace(/\[([^\]]+)\]\((?!http|#)([^\)]+)\)/g,
```
sebastianbenz: I'd be feeling better doing this via a markdown parser and `new URL(string, base)`.

matthiasrohmer:
Properly, Magic? 😁No I see your point but can't really reliably estimate the effort. Because to be consistent we then should reimplement all other RegEx based `rewrite*` methods in `MarkdownDocument` but - and that's what makes this kind of a black box - in a way that python-markdown produces the same result as now.

If we want to go that route [marked.js](https://marked.js.org/#/USING_PRO.md#parser) seems to have a nice API which would allow modifying a parsed but not compiled document.

PR #1837, opened 2019-04-10
Heavily optimize the load and run time performance of Stories Best Practices guide

pbakaus:
This PR dramatically improves the load and run time performance of the stories best practices guide by:

- Re-compressing the mp4's with heavy compression
- Adding `webm` as primary format for browsers that support it (even better compression)
- Significantly reducing the quality and size of the poster images

This fixes most issues with the performance of that page.


pbakaus:
Argh this PR is messed up, will create a new one..

PR #1842, opened 2019-04-11
Fix sample listing in playground.

matthiasrohmer:
This brings back the template dropdown for the playground. Open questions:

1) Do we want to check in the built sitemap (`examples/static/samples/samples.json`)?
2) Do you have a good idea on how to "translate" the category identifier `interactivity-dynamic-content` to `Interactivity & Content`?
3) Currently the `SamplesBuilder` pushes all samples to the map, are there some that should be excluded besides the ones having `disablePlayground: true` set?


examples/static/samples/samples.json
```diff
@@ -0,0 +1 @@
...e":"DoubleClick AMP Story Ads","url":"http://localhost:8084/documentation/examples/advertising-analytics/doubleclick_amp_story_ads"},{"title":"Publisher Served AMP Story Ads","url":"http://localhost:8084/documentation/examples/advertising-analytics/publisher_served_amp_story_ads"}]},{"name":"multimedia-animations","examples":[{"title":"Supported media types in AMP Stories","url":"http://localhost:8084/documentation/examples/multimedia-animations/supported_media_types_in_amp_stories"}]},{"name":"style-layout","examples":[{"title":"AMP Story Layouts","url":"http://localhost:8084/documentation/examples/style-layout/amp_story_layouts"},{"title":"Desktop and Landscape Mode Support","url":"http://localhost:8084/documentation/examples/style-layout/desktop_and_landscape_mode_support"}]},{"name":"user-consent","examples":[{"title":"User Consent","url":"http://localhost:8084/documentation/examples/user-consent/story_user_consent"}]},{"name":"visual-effects","examples":[{"title":"AMP Story Animations","url":"http://localhost:8084/documentation/examples/visual-effects/amp_story_animations"},{"title":"Ken Burns","url":"http://localhost:8084/documentation/examples/visual-effects/ken_burns"}]}]}}
```
sebastianbenz: do we need to commit this file?

platform/lib/build/samplesBuilder.js
```diff
@@ -258,6 +269,54 @@ class SamplesBuilder {
...
   }
 
+  /**
+   * Adds a sample to the sitemap object for the playground to then render
+   * a list of available samples
+   * @param {Vinyl} sample       The file from which the sample is parsed
+   * @param {Object} parsedSample The parsed sample
+   */
+  _addToSitemap(sample, parsedSample) {
+    const format = this._getSampleFormat(parsedSample);
```
sebastianbenz: drafts shouldn't be added to the sitemap

sebastianbenz:
I should first read the comment and then the code...

1. no, this shouldn't be committed
2. If you use the [same naming scheme as in amp-by-example](https://github.com/ampproject/amp-by-example/blob/master/lib/FileName.js), you can access the name via `sample.category.name`.
3. yes, `draft: true`. Good point about `playgroundL true` that's something that wasn't implemented before.

platform/lib/build/samplesBuilder.js
```diff
@@ -137,6 +140,10 @@ class SamplesBuilder {
             return;
           }
 
+          if (!parsedSample.document.metadata.disablePlayground) {
```
sebastianbenz: also check for draft

platform/lib/build/samplesBuilder.js
```diff
@@ -181,8 +188,12 @@ class SamplesBuilder {
         reject(error);
       });
 
-      stream.on('end', () => {
+      stream.on('end', async () => {
         this._log.success('Built samples.');
+        // Only write samples sitemap if it has been a full samples build
+        if (!watch && config.options['clean-samples'] === true) {
+          await this._generateSitemap();
```
sebastianbenz: do we need to wait here?

matthiasrohmer: No good point!

matthiasrohmer:
Using the abe.com functionality to get the category name would have impacts from rearranging file/paths down to Grow - something I don't dare currently 😉 I used the other built-in functionality of defining meta information in the categories' `index.json` files.

Ready for final review now.

matthiasrohmer:
Works as before, basically:

Bildschirmfoto 2019-04-12 um 12 00 14

platform/lib/build/samplesBuilder.js
```diff
@@ -137,6 +140,11 @@ class SamplesBuilder {
             return;
           }
 
+          if (!parsedSample.document.metadata.disablePlayground &&
+          !parsedSample.document.metadata.drafts) {
```
sebastianbenz: nit: formatting

platform/lib/build/samplesBuilder.js
```diff
@@ -258,6 +272,57 @@ class SamplesBuilder {
...
+    this._sitemap[format] = formatCategories;
+  }
+
+  /**
+   * Takes what has been saved to this._sitemap and adds a sitemap.json to
+   * the gulp stream that is usable by the playground
+   * @type {Vinyl}
+   */
+  async _generateSitemap() {
+    for (const format of Object.keys(this._sitemap)) {
```
sebastianbenz: FYI `for (const [key, value] of Object.entries(this._sitemap)) {`

platform/lib/build/samplesBuilder.js
```diff
@@ -258,6 +272,57 @@ class SamplesBuilder {
...
+      const categories = Object.keys(this._sitemap[format]).map((category) => {
```
sebastianbenz: same as above

platform/lib/build/samplesBuilder.js
```diff
@@ -258,6 +272,57 @@ class SamplesBuilder {
...
+      });
+
+      this._sitemap[format] = {
+        'title': '',
+        'name': format,
+        'categories': categories,
+      };
+    }
+
+    await writeFileAsync(SITEMAP_DEST, JSON.stringify(this._sitemap));
```
sebastianbenz: why do we need to await here?

platform/lib/build/samplesBuilder.js
matthiasrohmer: Nice, thanks!

platform/lib/build/samplesBuilder.js
matthiasrohmer: Didn't pay attention, thanks!

PR #1844, opened 2019-04-11
🐛  fix component visual link overlaping on hover

### LINKED ISSUE #1840: Animated elements overlap on Stories landing page
https://amp.dev/about/stories on Chrome on desktop -

When "Build your first story" animates, it overlaps with the matching element to its right.

I also see a bunch of HTML classes that seem partially built to me, like they're missing prefixes. Maybe this is ok. But they look like this:

```HTML
<main class="ap--main -ib"><section class="ap--stage -tt"><div class="-ry"><div class="-tt"><div class="-rb"><div class="-rw ap-a-ico"><svg><use xmlns:xlink="http://www.w3.org/1999/xlink" xlink:href="#bolt-solid"></use></svg></div><h2 class="-rk">AMP Stories</h2><h1 class="-rx"
```

keevkeev:
fixes  #1840

### DIFF
--- a/frontend/scss/components/organisms/component-visual.scss
+++ b/frontend/scss/components/organisms/component-visual.scss
@@ -25,7 +25,7 @@
   }
 
   &-link {
-    margin: 3em 7px 0;
+    margin: 3em 1em 0 0;
   }
 
   &-image {
--- a/frontend/scss/components/organisms/component-visual.scss
+++ b/frontend/scss/components/organisms/component-visual.scss
@@ -25,7 +25,7 @@
   }
 
   &-link {
-    margin-top: 3em;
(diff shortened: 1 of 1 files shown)


frontend/scss/components/organisms/component-visual.scss
```diff
@@ -25,7 +25,7 @@
   }
 
   &-link {
-    margin-top: 3em;
+    margin: 3em 7px 0;
```
matthiasrohmer: I know that's really picky, but can we use consistent units here? 😁

keevkeev: fixed 😈

PR #1845, opened 2019-04-11
♻️ Make header keyboard accessible.

matthiasrohmer:
Handling one of the most prominent answers from the beta form: the flyout navigation and the format toggle is now fully keyboard accessible and are built without any `input` hacks.

This also leads to the discussed change that the top level navigation items that have a flyout aren't clickable on their own anymore.


sebastianbenz:
Nice!

PR #1848, opened 2019-04-11
Typo fix for pin widget

kentbrew:
`data-do` should be `embedPin`, not `embedPing`.


kentbrew:
Abandoning because I can't see how to get it to check on my freshly-completed CLA.

PR #1853, opened 2019-04-11
Update single video template

calebcordry:
There were a couple of bugs. Added script for amp-video, and some css tweaks.

@ithinkihaveacat can you make sure I am uploading these to the correct spot? Thanks!

cc/ @jasti


ithinkihaveacat:
(https://github.com/ampproject/docs/issues/1862 is the issue to update READMEs.)

PR #1858, opened 2019-04-12
🐛 Use double opt-in newsletter link.

matthiasrohmer:


matthiasrohmer:
Let's directly do this the right way.

PR #1874, opened 2019-04-12
✨ Add shortcode for format filtering.

### LINKED ISSUE #1748: Create shortcode for format-filtered paragraphs
It would be nice to have a shortcode that allows to filter certain paragraphs for specific formats. E.g.

```
[filtered format="stories"]
This is only shown when the document is filtered for stories.
[/filtered]
```

/cc @pbakaus @CrystalOnScript

matthiasrohmer:
Fixes #1748.

```markdown
[filter formats="websites"]
This is only visible for [websites](?format=websites).
[/filter]

[filter formats='websites, email']
This is visible for [websites](?format=websites) & [email](?format=email).
[/filter]

[filter formats="stories"]
This is visible for [stories](?format=stories).
[/filter]
```

### DIFF
--- a/frontend/templates/layouts/default.j2
+++ b/frontend/templates/layouts/default.j2
@@ -1,3 +1,14 @@
+{% macro filter(formats='websites, ads, stories, email') -%}
+{% if not formats.strip() %}
+{{ caller() }}
+{% else %}
+{% set formats = formats.split(',') %}
+<p class="{% for format in formats %}ap--{{ format.strip() }} {% endfor %}">
+{{ caller() }}
+</p>
+{% endif %}
+{%- endmacro %}
+
 {% macro tip(title, type='note') -%}
 {% with %}
 {% set text = caller() %}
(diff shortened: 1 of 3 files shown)


sebastianbenz:
//cc @fstanis

frontend/templates/layouts/default.j2
```diff
@@ -1,3 +1,14 @@
+{% macro filter(formats='websites, ads, stories, email') -%}
+{% if not formats.strip() %}
+{{ caller() }}
+{% else %}
+{% set formats = formats.split(',') %}
+<p class="{% for format in formats %}ap--{{ format.strip() }} {% endfor %}">
```
pbakaus: Should this be a `<div>` instead? `<p>` could produce side effects, as they come with margins, and nested elements could be weird..

matthiasrohmer: Had a `div` before but Grow didn't add `p`s then which *made* it look weird 😁I'll test again...

matthiasrohmer: `\n` to the rescue. Now `div` is the wrapping element and Grow gracefully adds `p` - too many as always, but... 🙂

PR #1881, opened 2019-04-13
👷 Enable continuous integration

matthiasrohmer:
That was quite fun, sorry it grew so big. A brief overview of what I did so far:

This PR reenables Travis with three stages in total: 
1. Setup: Lints node, imports all needed documents, builds playground, boilerplate generator and samples, then checks for broken references in the Grow pod, if nothing failed until then uploads all built artifacts to GCS to not have to repeat those steps for the split Grow build
2. Build: Fetches artifacts uploaded by previous stage from GCS, builds Grow locales split to four VMs with three locales each, minifies the pages and uploads its results to GCS. I currently left out validating the pages as we have two faulty ones which would break the build.
3. Deploy: Fetches all artifacts from previous stages, collects statics files and runs actual deploy to staging

The changes done for stage two also enables the local command `npm run build:staging -- --locales en,es`- so no need to manually strip languages from `podspec.yaml` for a test build anymore 😁 

Collecting the static files was so deeply married into the JS build pipeline that I needed to extract it in order to not do it multiple times. While doing so I took the base from #1632 which enables virtual ZIPs.

I've extended the [/who-am-i](https://amp-dev-staging.appspot.com/who-am-i) endpoint to see which commit actually is live. It now returns something like

```json
{
  "environment": "staging",
  "build": {
    "number": 3522,
    "timestamp": "2019-04-13 20:25:23 UTC",
    "commit": {
      "sha": "90c610d6c5a7933fcd36375963b811e00360456c",
      "message": "🐛 Keep sample sitemap when advancing to build stage"
    }
  }
}
```

Currently this only handles builds for staging. Details for production are greatly appreciated, @sebastianbenz 😉 I know you need to build the docker image by using what's in `gulpfile.js` but from there I am basically lost. A prod key for GCloud also needs to be added to Travis. 

Last task I can think of is then to re-add the correct conditions (only deploy on `future` (?), ...) to the configuration.

Last but not least: this is already successfully building towards amp-dev-staging.appspot.com. For example see https://travis-ci.org/ampproject/docs/builds/519767994.


.gcloudignore
```diff
@@ -29,6 +29,7 @@ packager/
 /frontend/
 pages/
 !platform/pages
+/artifacts
```
sebastianbenz: can you rename this to `build` so that we have a single place for all build artifacts.

.travis.yml
```diff
@@ -3,23 +3,58 @@ node_js:
...
-  provider: gae
-  keyfile: ".staging.key.json"
-  project: amp-dev-staging
-  skip_cleanup: true
-  on:
-    branch: future
 before_install:
 - openssl aes-256-cbc -K $encrypted_d4a09416a845_key -iv $encrypted_d4a09416a845_iv
   -in .staging.key.json.enc -out .staging.key.json -d
+- gcloud auth activate-service-account --key-file=.staging.key.json
```
sebastianbenz: curios: why is this step needed now?

scripts/build/setup.sh
```diff
@@ -0,0 +1,84 @@
+#!/bin/bash
```
sebastianbenz: Can you please turn this into gulp tasks. Use the `sh` module if needed.

scripts/deploy.sh
```diff
@@ -0,0 +1,52 @@
...
+    unzip -o -q -d . $filename
+  done
+  echo -e "travis_fold:end:fetch\n"
+fi
+
+# Collect static files as they haven't been needed for the build itself
+node $root/platform/lib/build/staticsCollector.js
+
+# Persist Travis information in config to identify build via /who-am-i
+if [ -n "$TRAVIS_BUILD_NUMBER" ]; then
```
sebastianbenz: this should be created for every build, not only on travis.

scripts/deploy.sh
```diff
@@ -0,0 +1,52 @@
+#!/bin/bash
```
sebastianbenz: Can you also turn this into gulp tasks. It's ok to move parts into a shell script if needed.

platform/lib/build/staticsCollector.js
```diff
@@ -0,0 +1,161 @@
+/**
```
sebastianbenz: This file should also be a gulp task.

platform/lib/config.js
```diff
@@ -28,6 +28,20 @@ const GROW_OUT_DIR = utils.project.absolute('platform/pages');
 
 const ENV_DEV = 'development';
 const ENV_PROD = 'production';
+const AVAILABLE_LOCALES = [
```
sebastianbenz: We should then remove all available locales from the pod spec to avoid duplication.

.travis.yml
matthiasrohmer: Had auth errors for the `gsutil cp ...` calls otherwise.

scripts/deploy.sh
matthiasrohmer: I agree, shall we set `build.number` to `null` for manual builds then? The commit information would be available in all envs.

matthiasrohmer:
Thanks for taking a first look! All your suggestions make perfect sense, I hope to accommodate them somewhen over the next days.

Sharding per locale sadly doesn't have that much of a effect. See: [#3533 (33 min 14 sec)](https://travis-ci.org/ampproject/docs/builds/519767994) vs. [#3524 (43 min 3 sec)](https://travis-ci.org/ampproject/docs/builds/519767994).

sebastianbenz:
> Sharding per locale sadly doesn't have that much of a effect. See: #3524 (43 min 3 sec) vs. #3533 (33 min 14 sec).

a 25% improvement is a lot!

scripts/deploy.sh
sebastianbenz: sgtm

.travis.yml
```diff
@@ -3,23 +3,71 @@ node_js:
...
+      script: unbuffer gulp buildPages --locales pt_BR
+    - <<: *build
+      name: 'Build Pages (RU)'
+      script: unbuffer gulp buildPages --locales ru
+    - <<: *build
+      name: 'Build Pages (TR)'
+      script: unbuffer gulp buildPages --locales tr
+    - <<: *build
+      name: 'Build Pages (CN)'
+      script: unbuffer gulp buildPages --locales zh_CN
```
sebastianbenz: Is it possible to auto generate travis jobs? I don't like hard-coding the languages here.

gulpfile.js/build.js
```diff
@@ -0,0 +1,254 @@
...
+
+
+/**
+ * Builds playground and boilerplate generator, imports all remote documents,
+ * builds samples, lints Grow pod and JavaScript.
+ *
+ * @return {Promise}
+ */
+async function setupBuild() {
+  await sh('npm run lint:node');
```
sebastianbenz: Can you turn this into it's own gulp task and then use sequential.

gulpfile.js/build.js
```diff
@@ -0,0 +1,254 @@
...
+
+  // Those two are built that early in the flow as they are fairly quick
+  // to build and would be annoying to eventually fail downstream
+  await Promise.all([buildPlayground(), buildBoilerplate()]);
```
sebastianbenz: Use gulp sequential (combined with parallel)

gulpfile.js/build.js
```diff
@@ -0,0 +1,254 @@
...
+
+  await Promise.all([buildSamples(), importAll()]);
```
sebastianbenz: Use gulp sequential (combined with parallel)

gulpfile.js/build.js
```diff
@@ -0,0 +1,254 @@
...
+
+  // Grow can only be linted after samples have been built and possibly linked
+  // to pages have been imported
+  // TODO: Reenable after false-positives have been fixed
+  // await sh('npm run lint:grow');
```
sebastianbenz: same as above

platform/lib/pipeline/growReferenceChecker.js
```diff
@@ -79,9 +79,16 @@ class GrowReferenceChecker {
       stream.pipe(gulp.dest('./'));
 
       stream.on('end', () => {
-        this._log.complete(`Finished fixing. A total of ${this._brokenReferencesCount} had ` +
-        `errors. ${this._unfindableDocuments.length + Object.keys(this._multipleMatches).length}` +
-        'still have.');
+        if (Object.keys(this._multipleMatches).length == 0
+            && this._unfindableDocuments.length == 0) {
+          this._log.success('All references intact!');
+          process.exit(0);
```
sebastianbenz: I think it's better to let the caller decide how to handle the success case. Otherwise it's impossible to call this from a gulp task.

.travis.yml
matthiasrohmer: I agree. There's no way to seed jobs from a different source than .travis.yml so I thought about using a [build matrix](https://docs.travis-ci.com/user/build-matrix/) but that [didn't work out](https://travis-ci.org/ampproject/docs/builds/532377988). Somehow the expansion always gets assigned to the Test stage.

gulpfile.js/build.js
```diff
@@ -0,0 +1,411 @@
...
+function buildFrontend(done) {
+  return (gulp.parallel(sass, templates, icons))(done);
+}
+
+/**
+ * Builds the playground
+ * @return {Promise}
+ */
+async function buildPlayground() {
+  await sh('mkdir -p playground/dist');
```
matthiasrohmer: This is a dumb fix, but strangely the `clean-webpack-plugin` suddenly [failed](https://travis-ci.org/ampproject/docs/jobs/532013888) when the `playground/dist` wasn't there.

```
/home/travis/build/ampproject/docs/node_modules/clean-webpack-plugin/dist/clean-webpack-plugin.js:205
      throw error;
      ^
Error: ENOENT: no such file or directory, stat '/home/travis/build/ampproject/docs/playground/dist'
    at Object.statSync (fs.js:850:3)
```

Couldn't reproduce it locally though. You've got an idea or maybe a package in mind that we updated recently and might cause this?

gulpfile.js/build.js
```diff
@@ -0,0 +1,411 @@
...
+}
+
+
+/**
+ * Runs all importers
+ *
+ * @return {Promise}
+ */
+function importAll() {
+  return Promise.all([
```
matthiasrohmer: Wanted to migrate this to `gulp.parallel` as you suggested but it seems to fiddle with the scope of the functions which made them fail.

gulpfile.js/build.js
```diff
@@ -0,0 +1,411 @@
...
+  }
+}
+
+/**
+ * Transforms already built pages and does so while spawning multiple child
+ * processes to speed up processing
+ *
+ * @return {Promise}
+ */
+async function transformPages() {
```
matthiasrohmer: This task got a bit ugly as I tried to introduce sharding. I thought of it having a bigger effect, but it only saves ~10 seconds per locale. But the Travis VMs also only run 2 cores.

.travis.yml
```diff
@@ -3,23 +3,76 @@ node_js:
...
+  - setup
+  - name: build
+    if: branch = future
+  - name: deploy
+    if: branch = future
+jobs:
+  include:
+    - stage: setup
+      name: 'Check code, import documents, build samples, playground & boilerplate'
+      script: unbuffer gulp setupBuild
```
matthiasrohmer: A more specialized testing task for PRs might make sense to speed up things. Importing the documents is needed to check document references but could maybe get cached somehow.

platform/config/podspec.yaml
```diff
@@ -4,19 +4,6 @@ home: /content/amp-dev/index.html
 localization:
   root_path: /{locale}/
   default_locale: en
```
sebastianbenz: Did you intentionally remove all languages?

platform/lib/utils/travis.js
```diff
@@ -0,0 +1,67 @@
...
+
+const folds = {};
+
+/**
+ * Outputs markers needed for the Travis UI to make certain log sections
+ * collapsable
+ * @param  {String} label [description]
+ * @return {undefined}
+ */
+function fold(label) {
```
sebastianbenz: nice!

gulpfile.js/build.js
sebastianbenz: No idea

gulpfile.js/build.js
```diff
@@ -0,0 +1,411 @@
...
+  const shardId = config.options.shard;
+  const startIndex = shardId * shardPathCount;
+  const endIndex = shardId == shardCount - 1 ? paths.length : (shardId + 1) * shardPathCount;
+  // As slice omits the edge index include it explicitly
+  paths = paths.slice(startIndex, endIndex);
+
+  signale.await(`Shard ${shardId} [${startIndex} - ${endIndex}]: \
+      processing ${paths.length} files ...`);
+  // After the pages have been built by Grow create transformed versions
+  await new Promise((resolve, reject) => {
```
sebastianbenz: Is the await needed here?

gulpfile.js/build.js
```diff
@@ -0,0 +1,411 @@
...
+
+exports.clean = clean;
+exports.importAll = importAll;
+exports.buildPlayground = buildPlayground;
+exports.buildBoilerplate = buildBoilerplate;
+exports.buildFrontend = buildFrontend;
+exports.buildSamples = buildSamples;
+exports.buildPages = buildPages;
+
+exports.setupBuild = setupBuild;
```
sebastianbenz: Can we find a better name for setupBuild?

gulpfile.js/build.js
```diff
@@ -0,0 +1,411 @@
...
+exports.transformPages = transformPages;
+exports.fetchArtifacts = fetchArtifacts;
+exports.build = gulp.series(fetchArtifacts, gulp.parallel(buildSamples, buildFrontend),
+    buildPages);
+exports.collectStatics = collectStatics;
+exports.finalizeBuild = gulp.parallel(fetchArtifacts, collectStatics, persistBuildInfo);
+
+exports.fullBuild = gulp.series(clean, gulp.parallel(setupBuild, buildFrontend), buildPages,
```
sebastianbenz: I'd prefer this to be named `build`

gulpfile.js/build.js
sebastianbenz: That's really surprising. What's the bottleneck?

gulpfile.js/build.js
matthiasrohmer: You've got a suggestion? Going with the general > specific naming scheme `buildInit`, `buildBootstrap`, `buildStart` could maybe work. I am not really happy with `setupBuild` either.

.travis.yml
sebastianbenz: OK, let's keep it like this for now then.

gulpfile.js/build.js
sebastianbenz: how about `buildSlow` and `buildFast`? At least that looks like the separator between these two tasks. It also might make sense to split tasks into `internal` and `public` (if that makes sense).

README.md
```diff
@@ -24,26 +24,31 @@ that makes a merge over to the original repository beneficial.
     ```
 
 ## Develop
-To start developing start a pipeline that takes care of building the frontend, collecting all needed files and start Grow when finished. The pipeline then watches the sources files for changes and rebuilds them if needed.
+If it's your first time working on amp.dev it is recommended to bootstrap your local environment. To do so make sure you have setup a valid [GitHub access token](https://github.com/settings/tokens) in an environment variable named `AMP_DOC_TOKEN` like:
```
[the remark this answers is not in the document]
matthiasrohmer: Absolutely! We could fairly easily: the remainings of SpecImporter are already committed anyway. We'd just need to  commit/unignore `pages/content/amp-dev/documentation/components/reference/*.md` as well.

platform/config/podspec.yaml
matthiasrohmer: As per an earlier review: https://github.com/ampproject/docs/pull/1881/files#r275148312 😁 You correctly made the point that they are duplicated by https://github.com/ampproject/docs/blob/8daff99905fe8af1463e7a4e6263323aeaa6ff59/platform/lib/config.js#L30

sebastianbenz: hah! I thought pod spec.yaml would be the single source of truth, but the other way around is fine as well.

README.md
```diff
@@ -52,16 +57,23 @@ To run a local test build that does all the minifying and vends the static pages
...
 ```sh
-$ npm run build:local
+$ gulp fullBuild --env local
 $ npm run start:local
 ```
 
-## Build & Deploy
-To build and deploy to the staging environment on [Google App Engine](https://cloud.google.com/appengine/) run the following
+## Build
+Prior running a build it is recommended to run `gulp clean` to ensure your environment doesn't contain any remainings of previous builds - additionally check your working copy for eventual unintended local changes.
```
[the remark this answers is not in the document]
matthiasrohmer: Update the task and addd words of caution.

README.md
```diff
@@ -52,16 +57,23 @@ To run a local test build that does all the minifying and vends the static pages
 proxying them through to Grow you can run
 
 ```sh
-$ npm run build:local
+$ gulp fullBuild --env local
```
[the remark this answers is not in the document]
matthiasrohmer: As per a later comment it is now `build`.

gulpfile.js/build.js
matthiasrohmer: Not quite: `setupBuild` runs various smaller tasks in the first place for the sharded instances to not need to run them again.

I went with renaming it to `buildPrepare`.

pbakaus:
wow am I relieved to see this beast merged! congrats! :)

PR #1891, opened 2019-04-15
⏪ Remove Grow URL monkey patch

matthiasrohmer:
... as the old way of patching isn't compatible with the new Grow version we only tidy the URL for the two crucial places: sitemap.xml and canonicals.


matthiasrohmer:
Will set up a debugging Grow pod to get some background on possible environment differences.

Issue #1898, opened 2019-04-16
keyframe animations broken on Safari


keevkeev:
#### rolling text animation on the home page stacks websites and stories text on Safari

Please provide:

- Page link: https://amp.dev

#### visit amp.dev on Safari

What browser are you using?
- Safari
What O/S are you using?
- iOS, macOs High Sierra

keevkeev:
fixed by #1905

matthiasrohmer:
Sadly not a fix that works consistently 🙁Would be great if you could have another look!

matthiasrohmer:
You might also want to inspect the unoptimized version (https://amp.dev/index.amp.html) vs the unoptimized one (https://amp.dev/) - the first one seems to behave slightly better on Safari. Or at least only breaks every fifth reload instead of almost every one.

matthiasrohmer:
Hmpf ☹️ still broken for me. Reload https://amp-dev-staging.appspot.com several times in Safari.

keevkeev:
@matthiasrohmer which safari version are you using?

ithinkihaveacat:
I can reproduce this too on amp-dev-staging, in both the optimized and non-optimized versions. Latest Safari mobile.

matthiasrohmer:
Thanks @ithinkihaveacat for cross-checking! @keevkeev I am testing with Version 12.1 (14607.1.40.1.4) macOS.

keevkeev:
changing title to 'keyframe animations broken on Safari' as all animations triggered with keyframes are broken sometimes on initial load

keevkeev:
@matthiasrohmer, this issue cannot be replicated in a development environment as all keyframe animations are loading inline, however at build time we are extracting all keyframe animations into a separate style sheet and injecting it at the bottom of the body. My theory is that the content is that all the css and content is being loaded before the keyframe animation stylesheet at build time which causes this issue. All keyframe animations are broken on the page, not just the rolling text in the stage. As I don't have access to deploying to the staging environment I will need someone to else to run a build and test this.

PR #1905, opened 2019-04-16
🍏  fix text overlap for rolling animation

keevkeev:
Safari triggers rolling animation sooner than other browswers on inital page load, adding an inital animation delayprevents text from overlapping #1898


keevkeev:
you're welcome

Issue #1902, opened 2019-04-16
The page should respect the `prefers-reduced-motion` user preference


tomayac:
## Description of the issue found on ampproject.org

- Page link:
  https://amp.dev/
- Description of the documentation issue:

  The "levitate" animation on the homepage defined in [case-band.scss](https://github.com/ampproject/docs/blob/698bbabacff10a1bd966e4e8272d94ff6fa06acd/frontend/scss/components/organisms/case-band.scss) should respect the [`prefers-reduced-motion`](https://developers.google.com/web/updates/2019/03/prefers-reduced-motion#working_with_the_media_query) user preference.

CC: @sebil.

PR #1906, opened 2019-04-16
♿️ added prefers reduced motion for caseband

keevkeev:
Caseband animation is a bit heavy and some users may prefer to disable this animation. Media query is added to allow this.  #1902


keevkeev:
closes #1902

tomayac:
Thanks for the incredibly fast turn-around time! Massive applause for everyone involved 👏 on behalf of all our `prefers-reduced-motion` users!

PR #1908, opened 2019-04-17
🐛 Stop injecting unused deps into home.

matthiasrohmer:
While doing the live quickcheck I just noticed that the dependency injector adds the unused extensions `amp-ad`, `amp-sidebar` to the homepage.

This strangely isn't picked up by the validator.


matthiasrohmer:
https://stackoverflow.com/a/1732454 ☺️

PR #1913, opened 2019-04-17
✨ Add first season of video courses to landing page.

matthiasrohmer:
@morsssss, I anticipated you might want to have the video course now added to the landing page, was that correct? 😉Let me know if something else is urgently missing.

(Sorry for the unrelated pipeline fix commit, that bothered my for a while.)

PR #1915, opened 2019-04-17
Include videos in courses page, other minor changes

morsssss:
Include area to showcase our new coursework videos!
Change name and order of courses landing page in sidebar

### DIFF
--- a/pages/content/amp-dev/documentation/courses/index.html
+++ b/pages/content/amp-dev/documentation/courses/index.html
@@ -1,7 +1,8 @@
 ---
-$title: Learn AMP
+$title: Learn AMP with free courses
 $titles:
   navigation: Courses
+  sidebar: Learn AMP
   teaser: Courses
 $view: /views/overview/courses.j2
 $path: /documentation/courses/index.html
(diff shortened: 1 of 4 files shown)


pages/content/amp-dev/documentation/courses/index.html
```diff
@@ -1,13 +1,13 @@
 ---
-$title: Learn AMP with free courses
```
matthiasrohmer: I actually liked that this showed what kind of resource the user can expect over other potential SERP items 🙂Were there specific reasons going for just "Learn AMP"?

morsssss:
@matthiasrohmer , just added here the CSS tweak from #1913 .

pages/content/amp-dev/documentation/courses/index.html
morsssss: Yes - I thought the previous title looked too long to me in the sidebar - too much like an ad. I hadn't realized that this is what also appears in the SERP. Of course, it is. Don't suppose there's a way to get one thing in the sidebar and another in the `<title>` tag?

matthiasrohmer:
I am more than okay with my changes being submitted!

PR #1917, opened 2019-04-17
🎨 Verify release tag prior build.

matthiasrohmer:
Not to sure if this makes sense: to prevent something similar as this morning happening we could append the current timestamp to build tags. Otherwise creating a new build based on the same commit without giving an explicit tag wouldn't be possible.


matthiasrohmer:
I see. Added a `verifyTag` task!

matthiasrohmer:
🤦‍♂️Thanks for taking such a thorough look!

PR #1932, opened 2019-04-21
Fix/packager proxy

sebastianbenz:


sebastianbenz:
This is no longer need due to different serving setup.

PR #1936, opened 2019-04-21
add lru cache for recently used pages

sebastianbenz:


platform/lib/routers/pages.js
```diff
@@ -160,7 +165,7 @@ if (config.isDevMode()) {
     }
   });
 
-  pages.get('/*', async (request, response, next) => {
+  pages.use('/*', async (request, response, next) => {
```
matthiasrohmer: Is going back to `use` intentional?

PR #1941, opened 2019-04-21
♻️ Use req.url instead of req.path for cache and log

matthiasrohmer:
```
[CACHE] miss /documentation/components/?format=stories
cache count 1
[TIMING]: /documentation/components/?format=stories: 0.114s
```


matthiasrohmer:
Closed in favour of #1942.

PR #1947, opened 2019-04-22
♻️ Statically format filter pages.

matthiasrohmer:
Filters to `pages-transformed` currently to be able to retest.


platform/lib/build/pageTransformer.js
```diff
@@ -107,23 +110,111 @@ class PageMinifier {
           let html = canonicalPage.contents.toString();
           html = scope.minifyPage(html, canonicalPage.path);
 
-          const ampPath = canonicalPage.relative.replace('.html', '.amp.html');
-          const optimizedHtml = await scope.optimize(html, ampPath);
```
matthiasrohmer: Needs to be reactivated.

platform/lib/build/pageTransformer.js
```diff
@@ -107,23 +110,111 @@ class PageMinifier {
...
 
+          for (const page of [canonicalPage, ampPage, ...filteredPages]) {
+            this.push(page);
+          }
+
+          scope._log.success(`Transformed ${canonicalPage.path}`);
           callback();
         }))
-        .pipe(gulp.dest('./'));
+        .pipe(gulp.dest('./pages-transformed'));
```
matthiasrohmer: Needs to be `./` again.

platform/lib/routers/pages.js
```diff
@@ -231,27 +216,26 @@ if (!config.isDevMode()) {
...
-        // Check if there's a manually filtered variant ...
-        const manualRequestPath = requestPath.replace('.html', `.${format}.html`);
+        // Check if there's a manually filtered variant and respect AMP variant
+        if (requestPath.endsWith('.amp.html')) {
+          const manualRequestPath = requestPath.replace('.amp.html', `.${format}.amp.html`);
+        } else {
+          const manualRequestPath = requestPath.replace('.html', `.${format}.html`);
+        }
+
         if (await fs.existsSync(utils.project.pagePath(manualRequestPath))) {
```
sebastianbenz: use `existsAsync` (defined in the same file)

platform/lib/common/filteredPage.js
```diff
@@ -43,10 +43,11 @@ const FILTERED_ROUTES = [
  */
 function filterPage(format, dom, force) {
   if (!isAvailable(format, dom) && !force) {
-    return;
+    return false;
```
sebastianbenz: return either dom as well for consistency or return true at the end

platform/lib/build/pageTransformer.js
```diff
@@ -141,14 +141,16 @@ class PageTransformer {
...
-    // Skip pages that have been manually filtered and therfore have a path like
+    // Skip pages that have been manually filtered and therefore have a path like
     // - guides-and-tutorials/index.websites.html
     // - guides-and-tutorials/index.email.amp.html
-    if (path.match(/\.(websites|stories|ads|email)(\.amp)?\.html/)) {
-      return true;
+    // But still the matching format should be filtered, therefore pass it back
+    const format = path.match(/\.(websites|stories|ads|email)(\.amp)?\.html/);
+    if (format) {
+      return format[1];
```
sebastianbenz: Is it intentional that only this path returns a format?

sebastianbenz: vs true/false

platform/lib/build/pageTransformer.js
```diff
@@ -116,15 +119,108 @@ class PageMinifier {
...
+    return filteredPages;
+  }
+
+  /**
+   * @param  {String} html
+   * @return {String}
+   */
+  filterHtml(html, format, force) {
+    const dom = cheerio.load(html);
+    if (filterPage(format, dom, force)) {
```
sebastianbenz: nit: using  an early exit avoids indentation

sebastianbenz: it's also weird that only one code path returns a value

platform/lib/build/pageTransformer.js
matthiasrohmer: Yes and no, those should have been two separate methods as now.

platform/lib/build/pageTransformer.js
matthiasrohmer: `filterPage` now returns `true|false`.

platform/lib/build/pageTransformer.js
```diff
@@ -116,14 +120,123 @@ class PageMinifier {
...
+        continue;
+      }
+
+      for (const format of FORMATS) {
+        if (this._hasManualFiltered(page, format)) {
+          continue;
+        }
+
+        const filteredHtml = this.filterHtml(html, format);
+        if (filteredHtml) {
```
sebastianbenz: this is always true as you're now returning the input html string.

matthiasrohmer: Thanks, returns `undefined` for unfiltered HTML again now, which is already gracefully handled by both `pageTransformer.js` and `pages.js` router.

PR #1975, opened 2019-04-24
Shortlink for courses

morsssss:


morsssss:
True - so far, it looks like the other shortlinks are verbs.

And of course, this shortlink would really go to the "courses" area of the site!

On the other hand, I thought "learn" made for a snazzier, more memorable shortlink with a better implied call to action. That's why I preferred it. But if you prefer "courses" I will not object!

pbakaus:
"learn" it is!

Issue #1694, opened 2019-04-02
Add SSR for category filtering


matthiasrohmer:
To be able to link to a specific component, example and soon guides category we should add SSR to the currently `amp-bind` base category filtering on the overview pages.

PR #2015, opened 2019-05-05
Overhauled tools section

pbakaus:
This PR is a complete redesign and expansion of the Tools section on amp.dev. WIP.


pbakaus:
@matthiasrohmer would be cool if you could review this one at your earliest convenience, I'd like to get this page up and running for I/O, so need to get it deployed on Monday.. a little bit of a rush job but tried to conform to amp.dev code style as much as possible. Any improvements or feedback welcome!

pbakaus:
@matthiasrohmer thanks! I just made some tweaks, might want to re-screenshot (I mostly changed the font size of the paragraph inside the cards, feels better now, still low information density but more workable). On filter bubbles: not a bad idea, but the dropdown isn't as visible than these. However, I definitely would like the ability to link to a section that is pre-filtered. If you can help with this and try it out, that would be awesome.

pbakaus:
@matthiasrohmer note for development, if you want to make changes, you can just push to the new-tools branch.

matthiasrohmer:
I can't think of a way that would quickly enable SSR for this page with what we currently have. The format-filtering works very specifically with the documentation pages - and also in a destructive manner 🙁

pbakaus:
@matthiasrohmer mhh.. if I wanted to attempt this today, where would I start? Also, what do you mean by destructive?

matthiasrohmer:
@pbakaus, destructive in a way that the statically filtered documents actually loose the information needed for an unfiltered state. That means the dynamic filtering you've built using `amp-bind` - which is nice - wouldn't work anymore but would need to link to another statically filtered page.

The logic for the static filtering is contained in `platform/lib/common/filteredPage.js`. If losing the dynamic filtering for now is an okay compromise you could just use the normal filter classes (`ap--websites`, `ap--stories`, ...) to mark filtered elements, enable the tools page in FILTERED_ROUTES and then replace the bound buttons with anchors to `?format=websites`, ...

In the long run we've already planned to head for a more sophisticated filtering logic based on SSR/templating - also needed for things like #1694 and to increase overall build speed compared to the DOM implementation.

Flagging @sebastianbenz, to check if he has any other quick ideas. Otherwise let me know if I can help somehow.

sebastianbenz:
+1 to what Matthias suggested. I'd go with SSR'd format filter and not use client-side filtering.

PR #2042, opened 2019-05-08
🔗Add link to ad manager docs

calebcordry:
@jasti and I thought it was a good idea to add a link here, since publishers are often uploading the ads on behalf of the advertiser.


calebcordry:
Works for me, thanks!

PR #2047, opened 2019-05-09
💄Update boilerplate generator theme

### LINKED ISSUE #1697: Update Boilerplate Generator theme
Buttons and code snippets are no longer up-to-date. 

Screenshot 2019-04-02 at 17 48 56

lluerich:
fixes #1697

### DIFF
--- a/boilerplate/templates/styles/code-snippet.scss
+++ b/boilerplate/templates/styles/code-snippet.scss
@@ -1,0 +1,80 @@
+/*
+Monokai style
+*/
+
+.hljs {
+  display: block;
+  overflow-x: auto;
+  padding: 0.5em;
+  background: #272822; color: #ddd;
+}
+
+.hljs-tag {
+  color: #f8f8f2;
+}
+
+.hljs-number {
+  color: #ae81ff;
+}
+
+.hljs-keyword,
+.hljs-selector-tag,
+.hljs-literal,
+.hljs-strong,
+.hljs-name {
+  color: #f15a5a;
+}
+
+.hljs-code,
+.hljs-attribute {
+  color: #66d9ef;
+}
+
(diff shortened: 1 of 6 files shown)


lluerich:
Forgot to add screenshots: 
boilerplate-generator-1
boilerplate-generator-2

boilerplate/templates/styles/code-snippet.scss
```diff
@@ -0,0 +1,80 @@
+/*
+Monokai style
```
sebastianbenz: I think a better place for this would be `/frontend/scss/`.

sebastianbenz: It would indeed be a better place, but I forgot that the boilerplate generator uses it own build. So, please ignore and leave things as they are :-).

lluerich: So should I revert the changes completely or just leave them in the state I submitted yesterday? :-)

lluerich:
I changed the behaviour of the dropdown to only open on click, because I saw no way to have both the hover effect and let the dropdown collapse once something is selected or the selected item is clicked.

sebastianbenz:
> I changed the behaviour of the dropdown to only open on click, because I saw no way to have both the hover effect and let the dropdown collapse

👍


Two other things I've noticed now that the hover is gone, can you please:

1. increase the button height to 2rem
2. increase the font-size inside the button to 16px

We also should take a look at the normal format switcher //cc @matthiasrohmer

sebastianbenz:
Thanks! ...and I'm sorry, but I forgot two other things:

1. Can you set the max-width for the format toggle to 240px
2. the caret needs to reflect the collapsed state

sebastianbenz:
Perfect! Thanks!

PR #2048, opened 2019-05-09
It is the last multipart that needs to end with double dashes

kevinsimper:
https://en.wikipedia.org/wiki/MIME#Multipart_messages
> The last boundary must have two hyphens at the end.


sebastianbenz:
@fstanis ping

fstanis:
LGTM, thank you!

PR #2050, opened 2019-05-10
Update product page add-to-cart API endpoint.

### LINKED ISSUE #1996: "Add to Cart" in product page demo failing
## Description of the issue found on ampproject.org

Please provide:

- Page link: https://amp.dev/documentation/examples/e-commerce/product_page/
- The "Add To Cart" functionality of the product page is failing ([the endpoint](https://amp.dev/samples_templates/product_page/add_to_cart?__amp_source_origin=https%3A%2F%2Famp.dev) being called is returning 404).

1. Go to [product demo](https://amp.dev/documentation/examples/e-commerce/product_page/) and scroll down to the "add to cart section".
2. Click on "add to cart".

The endpoint being called returns 404.

## Proposed fix

The demo calls an endpoint uploaded to [this Glitch repo](https://glitch.com/~abe-cart-service).
We need to either add this logic to amp.dev server, or make sure that the page calls the appropriate endpoint, and the service contains the proper amp.dev headers for CORS.

Additionally, we need to check that the related [Shopping Cart demo](https://amp.dev/documentation/examples/e-commerce/shopping_cart/?referrer=ampbyexample.com) is working properly.

keevkeev:
fixes #1996


sebastianbenz:
I'm afraid that this is not going to be that simple:

* I've just noticed, that the sample does not match the [latest version from ampbyexample](https://github.com/ampproject/amp-by-example/blob/master/src/60_Samples_%2526_Templates/Product_Page.html). We should migrate it. 
* We should migrate the add-to-cart endpoint to amp.dev (the glitch backend also needs a bit of refactoring/clean up).
* The documentation needs to be updated to reflect the new add-to-cart pattern. 

However, thinking more about this, it probably makes more sense to remove the ampbyexample product sample and link to the more polished and feature complete [ecommerce sample](https://camp.samples.amp.dev) instead. Instead we should have then single focused samples to explain specific patterns such as add-to-cart, etc. 

//cc @morsssss and @demianrenzulli for thoughts.

keevkeev:
@morsssss any update?

morsssss:
Hi! Just back from I/O and seeing this.

Of course I'd be honored if you felt that our new e-comm demo was sufficiently slicker than the existing samples, such that you wanted to use the demo instead. I do hope that our design and experience are nice!

I just think this decision is really up to you all. We do want to keep expanding our demo so that it has more features... like the ability to further mutate the cart, like search with autocompletion, etc. And we've written a couple of guides and plan to write more. So this could work well together.

(Minor note: reviewing the existing docs at `/documentation/examples/e-commerce/shopping_cart/` , I see a typo:  *credentails="include"*)

sebastianbenz:
Thanks Ben! I'll merge this for now and gather more opinions.

PR #2067, opened 2019-05-13
Handler to serve existing playground js/css file if hash is not correct

tharders:
https://github.com/ampproject/docs/issues/2029


playground/backend/index.js
```diff
@@ -27,8 +27,12 @@ playground.use(express.static(
       setHeaders: setCustomCacheControl,
     },
 ));
+
 playground.use('/api', require('./api.js'));
 
+playground.use( require('./resourceFallbackHandler.js'));
```
sebastianbenz: can you add a comment explaining that this needs to run last

playground/backend/resourceFallbackHandler.js
```diff
@@ -0,0 +1,51 @@
...
+// eslint-disable-next-line new-cap
+const handler = express.Router();
+
+
+/**
+ *  Handle requests to files with hash that where not found and serve the existing file.
+ *  This will prevent a 404 error while CDN or cluster nodes have different versions of our app.
+ *  See https://github.com/ampproject/docs/issues/2029
+ */
+handler.get(/\/(critical|main)\..+\.(js|css)/, (request, response, next) => {
```
sebastianbenz: Can we make this a generic fallback handler for all versioned URLs using this scheme PREFIX(.PREFIX)*.SHA.[js|css|png|jpg|mp4]?

playground/backend/resourceFallbackHandler.js
```diff
@@ -0,0 +1,51 @@
...
+  fs.readdir(path.join(__dirname, '../dist'), (err, files) => {
+    if (!err) {
```
sebastianbenz: Please do an early exit instead to avoid nesting.

playground/backend/resourceFallbackHandler.js
```diff
@@ -0,0 +1,51 @@
...
+      const fileNamePattern = new RegExp(request.params['0'] + '\\..+\\.' + request.params['1']);
+
+      const availableFile = files.find((fileName) => {
+        return fileNamePattern.test(fileName);
+      });
+
+      if (availableFile) {
+        setNoCache(response);
```
sebastianbenz: 👍

playground/backend/resourceFallbackHandler.js
```diff
@@ -0,0 +1,51 @@
...
+handler.get(/\/(critical|main)\..+\.(js|css)/, (request, response, next) => {
+  fs.readdir(path.join(__dirname, '../dist'), (err, files) => {
+    if (!err) {
+      const fileNamePattern = new RegExp(request.params['0'] + '\\..+\\.' + request.params['1']);
+
+      const availableFile = files.find((fileName) => {
+        return fileNamePattern.test(fileName);
+      });
+
+      if (availableFile) {
```
sebastianbenz: same as above, please do an early exit:

```
if (!availableFile) {
   return next();
}

File: playground/backend/resourceFallbackHandler.js [hunk 6]
```diff
@@ -0,0 +1,51 @@
...
+
+/**
+ *  Handle requests to files with hash that where not found and serve the existing file.
+ *  This will prevent a 404 error while CDN or cluster nodes have different versions of our app.
+ *  See https://github.com/ampproject/docs/issues/2029
+ */
+handler.get(/\/(critical|main)\..+\.(js|css)/, (request, response, next) => {
+  fs.readdir(path.join(__dirname, '../dist'), (err, files) => {
+    if (!err) {
+      const fileNamePattern = new RegExp(request.params['0'] + '\\..+\\.' + request.params['1']);
```
sebastianbenz: Can you make this more robust against path changes?

playground/backend/resourceFallbackHandler.js
tharders: The handler now matches any urls (including paths) that have a hash in the file name.
A hash is considered a block with lower case character and digits that is at least 15 characters long and separated by dots from the rest of the URL

playground/backend/resourceFallbackHandler.js
```diff
@@ -0,0 +1,51 @@
...
+const handler = express.Router();
+
+
+/**
+ *  Handle requests to files with hash that where not found and serve the existing file.
+ *  This will prevent a 404 error while CDN or cluster nodes have different versions of our app.
+ *  See https://github.com/ampproject/docs/issues/2029
+ */
+handler.get(/\/(critical|main)\..+\.(js|css)/, (request, response, next) => {
+  fs.readdir(path.join(__dirname, '../dist'), (err, files) => {
```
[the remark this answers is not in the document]
tharders: The handler is now moved into the lib folder and can be reused. The base path has to be provided.

playground/backend/index.js
```diff
@@ -27,8 +28,13 @@ playground.use(express.static(
       setHeaders: setCustomCacheControl,
     },
 ));
+
 playground.use('/api', require('./api.js'));
 
+// The fallback handler has to be last (kick in only if the request was not handled)
+playground.use( resourceFallbackHandler(path.join(__dirname, '../dist')));
```
sebastianbenz: please move `../dist` into a constant as it's used multiple times

platform/lib/routers/resourceFallbackHandler.js
```diff
@@ -0,0 +1,82 @@
...
+    const requestPath = request.params['0'];
+    const fileNamePrefix = request.params['1']; // including the dot at the end
+    const fileNamePostfix = request.params['2']; // starting with a dot
+
+    try {
+      const folder = path.join(rootDir, requestPath);
+      const availableFile = findFile(folder, fileNamePrefix, fileNamePostfix);
+
+      if (availableFile) {
+        setNoCache(response);
```
sebastianbenz: Thinking about this again, it's probably OK to use a less strict caching policy (e.g. e-tag based caching would be OK).

tharders: We do not send the correct content for the requested URL, but a newer or older version. So the response should not be cached, especially when we serve an older version.
What we do is not the ideal solution to the problem, but better than a 404 error.

sebastianbenz: If the response content has not changed, the browser does not need to re-download it. So setting `max-age: 0` and `ETag: 123455` might be a better approach. Afaik express sets the Etag by default. `setNoCache`currently force the browser to always re-download.

https://developers.google.com/web/fundamentals/performance/optimizing-content-efficiency/http-caching#validating_cached_responses_with_etags

tharders: When the content is not changed, webpack should not generate new hashes for the files.
So I would leave it at that, since it is an edge case anyway.

sebastianbenz: I'm talking about the Etags generated by express based on the response content. Use case would be reloads while the js/css assets are stale.

sebastianbenz: Anyway, no need to change anything. Browsers make use of Etags even if `private` and `no-cache` directives are given.

Issue #2051, opened 2019-05-10
Host runtime log messages


sebastianbenz:
AMP runtime messages of the form:

```
  message(`${foo} failed to ${bar}.`)
```

will be transpiled into:

```
  linkedMessage('0', foo, bar)
```

'0' is indirected to a table linked elsewhere. The most appropriate place to put this is probably amp.dev.

amp.dev needs infra to serve these error messages and interpolate the message:

```
  https://error.amp.dev/?e=0&s[]=amp-list&s[]=initialize&v=XXXX
```

Would then render HTML:

```
  amp-list failed to initialize.
```

`v` is an identifier for runtime version since the message sets divert. The table will be static and JSON-formatted. The keys are error ids as base36 strings and the values are the messages in printf interpolation format.

```
  {
    '0': '%s failed to %s.',
    // ...
  }
```

Things we need to do:

[] Download and cache versioned error message JSON from cdn.ampproject.org
[] Implement log.amp.dev endpoint serving HTML error messages
[] Design the error message page

//cc @alanorozco

sebastianbenz:
@alanorozco have you got a link to a sample messages.json file?

patrickkettner:
@sebastianbenz what happened with this

sebastianbenz:
Good question. @alanorozco might know more..

PR #2079, opened 2019-05-14
Setup runtime logging endpoint

sebastianbenz:
* adds a new endpoint for the log.amp.dev subdomain.
* moves server-side templates into a helper class

The actual log formatting is not yet implemented. See #2051.

### DIFF
--- a/platform/lib/routers/runtimeLog.js
+++ b/platform/lib/routers/runtimeLog.js
@@ -1,0 +1,47 @@
+/**
+ * Copyright 2019 The AMP HTML Authors. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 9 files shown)


platform/lib/routers/runtimeLog.js
```diff
@@ -0,0 +1,47 @@
...
+
+// eslint-disable-next-line new-cap
+const runtimeLog = express.Router();
+const logFormatter = new LogFormatter();
+
+runtimeLog.get('/', async (request, response) => {
+  const message = {
+    version: request.query.v,
+    id: request.query.id,
+    params: request.query.s,
```
matthiasrohmer: Does it make sense to document this somewhere besides in #2051?

sebastianbenz: Yes, I'll flesh this out more in subsequent PRs.

PR #2085, opened 2019-05-14
Updating roadshow cities

morsssss:
Adding new roadshow cities, moving past cities into the past. Because they are. In the past.

Also made a CSS change to ensure that event cards don't vary in size if they have text that pushes them beyond 350 px. Please feel to modify this as you wish!


morsssss:
Also, AFAIK the city dots are now part of the main SVG. I can no longer add them via CSS. Is that right?

If so, can someone at least add some dots for Africa? We went there in March and the continent has lost the dots it had on ampproject.org!

morsssss:
thank you!

PR #2090, opened 2019-05-15
🐛 added title to email samples

### LINKED ISSUE #2059: Parse title from URL for email samples in playground
...otherwise the playground won't show a title in the header

keevkeev:
Without a title in the samples files, the amp playground will render 'untitled' in the header which is undesired. closes  #2059


fstanis:
Left a comment in #2059 as well - I don't think we should use `<title>` in emails as it causes confusion (and I also hope we can just remove it from the spec all together).

I suggest we hold off until we figure out how to solve this.

matthiasrohmer:
Rejected in favor of letting the playground extract the title from the filenames.

Issue #2054, opened 2019-05-10
Migrate all ampbyexample backends to amp.dev


sebastianbenz:
...and fix/improve any broken samples along the way

//cc @matthiasrohmer @lluerich @tharders

CrystalOnScript:
The [amp-autocomplete](https://amp.dev/documentation/examples/components/amp-autocomplete/?format=websites#amp-autocomplete) are not working. @caroqliu pointed out the DOM is also not rending `<input type="text">` causing the examples to fail.

lluerich:
@sebastianbenz I finished looking through all the examples. Most of them are hosted on ampbyexample.com, but there are also some hosted on glitch and herokuapp. Since the title of the ticket concerns only ambyexample backends, I suggest, that I start migrating them first.

sebastianbenz:
SGTM. Can you start migrating the broken samples first?

>

PR #2093, opened 2019-05-15
♻️ Migrate amp_inputmask backend API

lluerich:
Migrated the backend/amp-inputmask.go from ampbyexample.com to amp.dev.
I fixed the examples and updated the second example to match its headline.

### DIFF
--- a/examples/api/amp-inputmask.js
+++ b/examples/api/amp-inputmask.js
@@ -22,9 +22,7 @@ const upload = multer();
 // eslint-disable-next-line new-cap
 const examples = express.Router();
 
-examples.post('/default', upload.none(), handleRequest);
-examples.post('/postal', upload.none(), handleRequest);
-examples.post('/phone', upload.none(), handleRequest);
+examples.post(['/default', '/postal', '/phone'], upload.none(), handleRequest);
 
 function handleRequest(request, response) {
   const code = request.body ? request.body['code'] : '';
--- a/examples/api/amp-inputmask.js
+++ b/examples/api/amp-inputmask.js
@@ -1,0 +1,64 @@
+/**
+ * Copyright 2019 The AMP HTML Authors. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS-IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+'use strict';
+
+const express = require('express');
+const multer = require('multer');
+const upload = multer();
+
+// eslint-disable-next-line new-cap
+const examples = express.Router();
+
+examples.all('/default', upload.none(), (request, response) => {
+  if (request.method !== 'POST') {
+    response.send('post only');
+    return;
+  }
+  const code = request.body ? request.body.code : '';
+  const codeUnmasked = request.body ? request.body.codeUnmasked : '';
+  response.json({
+    code,
+    'code-unmasked': codeUnmasked,
+  });
+});
+
+examples.all('/postal', upload.none(), (request, response) => {
+  if (request.method !== 'POST') {
+    response.send('post only');
+    return;
+  }
+  const code = request.body ? request.body['code'] : '';
+  const codeUnmasked = request.body ? request.body['code-unmasked'] : '';
+  response.json({
+    'code': code,
+    'code-unmasked': codeUnmasked,
+  });
+});
+
+examples.all('/phone', upload.none(), (request, response) => {
+  if (request.method !== 'POST') {
+    response.send('post only');
+    return;
+  }
+  const phone = request.body ? request.body['code'] : '';
+  const phoneUnmasked = request.body ? request.body['code-unmasked'] : '';
+  response.json({
+    'code': phone,
+    'code-unmasked': phoneUnmasked,
+  });
+});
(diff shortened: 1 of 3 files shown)


examples/api/amp-inputmask.js
```diff
@@ -0,0 +1,64 @@
...
+'use strict';
+
+const express = require('express');
+const multer = require('multer');
+const upload = multer();
+
+// eslint-disable-next-line new-cap
+const examples = express.Router();
+
+examples.all('/default', upload.none(), (request, response) => {
```
matthiasrohmer: Why not use `examples.post` in the first place? Also valid for the other routes.

examples/api/amp-inputmask.js
```diff
@@ -0,0 +1,64 @@
...
+  if (request.method !== 'POST') {
+    response.send('post only');
+    return;
+  }
+  const code = request.body ? request.body.code : '';
+  const codeUnmasked = request.body ? request.body.codeUnmasked : '';
```
matthiasrohmer: Is it intentional it's `request.body.codeUnmasked` instead of `request.body['code-unmasked']` here?

examples/source/1.components/amp-inputmask.html
```diff
@@ -44,7 +44,7 @@
 
      In the mask, `L` allows the user to enter an alphabetic character. `0` denotes a numeric character. `_` is a literal space character. See the full mask specification at [the `amp-inputmask` documentation](https://github.com/ampproject/amphtml/blob/master/extensions/amp-inputmask/README.md)
    -->
-  <form class="sample-form" method="post" action-xhr="<% hosts.backend %>/components/amp-inputmask/postal" target="_top">
+  <form class="sample-form" method="post" action-xhr="/documentation/examples/api/postal" target="_top">
```
matthiasrohmer: To be consistent I'd find it good practice to keep using absolute URLs here and prepend with `<% hosts.platform %>`.

examples/source/1.components/amp-inputmask.html
```diff
@@ -58,8 +58,8 @@
    <!--
     If the `mask-output` attribute is set to `"alphanumeric"`, `amp-inputmask` will add a hidden input with only the alphanumeric characters from the original input. The default value is `"raw"`
    -->
-  <form class="sample-form" method="post" action-xhr="<% hosts.backend %>/components/amp-inputmask/postal" target="_top">
-    <label>Postal code: <input name="code" mask="L0L_0L0" placeholder="A1A 1A1" mask-output="alphanumeric"></label>
+  <form class="sample-form" method="post" action-xhr="/documentation/examples/api/phone" target="_top">
```
matthiasrohmer: Also use absolute URL.

examples/api/amp-inputmask.js
lluerich: This makes total sense, but the original amp-inputmask.go has a specific response to all request that are not via POST. If that is not necessary, we could switch over to your suggestion.

sebastianbenz: Please do as Matthias suggested, the original implementation was due to Go's more rudimentary request handling APIs.

examples/api/amp-inputmask.js
```diff
@@ -0,0 +1,64 @@
...
+  const code = request.body ? request.body['code'] : '';
+  const codeUnmasked = request.body ? request.body['code-unmasked'] : '';
+  response.json({
+    'code': code,
+    'code-unmasked': codeUnmasked,
+  });
+});
+
+examples.all('/phone', upload.none(), (request, response) => {
+  if (request.method !== 'POST') {
```
sebastianbenz: The implementations are almost identical, please extract into a single parameterized method.

examples/source/1.components/amp-inputmask.html
```diff
@@ -44,7 +44,7 @@
 
      In the mask, `L` allows the user to enter an alphabetic character. `0` denotes a numeric character. `_` is a literal space character. See the full mask specification at [the `amp-inputmask` documentation](https://github.com/ampproject/amphtml/blob/master/extensions/amp-inputmask/README.md)
    -->
-  <form class="sample-form" method="post" action-xhr="<% hosts.backend %>/components/amp-inputmask/postal" target="_top">
+  <form class="sample-form" method="post" action-xhr="<% hosts.platform %>/documentation/examples/api/postal" target="_top">
```
sebastianbenz: Can we use relative URLs instead to reduce noise.

examples/source/1.components/amp-inputmask.html
```diff
@@ -58,8 +58,8 @@
    <!--
     If the `mask-output` attribute is set to `"alphanumeric"`, `amp-inputmask` will add a hidden input with only the alphanumeric characters from the original input. The default value is `"raw"`
    -->
-  <form class="sample-form" method="post" action-xhr="<% hosts.backend %>/components/amp-inputmask/postal" target="_top">
-    <label>Postal code: <input name="code" mask="L0L_0L0" placeholder="A1A 1A1" mask-output="alphanumeric"></label>
+  <form class="sample-form" method="post" action-xhr="<% hosts.platform %>/documentation/examples/api/phone" target="_top">
```
sebastianbenz: see above

examples/api/amp-inputmask.js
```diff
@@ -22,43 +22,17 @@ const upload = multer();
...
-  const code = request.body ? request.body.code : '';
-  const codeUnmasked = request.body ? request.body.codeUnmasked : '';
-  response.json({
-    code,
-    'code-unmasked': codeUnmasked,
-  });
-});
+examples.post('/default', upload.none(), handleRequest);
+examples.post('/postal', upload.none(), handleRequest);
+examples.post('/phone', upload.none(), handleRequest);
```
sebastianbenz: simpler: 

```
examples.post(['/default', '/postal', '/phone'] , upload.none(), handleRequest);
```

lluerich: You're right, I just changed it to your suggestion.

PR #2096, opened 2019-05-15
Allow samples to override title via metadata

fstanis:


sebastianbenz:
Which problem does this solve?

matthiasrohmer:
@sebastianbenz, would allow email samples having titles without an actual `<title>` element. Also see #2059.

sebastianbenz:
This doesn't solve the problem. See me comment in the issue.

fstanis:
Ugh, you're right, I just realized this won't affect the playground...

PR #2097, opened 2019-05-15
Allow samples to specify author like articles

fstanis:
I think the author / contributor mechanism for articles is a good way to know whom to contact to flag issues - this PR gives us the ability to the same with samples, using the metadata section.


fstanis:
> Now we only need someone to write a script that initializes the fields for the existing samples from Github.

It wasn't as easy as I thought when I started, but #2102 is an attempt at this.

PR #2104, opened 2019-05-16
Create email sample for "infinite cards"

fstanis:
This sample demonstrates an email that can display an infinite number of items by allowing the user to dismiss existing items, which results in loading new ones.


examples/source/interactivity-dynamic-content/Infinite_Cards_Email.html
```diff
@@ -0,0 +1,123 @@
...
+      {{/items}}
+    </div>
+  </template>
+
+  <!--
+    Use `amp-list` to render the initial items from the server using the template defined above referred by its `id`.
+
+    The height matches the combined height of our cards and their margins. The initial server response defines the number of cards to be displayed (in this case four).
+  -->
+  <amp-list items="." single-item template="list-template" src="https://amp.dev/documentation/examples/api/photo-stream?items=4&width=160&height=90" layout="fixed-height" height="280">
```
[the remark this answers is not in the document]
fstanis: That's not valid in emails, not sure what the best approach is. My aim is for users to be able to copy-paste Infinite_Cards_Email.html and send it as-is.

examples/source/interactivity-dynamic-content/Infinite_Cards_Email.html
```diff
@@ -0,0 +1,123 @@
...
+preview: amp4email
+skipValidation: 'true'
+author: fstanis
+
+--->
+
+<!--
+## Introduction
+
+This sample demonstrates how to display a feed of data, allowing the user to go through
```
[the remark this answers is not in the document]
fstanis: So, this is an interesting problem. Emails allow XHR forms with `method="get"`, but AMP for websites doesn't, which leads to unintended / undefined behavior, specifically, it falls victim to XHR caching.

In an email, you get a different image every time you press "Next". On the web, to achieve this, you'd have to add `<input type="hidden" name="r" value="RANDOM" data-amp-replace="RANDOM">` (which is invalid in emails).

I could also modify the photo-stream endpoint to make it support POST and then change the form method to that. This may be the easiest solution. Any thoughts on that?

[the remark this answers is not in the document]
sebastianbenz: I still don't understand the problem you're solving.

examples/source/interactivity-dynamic-content/Infinite_Cards_Email.html
```diff
@@ -0,0 +1,123 @@
+<!---
+
+preview: amp4email
+skipValidation: 'true'
```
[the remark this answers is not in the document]
fstanis: Ugh, I just copied it from another sample. I'll remove and also send another PR to update all other email samples where we put this.

examples/source/interactivity-dynamic-content/Infinite_Cards_Email.html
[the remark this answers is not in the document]
fstanis: The idea is for the user to be able to refresh parts of an amp-list without refreshing the whole list.

I updated it to use post so it no longer has the caching issue and should work correctly now. PTAL.

examples/source/interactivity-dynamic-content/Infinite_Cards_Email.html
```diff
@@ -0,0 +1,120 @@
...
+    ## Initial list of cards
+
+    Define a template for the initial cards and their layout and give it an `id`. This template is used by `amp-list` for initially fetching several items from the server.
+
+    It contains in itself an `amp-form` for each item which references the template defined above referred by its `id`. This allows us to use a template inside another template.
+
+    Note: This template contains the same markup (in this case, a single `amp-img`) as used in the template above to render the initial items. This is wrapped inside `<div class="initial-content">` which becomes hidden the first time the user submits the form.
+  -->
+  <template id="list-template" type="amp-mustache">
+    <div class="card-container">
```
sebastianbenz: nit: I prefer to keep html markup clean and only add divs if really necessary. You can replace `card-container` and `card` by targeting your amp-list directly (assuming it has an id): 

```
#photo-list {
}
#photo-list  [role="listitem"] {
}
```

examples/source/interactivity-dynamic-content/Infinite_Cards_Email.html
```diff
@@ -0,0 +1,120 @@
...
+      {{#items}}
+      <form class="card" method="post" action-xhr="https://amp.dev/documentation/examples/api/photo-stream?single&width=160&height=90">
```
sebastianbenz: replace `https://amp.dev` with `<% hosts.platform %> so that'll also work locally

examples/source/interactivity-dynamic-content/Infinite_Cards_Email.html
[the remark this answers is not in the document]
sebastianbenz: replace https://amp.dev with `<% hosts.platform %> so that'll also work locally

examples/source/interactivity-dynamic-content/Infinite_Cards_Email.html
```diff
@@ -0,0 +1,120 @@
...
+        <div class="initial-content">
+          <amp-img src="{{imageUrl}}" layout="fixed" width="160" height="90"></amp-img>
+        </div>
+
+        <div submit-success template="item-template"></div>
+
+        <div submitting>Loading next image...</div>
+
+        <div class="next-button">
```
sebastianbenz: same as above, is the wrapper div needed?

fstanis: I guess not, I made the button wider instead.

examples/source/interactivity-dynamic-content/Infinite_Cards_Email.html
fstanis: Done.

examples/source/interactivity-dynamic-content/Infinite_Cards_Email.html
fstanis: Done, sort of. Emails don't currently support attribute selectors, so I had to keep the classes.

fstanis:
> In general, I think this sample is too complex.

I simplified it a bit further. The main idea is to show how you can only update part of an amp-list by using forms and nested templates.

> The form based update is great, but why does it need to be embedded inside an amp-list? If there's a common use case for why it needs to be there it'd make sense, but in this sample it feels a bit random.

Sadly, the main reason is that the lack of SSR means that the go-to pattern is to wrap large parts of the email in an amp-list. No real way around that, I guess.

examples/source/interactivity-dynamic-content/Infinite_Cards_Email.html
```diff
@@ -0,0 +1,100 @@
...
+<body>
+  <!--
+    ## Single item template
+
+    Define a template for a single item inside a card and give it an `id`. This template is used by `amp-form` for displaying new items.
+
+    In this case, we use a single `amp-img`.
+  -->
+  <template id="item-template" type="amp-mustache">
+    <amp-img src="{{items.imageUrl}}" layout="fixed" width="160" height="90"></amp-img>
```
sebastianbenz: `items.imageUrl` sounds wrong. Shouldn't it be `item.imageUrl`?

examples/source/interactivity-dynamic-content/Infinite_Cards_Email.html
```diff
@@ -0,0 +1,100 @@
...
+      bottom: 0;
+      width: 100%;
+    }
+  </style>
+</head>
+<body>
+  <!--
+    ## Single item template
+
+    Define a template for a single item inside a card and give it an `id`. This template is used by `amp-form` for displaying new items.
```
sebastianbenz: You're changing the way you explain your example from the explanatory "We use ..." to the tutorial style "Define ..". I prefer the first style.

examples/source/interactivity-dynamic-content/Infinite_Cards_Email.html
```diff
@@ -0,0 +1,100 @@
...
+    In this case, we use a single `amp-img`.
+  -->
+  <template id="item-template" type="amp-mustache">
+    <amp-img src="{{items.imageUrl}}" layout="fixed" width="160" height="90"></amp-img>
+  </template>
+
+  <!--
+    ## Initial list of cards
+
+    Define a template for the initial cards and their layout and give it an `id`. This template is used by `amp-list` for initially fetching several items from the server.
```
sebastianbenz: Explain why you're using amp-list. E.g.: We use an amp-list to fetch up-to-date content from our server.

examples/source/interactivity-dynamic-content/Infinite_Cards_Email.html
```diff
@@ -0,0 +1,100 @@
...
+
+    It contains in itself an `amp-form` for each item which references the template defined above referred by its `id`. This allows us to use a template inside another template.
```
sebastianbenz: This sentence does not help here: `This allows us to use a template inside another template.` Explain why you need this.

examples/source/interactivity-dynamic-content/Infinite_Cards_Email.html
fstanis: I aligned it with what `<amp-list single-item>` does (it assumes there's `{items: {...}}` by default unless `items` attribute is set to something else).

fstanis:
Done. PTAL

sebastianbenz:
@fstanis can you please also address the other comments

fstanis:
I think all comments should be addressed now.

PR #2105, opened 2019-05-16
Create faux lightbox overlay email example

fstanis:
This example is an alternative to `amp-lightbox`, mainly aimed at users faced with [#22297](https://github.com/ampproject/amphtml/issues/22297).

### DIFF
--- a/examples/source/interactivity-dynamic-content/Faux_Lightbox_Overlay.html
+++ b/examples/source/interactivity-dynamic-content/Faux_Lightbox_Overlay.html
@@ -60,36 +60,24 @@ It provides an alternative to using the [`amp-lightbox`](/documentation/componen
     We use the `overlays` object to hold the state of all overlays on the page. Each overlay has a separate property inside this object it binds to, in this case `overlays.o1`.
 
     Pressing the "Show overlay" button changes the state, displaying both the overlay and the background on the whole page.
-  -->
-  <div class="overlay-container">
-    <p style="padding-top: 100px">Lorem ipsum dolor sit amet, consectetur adipiscing elit. Mauris non porta metus, eget lobortis metus. Duis vel arcu quis leo auctor vulputate commodo ac magna. Curabitur suscipit sit amet nisl quis pulvinar. Donec at purus non nisl molestie fringilla. Quisque consequat ligula ut ligula pellentesque mollis. Phasellus a rhoncus risus. Curabitur molestie lorem metus, placerat lobortis est molestie finibus. Pellentesque viverra nulla in turpis ultrices, ac placerat dolor semper. Praesent rutrum ullamcorper mi, non suscipit magna finibus ut. Interdum et malesuada fames ac ante ipsum primis in faucibus. Sed nulla lectus, lacinia vitae nisi vitae, laoreet cursus magna. Praesent lacus est, vestibulum sed urna non, tristique fermentum felis. Nunc posuere porta nunc non auctor. Praesent fringilla rutrum iaculis. Duis euismod facilisis ligula, vel auctor mi pulvinar vitae.</p>
-    <button on="tap:AMP.setState({ overlays: { o1: true } })">Show overlay</button>
 
-    <div class="overlay-content" hidden [hidden]="!overlays.o1">
-      This is an overlay.
+    Clicking on the background closes all visible overlays by setting the `overlays` state to null.
+  -->
+  <div>
+    <div class="overlay-container">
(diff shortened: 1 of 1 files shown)


examples/source/interactivity-dynamic-content/Faux_Lightbox_Overlay.html
```diff
@@ -0,0 +1,96 @@
...
+<!-- -->
+<body>
+  <!--
+    Wrap the initially visible content as well as overlay content into an element with the `lightbox-container` class.
+
+    We use the `overlays` object to hold the state of all overlays on the page. Each overlay has a separate property inside this object it binds to, in this case `overlays.o1`.
+
+    Pressing the "Show overlay" button changes the state, displaying both the overlay and the background on the whole page.
+  -->
+  <div class="lightbox-container">
```
sebastianbenz: Would it be possible to add the class to the body?

examples/source/interactivity-dynamic-content/Faux_Lightbox_Overlay.html
```diff
@@ -0,0 +1,96 @@
...
+      This is an overlay.
+    </div>
+  </div>
+
+  <!--
+    A single page can contain multiple overlays.
+
+    In this case, tapping on the text itself opens the second overlay. The state property used by this overlay is different than the one above: `overlays.o2`.
+  -->
+  <div class="lightbox-container">
```
sebastianbenz: I find it a bit weird to have multiple Lightbox containers.

examples/source/interactivity-dynamic-content/Faux_Lightbox_Overlay.html
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    .lightbox-container {
+      position: relative;
+    }
+
+    .lightbox-overlay {
+      position: absolute;
+      top: 0;
+      z-index: 1;
```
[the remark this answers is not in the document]
fstanis: Done. 100 is the maximum allowed in Gmail.

examples/source/interactivity-dynamic-content/Faux_Lightbox_Overlay.html
```diff
@@ -0,0 +1,96 @@
...
+
+    <div class="lightbox-overlay" hidden [hidden]="!overlays.o1">
+      This is an overlay.
+    </div>
+  </div>
+
+  <!--
+    A single page can contain multiple overlays.
+
+    In this case, tapping on the text itself opens the second overlay. The state property used by this overlay is different than the one above: `overlays.o2`.
```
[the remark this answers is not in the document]
fstanis: Done.

examples/source/interactivity-dynamic-content/Faux_Lightbox_Overlay.html
fstanis: No - the container is used as base for positioning, i.e. the location on which the overlay should be made on top of. Having it on the body would just result in the overlay being on top of the whole document (similar to lightbox).

examples/source/interactivity-dynamic-content/Faux_Lightbox_Overlay.html
fstanis: Addressed above, it's used to "anchor" the `position: absolute` for the content. I renamed it overlay-container, but not sure if that makes more sense.

examples/source/interactivity-dynamic-content/Faux_Lightbox_Overlay.html
sebastianbenz: Hm, this doesn't really sound like a solution (or I didn't understand it correctly). Wouldn't this vary depending on desktop vs mobile viewports?

fstanis: The idea is to display a dismissable overlay on top of another element (i.e. use the same position as that element) while at the same time displaying a lightbox-like effect, hiding the rest of the document.

sebastianbenz:
Circling back on this after our discussion via chat. I don't think that this sample demonstrates good UX and does not solve the problem of a lack of Lightbox support. My suggestion would be to re-focus the sample on a good UX pattern that we can recommend.

fstanis:
I think that's a fair point. I'll think about it a bit and reopen if I come up with a good sample.

modernwarfareuplink:
@fstanis  IMHO, 
Can we not tackle this [#22297](https://github.com/ampproject/amphtml/issues/22297) by the following
1. you can use the scrollToTop option for fixing this issue, I don't think this stands strong on deprecating or removing this feature.

fstanis:
I don't think so.

Email is displayed in an iframe which can't be scrolled (it always has sufficient height to display everything). The parent (mail client) can be scrolled, but the top of the parent is not guaranteed to be the top of the email (e.g. what if you're seeing two emails at the same time).

PR #2141, opened 2019-05-21
Enable valid optimized AMP

sebastianbenz:
* switch to latests optimizer alpha
* change packager to filter routes based on host
* migrate to new optimizer API


sebastianbenz:
@matthiasrohmer two questions:

1. do we serve non-amp pages on the platform host?
2. redirects: are there any other valid AMP path patterns?

matthiasrohmer:
1. No, none that I know of besides `pages/static/serviceworker.html`. You're asking to make sure we're not breaking non-AMP pages? I've added a quick check to the BMW.com optimizer that checks for either `amp` or `⚡️` on `<html>` to make sure it only transforms AMP pages. Not the most solid check but, ...

2. `amp.html` should match the current setup. We did use the optimizer's default `?amp=1` for a while though when we were not yet optimizing statically. Not sure if it's worth adding.

platform/lib/routers/packager.js
```diff
@@ -43,7 +43,12 @@ const packager = (request, response, next) => {
     return;
   }
   // Don't package non valid AMP pages
-  if (!request.path.endsWith('.amp.html')) {
+  let pagesHost = config.hosts.platform.host;
+  if (config.hosts.platform.port) {
+    pagesHost += `:${config.hosts.platform.port}`;
+  }
+  if (request.get('host') !== pagesHost) {
```
matthiasrohmer: Oh I guess you we're asking to make sure the packager doesn't break on non-valid pages... How robust are the transformers of the packager?

Are samples using an experimental component for example are considered valid? Or would that break?

sebastianbenz: Transformations are the same as AMP Optimizer (and AMP Cache). Sending non-valid pages to the packager means just more overhead.

sebastianbenz:
1. This is about packager. Optimizer checks for the `amp` attribute in the `html` tag as well. 
2. ok

PR #2143, opened 2019-05-21
👽 Adapt boilerplate generator build to new optimizer API

matthiasrohmer:
🙂

### DIFF
--- a/boilerplate/build.js
+++ b/boilerplate/build.js
@@ -17,7 +17,6 @@
 'use strict';
 
 const AmpOptimizer = require('amp-toolbox-optimizer');
-const runtimeVersion = require('amp-toolbox-runtime-version');
 const path = require('path');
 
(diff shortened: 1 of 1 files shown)


boilerplate/build.js
```diff
@@ -63,7 +63,8 @@ async function generateOptimizedAmpFiles(output) {
 
 async function optimizeAmp(html) {
   const ampRuntimeVersion = await runtimeVersion.currentVersion();
-  return await ampOptimizer.transformHtml(html, {
+  const optimizer = AmpOptimizer.create();
+  return await optimizer.transformHtml(html, {
     ampUrl: './' + AMP_PATH,
```
sebastianbenz: The two parameters are no longer needed

matthiasrohmer: Makes sense.

PR #2150, opened 2019-05-22
✨ Add new show more button example

lluerich:
Besides just fixing the broken show more button example, I added a new example that shows the built-in functionality to display a show more button via amp-list attribute.
Both examples now use the photo-stream API.


examples/source/interactivity-dynamic-content/Show_More_Button.html
```diff
@@ -34,20 +34,65 @@
...
+    </div>
+    <amp-list-load-more load-more-failed>
+      ERROR
+    </amp-list-load-more>
+    <amp-list-load-more load-more-end>
+      END
+    </amp-list-load-more>
+  </amp-list>
+
   <!-- You can implement a show more button by using `amp-list` and `amp-bind`, where
```
sebastianbenz: Can you add a new heading: `## Show more using amp-list and amp-bind`

sebastianbenz: Can you also add a intro sentence along the lines of: "You can also manually implement the show more functionality using amp-bind and amp-state."

lluerich: Would it suffice to edit the paragraph starting line 85 to something like so: "You can also manually implement a show more button by using `amp-list` and `amp-bind`, where `amp-list`[...]"

sebastianbenz: Yes, + the heading

lluerich: PTAL

examples/source/interactivity-dynamic-content/Show_More_Button.html
```diff
@@ -82,7 +82,8 @@
     </amp-list-load-more>
   </amp-list>
 
-  <!-- You can implement a show more button by using `amp-list` and `amp-bind`, where
+  <!-- ## Show more using amp-list and amp-bind -->
+  <!-- You can also implement a show more button manually by using `amp-list` and `amp-bind`, where
   `amp-list` `src` data are bind to the value of an `amp-state` element. Here we are using two `amp-state`: `productsState`
```
sebastianbenz: one nit (didn't see it earlier): `productsState` doesn't really make sense anymore. Use `photos` instead.

lluerich: PTAL

PR #2151, opened 2019-05-22
Add missing size for hello world ad

tharders:
Currently the ad is not shown, because the size is empty:
https://preview.amp.dev/documentation/examples/introduction/amphtml_ads_hello_world/embed

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,9 +29,10 @@ jobs:
   include:
     - stage: preparation
       name: 'Check code, import documents, build samples, playground & boilerplate'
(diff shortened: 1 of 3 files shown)


matthiasrohmer:
Sorry, @tharders for polluting your PR - it made a good test case 😉 I haven't been thinking carefully enough when I was trying to implement the early-exit for PRs.

What we really want is having all our test commands in the `before_script` phase as only those scripts make a build fail/error immediately.

@sebastianbenz, check status now reports correctly.

PR #2158, opened 2019-05-23
🐛 Fix broken references and reenable Grow Reference Checker.

matthiasrohmer:
With this in place we should be good to create the production branch.


sebastianbenz:
Is the build failing because of samples not being generated?

PR #2176, opened 2019-05-24
Add ~/bin to path for grow exec

tharders:
Problem was, that the grow installation without pip does not put grow in the path, but specifies an alias to ~/bin/grow. This is not available when calling "grow" via node.
* Extra lib for grow to add ~/bin to the path
* To suport blanks in parameters sh can now be called with an array


tharders:
Problem with multiple whitespace in the command line is fixed

PR #2179, opened 2019-05-24
Style and layout: Updated auto-layout table

machal:
This was misleading. Both `width` and `height` are needed for `layout="fixed"`.

Demo: https://jsbin.com/rujipoh/6/edit?html,console


machal:
@googlebot I signed it!

pbakaus:
Good catch, thanks!

PR #2184, opened 2019-05-24
Restore dynamic importing of news/YT videos to show on homepage

pbakaus:
This brings back the automatic news/YouTube importer that automatically refreshes the content on the homepage.

Note: Thumbnail support for the blog isn't in this release yet, as this is dependent on an outstanding change on the AMP blog to not display the featured image in blog posts. Coming with subsequent commit.

/cc @cramforce as it will make him happy :)


pbakaus:
Note: Curiously, I had to disable the `o-invalid-this` from ESLint, as I didn't see a way to not use it when dealing with the request lib (see line 55). If you're unhappy about this @matthiasrohmer and @sebastianbenz, let me know how to fix :)

Issue #2194, opened 2019-05-28
Sidebar has no close button


sebastianbenz:
Two issues:

1. There's no way to close the sidebar
2. The sidebar is not scrolled to the top when opened.

<img width="410" alt="Screen Shot 2019-05-28 at 09 49 09" src="https://user-images.githubusercontent.com/380472/58460409-dd6e1000-812d-11e9-8a1b-55615a0fb1d9.png">

matthiasrohmer:
>The sidebar is not scrolled to the top when opened.

That's actually the desired behaviour: the sidebar should scroll to to the current active element. `amp-accordion` in the screenshot.

sebastianbenz:
Ok. The current behavior is still confusing as I'm lost within the massive sidebar content. 

Suggestions:

* add a fixed header with format toggle and close button
* collapse non-active sections to make it easier to navigate

Sidenote: I'm not a fan of the `Toggle Sidebar` button label as it doesn't tell me what's inside the sidebar. In particular as there's also a table of contents and a hamburger menu...

//cc @CrystalOnScript

sebil:
I've created a PR that fixes the no-closing issue on mobile.
Concerning the 2. issue "sidebar is not scrolled to the top", I'm waiting for feedback.

sebil:
Next step is making the format toggle sticky/fixed.

sebil:
So format-toggle is sticky now.

I've decided not putting the close button into the header of the sidebar (next to the format-toggle), because of known UX issues that come along with that. Having a close button in the lower right corner which can be reached quick with the thumb on a mobile device is a better solution, I think. What do you think?

Next step is collapsing non-active sections in the components sidebar. The other sidebars already have that feature.

sebil:
I've closed the PR #2210 for a new solution.
Right now it is not possible to have the _sticky format-toggle_ and _scrolling to to the current active element_ at the same time.
I need to figure out a new way to keep both features alive.

sebil:
Opened new PR with better solution for closing issue: #2239

sebil:
I just added a new PR #2266 for more improvements.

PR #2202, opened 2019-05-28
:bug: Decrease width of mobile sidebar

sebil:
Fixes no closing issue #2194


sebastianbenz:
I don't think that this fixes #2194. We should have an explicit close button instead. I don't think it's common knowledge that clicking the background will close the sidebar -- or am I missing anything?

sebil:
You're right, that could be not obvious enough.
I've created another PR #2210 that adds a close button.

PR #2206, opened 2019-05-28
Use binding=refresh|no in amp-list wherever possible

choumx:
Partial for https://github.com/ampproject/amphtml/issues/21901. Adding the `binding` attribute avoids a performance pitfall in amp-list.

- Use `binding=no` when the template does not contain any bindings.
- Use `binding=refresh` when the template contains bindings.
- [Pitfall & default behavior] Use `binding=always` when the template contains bindings **and** the amp-list needs to block initial render on evaluation of those bindings. 

/cc @morsssss


boilerplate/templates/index.html
```diff
@@ -257,7 +257,8 @@ <h3>Boilerplate</h3>
                 [src]="config"
                 single-item
                 items="config"
-                hidden [hidden]="!config.changed">
+                hidden [hidden]="!config.changed"
+                binding="refresh">
```
sebastianbenz: Is this OK, because the initial render does not render any content at all?

examples/source/1.components/amp-list.html
```diff
@@ -122,7 +122,8 @@
   The list content is rendered via an [amp-mustache template]({{g.doc('/content/amp-dev/documentation/components/reference/amp-mustache.md', locale=doc.locale).url.path}}). The template can be specified either by id, or by using a nested element.
```
sebastianbenz: This is the first amp-list sample, can you mention the `binding=no` attribute here.

sebastianbenz:
I went ahead and made the changes. Thanks @choumx .

choumx:
Sorry was out for a few days. Thanks @sebastianbenz!

PR #2210, opened 2019-05-29
Fix/sidebar

sebil:
This PR adds a 'close-sidebar' button to the mobile sidebar.
Due to the limited width of the sidebar, which would cause overflow issues, the close button will be placed next to and not within the sidebar.
Another improvement could be to move the close button to a separate file and include it in all sidebars to keep the code dry.

1. Update:
The format-toggle get sticky inside the mobile sidebar.


sebil:
Wait, before you merge this. I found a bug. I'm on it.

sebil:
Closing PR for a new solution.

PR #2212, opened 2019-05-29
Implement grow extension to convert example code

tharders:
* inline Preview for example source
* unit test content for example
* special build target to only build the unit tests contents
* first unit test

Unit test contents are also build with the regular build

Currently the preview feature is not added to any existing documentation file, but only to the unit-test contents.
Example:
http://localhost:8080/tests/examples/template
There are some examples where currently the 'inline' preview is selected and shown although they should not because they use <head> elements.


pages/content/unit-tests/examples/import.md
```diff
@@ -0,0 +1,25 @@
+---
+$title: Custom element dependency
+$order: 5
+---
+
+Start
+
+[example  preview="inline" playground="true" import="amp-accordion"]
```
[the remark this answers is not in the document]
sebastianbenz: rename to `imports` to make it clear that there can be multiple.

pages/content/unit-tests/examples/full_html.md
```diff
@@ -0,0 +1,38 @@
+---
+$title: Full document example
+$order: 8
+---
+
+Start
+
+[example preview="inline" playground="true"]
```
sebastianbenz: We should make `preview="inline" playground="true"` the default and test this case as well.

pages/extensions/amp_example_preview/util/example_extractor.py
```diff
@@ -0,0 +1,82 @@
+import re
+
+EXAMPLE_PATTERN = re.compile(r'\[\s*example(\s[^\]]*)?\](.*?\n```html *\n(.*?)\n```.*?)\[\s*/\s*example\s*\]',
+                             re.IGNORECASE | re.MULTILINE | re.DOTALL)
+
+PREVIEW_ATTRIBUTE_PATTERN = re.compile(r'\spreview\s*=\s*"([^"]+)"')
+PLAYGROUND_ATTRIBUTE_PATTERN = re.compile(r'\splayground\s*=\s*"([^"]+)"')
+IMPORT_ATTRIBUTE_PATTERN = re.compile(r'\simport\s*=\s*"([^"]+)"')
+TEMPLATE_ATTRIBUTE_PATTERN = re.compile(r'\stemplate\s*=\s*"([^"]+)"')
```
sebastianbenz: Parsing these via regex is overkill (and harder to maintain). Instead parse these into a settings map using string split.

pages/extensions/amp_example_preview/util/markdown_processor.py
```diff
@@ -0,0 +1,79 @@
...
+        inline_example = match.inlineExample
+
+        count = count + 1
+
+        doc.pod.logger.debug('Found example {} in {}'.format(inline_example.index, doc.pod_path))
+
+        result = result + content[pos:match.startTagStart]
+
+        result = result + '' \
+                 + '<!-- preview id={} mode={} playground={}\n'.format(
```
sebastianbenz: I'd prefer to encode the content into something that's easy to parse (e.g. JSON).

sebastianbenz: Would it be possible to write these into the frontmatter and reuse grow's frontmatter parser?

pages/content/unit-tests/examples/import.md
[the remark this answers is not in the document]
tharders: Good idea!

pages/content/unit-tests/examples/multi_examples_preview.md
```diff
@@ -0,0 +1,101 @@
+---
+$title: Multiple examples in one file
+$order: 12
+---
+
+First
+
+<!-- xexample
```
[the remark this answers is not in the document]
tharders: Oops, accidentally checked in this file

pages/extensions/amp_example_preview/util/preview_generator.py
```diff
@@ -0,0 +1,92 @@
...
+                amp_templates.discard(existing_import.group(2))
+
+            existing_import = IMPORT_PATTERN.search(content, existing_import.end(0))
+
+    # TODO: support different custom element and template versions
+
+    if amp_imports is not None and len(amp_imports) > 0:
+        for custom_element in amp_imports:
+            output = output + AMP_CUSTOM_DEPENDENCY_TEMPLATE.format(
+                    type='element', dependency=custom_element, version='0.1')
```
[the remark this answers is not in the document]
tharders: Nice! Did not know this exists.
Will also make the amp_dependency_injector.py use this.

pages/content/unit-tests/examples/full_html.md
tharders: @sebastianbenz  since the inline preview does not work for every example I rather would not enable the preview by default. There is the default_settings.md to test the default, which currently is preview=none and playground=true

pages/content/unit-tests/examples/template.md
```diff
@@ -0,0 +1,31 @@
...
+
+Start
+
+[example 
+     preview="inline" playground="true"
+     import="amp-date-display"
+     template="amp-mustache"]
+
+```html
+{% raw %}<amp-date-display timestamp-seconds="2147483648" layout="fixed-height" height="20">
```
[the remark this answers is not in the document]
tharders: @sebastianbenz This is to stop jinja2 evaluating the curly brackets inside the example

pages/content/unit-tests/examples/full_html.md
sebastianbenz: sgtm for now. We might be able to automatically determine the preview type in future.

pages/extensions/amp_example_preview/util/markdown_processor.py
tharders: Content is json encoded now.  Because some examples contain j2 expressions to generate urls, etc we have to evaluate the source in the template. I don't think this would all work in the frontmatter, even if we could use it otherwise.

pages/extensions/amp_example_preview/util/example_extractor.py
tharders: I create an attribute map now, but still use regex to find the attributes.

pages/extensions/amp_example_preview/util/preview_generator.py
```diff
@@ -0,0 +1,92 @@
+import re
+
+PREVIEW_TRIGGER = '<!-- preview'
```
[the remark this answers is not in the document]
tharders: I introduced an extra class for generating and parsing the preview and also have a constants.py where other shared constants are held

pages/extensions/amp_example_preview/util/preview_generator.py
```diff
@@ -0,0 +1,92 @@
...
+    match = PREVIEW_PATTERN.search(content)
+    while match:
+        settings = match.group(1)
+        doc.pod.logger.debug('generate preview' + settings)
+
+        output = output + content[pos:match.start(0)]
+
+        inline_preview = 'mode=inline' in settings
+        if inline_preview:
+            output = output + '<div class="ap-o-code-preview">\n' \
```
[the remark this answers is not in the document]
tharders: We use jinja2 templates now.
I was not 100% sure where to put the template.
It ended up here:
frontend/templates/views/partials/code-preview/code-preview.j2
If you have a better place, just let me know

pages/extensions/amp_example_preview/util/preview_generator.py
```diff
@@ -0,0 +1,92 @@
...
+
+    if amp_imports is not None and len(amp_imports) > 0:
+        for custom_element in amp_imports:
+            output = output + AMP_CUSTOM_DEPENDENCY_TEMPLATE.format(
+                    type='element', dependency=custom_element, version='0.1')
+
+    if amp_templates is not None and len(amp_templates) > 0:
+        for custom_template in amp_templates:
+            output = output + AMP_CUSTOM_DEPENDENCY_TEMPLATE.format(
+                    type='template', dependency=custom_template, version='0.2')
```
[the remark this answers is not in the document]
tharders: Done

pages/content/unit-tests/examples/head_style.md
```diff
@@ -0,0 +1,23 @@
+---
+$title: Example with style
+$order: 7
+---
+
+Start
+
+[example  preview="inline" playground="true"]
```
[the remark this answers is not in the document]
tharders: I changed the preview to "none" for the examples with more than a body for now.
We can change that once we have the iframe preview.

pages/content/unit-tests/_blueprint.yaml
```diff
@@ -0,0 +1,6 @@
+$view: /views/default.j2
```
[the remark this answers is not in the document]
tharders: done

package.json
```diff
@@ -18,6 +18,7 @@
     "build:local": "gulp build --env local",
     "build:staging": "gulp build --env staging",
     "build:prod": "gulp build --env production",
+		"build:unittests" : "gulp buildForUnitTests --env test --locales en --include_paths unit-tests",
```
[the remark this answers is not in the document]
tharders: done

pages/extensions/amp_example_preview/util/preview.py
```diff
@@ -0,0 +1,98 @@
...
+    def from_json(jsonData):
+        """
+        :type jsonData: str
+        """
+        result = ExamplePreview()
+        result.__dict__ = json.loads(unescape_html(jsonData))
+        return result
+
+    @staticmethod
+    def for_attributes(index, mode, playground, source):
```
sebastianbenz: Encapsulate all options in an object to simplify your code. With the current approach, adding a new flag requires changing code in multiple locations.

pages/extensions/amp_example_preview/util/preview.py
```diff
@@ -0,0 +1,98 @@
...
+        :type source: str
+        """
+        result = ExamplePreview()
+        result.index = index
+        result.mode = mode
+        result.playground = playground
+        result.source = source
+        return result
+
+def escape_html(content):
```
sebastianbenz: Can't we use an existing method for this?

frontend/templates/views/partials/code-preview/code-preview.j2
```diff
@@ -0,0 +1,12 @@
+{% if preview.mode == 'inline' %}
```
sebastianbenz: Does it make sense to use this in the examples template as well? //cc @matthiasrohmer

pages/extensions/amp_example_preview/util/preview.py
tharders: html module comes with Python 3.2 which is not available with grow.
Found that xml.sax.saxutils escape/unescape will also do what we need.

pages/extensions/amp_example_preview/util/preview.py
tharders: Done

frontend/templates/views/partials/code-preview/code-preview.j2
tharders: I think we should look at that when we have implemented the playground link and iframe preview, but my gut feeling is that there will be enough differences to keep it separated.

pages/extensions/amp_example_preview/util/markdown_processor.py
```diff
@@ -0,0 +1,73 @@
...
+        count = count + 1
+
+        doc.pod.logger.debug('Found example {} in {}'.format(inline_example.index, doc.pod_path))
+
+        result = result + content[pos:match.startTagStart]
+
+        preview = ExamplePreview(index=inline_example.index,
+                                 mode=inline_example.preview,
+                                 playground=inline_example.playground,
+                                 source=inline_example.source)
```
sebastianbenz: Sorry for not being clear enough. What I meant is to encapsulate options into an object:

```
opts = {
 
}
```

which is then passed around `ExamplePreview(opts)`. This avoids changing the ExamplePreview's constructor whenever adding a new option.

tharders: The current solution provides those benefits. What would be the benefit of using a dictionary instead of the named varargs?

sebastianbenz: all good - missed the python magic

PR #2213, opened 2019-05-29
♻️ Migrate paged_list API to amp.dev

lluerich:
Migrated the backend/paged-list.go from ampbyexample.com to amp.dev.
This updates and fixes the example on https://amp.dev/documentation/examples/interactivity-dynamic-content/paged_list/.


examples/source/interactivity-dynamic-content/Paged_List/api.js
```diff
@@ -0,0 +1,85 @@
...
+
+class AmpListResponse {
+  constructor(items) {
+    this.items = items;
+  }
+}
+
+examples.get('/search', (request, response) => {
+  const {query} = request;
+  const page = Number(query.page) > 0 ? Number(query.page) : 1;
```
sebastianbenz: You can simplify this to: 

```
let page = Number(request.query.page);
if (page < 0) {
  page = 1
}
```

examples/source/interactivity-dynamic-content/Paged_List/api.js
```diff
@@ -0,0 +1,85 @@
...
+'use strict';
+
+const express = require('express');
+
+// eslint-disable-next-line new-cap
+const examples = express.Router();
+const MAX_PAGE_COUNT = 5;
+const ITEMS_PER_PAGE = 4;
+
+class ProductListing {
```
sebastianbenz: Defining classes for simple data objects is overkill in javascript. 

```
const productListing = {
  title: ...,
  ...
}
```

is easier to understand in this case.

examples/source/interactivity-dynamic-content/Paged_List/api.js
lluerich: Alrigiht, but I also added a check if page is NaN, because by default it should return page 1, shouldn't it?

sebastianbenz: 👍

lluerich: PTAL

examples/source/interactivity-dynamic-content/Paged_List/api.js
lluerich: PTAL

PR #2214, opened 2019-05-29
♻️ Refactor travis conf to fix unintended builds

matthiasrohmer:
- Moves authentication from `before_install` phase to `script` phase as its only needed there. Otherwise all PRs coming from forks would always fail as they don't have access to the encrypted environment variables
- Adds job type check to stage conditions: `type != pull_request`
- Early exits the preparation stage if the build is a PR or isn't coming from `future`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,20 +10,20 @@ addons:
 cache:
 - pip
 - npm
-before_install:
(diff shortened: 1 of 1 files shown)


matthiasrohmer:
@sebastianbenz, PTAL. Early-exit doesn't seem to be a thing with Travis...

PR #2219, opened 2019-05-31
 🐛 Copy sitemap when building pages.

### LINKED ISSUE #2218: Block shared html files from search engines
For example:

```
https://amp.dev/shared/header
https://amp.dev/ru/shared/header
```

This started appearing in search console...

matthiasrohmer:
Fixes #2218, the sitemap for Grow is currently missing. Not quite sure how crawlers picked up shared pages in the first place as both header and footer have

```yaml
$sitemap:
  enabled: false
```

in their frontmatter. Therefore added `Disallow: /shared` to the production robots.txt

### DIFF
--- a/pages/static/robots/prod.txt
+++ b/pages/static/robots/prod.txt
@@ -1,4 +1,5 @@
 User-agent: *
 Allow: /
+Disallow: /shared
 
 Sitemap: https://amp.dev/sitemap.xml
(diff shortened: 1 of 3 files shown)


pages/static/robots/prod.txt
```diff
@@ -1,4 +1,5 @@
 User-agent: *
 Allow: /
+Disallow: /shared
```
sebastianbenz: We also need to disallow the language specific shared files (see my sample in the issue).

matthiasrohmer: They now don't make it to the build anymore with changing the sources for the page transformer: `!${project.paths.GROW_BUILD_DEST}/{*/shared,shared}/*.html`

Sorry, could have mentioned that.

sebastianbenz: clever

PR #2237, opened 2019-06-03
✨ List experiments on experiments page

### LINKED ISSUE #2175: Dynamically list origin trials on experiments page
As of #2170. To have a common place to refer to for ongoing origin trials it should be possible to add a `origin_trial: true` flag to the component docs frontmatter. A component with this flag enabled should then be automatically added to the experiments page.

matthiasrohmer:
Fixes #2175. Allows component documentation pages from ampproject/amphtml to have either `experimental` and/or `origin_trial` in their frontmatter to get then listed on the [experiments page](https://amp.dev/documentation/guides-and-tutorials/learn/experimental).

Open for suggestions to rename the frontmatter keys but I thought we might need differentiation as not every experimental component is available as an original trial.

Additionally I moved "There is a risk that some experiments will never ship as features on the AMP Project." to the top of the page because I found that bit quite important 😉 

@CrystalOnScript, please feel free to give copy feedback!


sebastianbenz:
@CrystalOnScript can you please take a look

PR #2249, opened 2019-06-04
♻️ Migrate amp-form API to amp.dev

lluerich:
Migrated the backend/amp-form.go from ampbyexample.com to amp.dev.
This updates and/or fixes the examples [here](https://amp.dev/documentation/examples/components/amp-analytics/), [here](https://amp.dev/documentation/examples/components/amp-autocomplete/), [here](https://amp.dev/documentation/examples/components/amp-date-picker/), [here](https://amp.dev/documentation/examples/components/amp-form/), [here](https://amp.dev/documentation/examples/interactivity-dynamic-content/conference_survey_email/?format=email) and [here](https://amp.dev/documentation/examples/interactivity-dynamic-content/multi_page_flow/).

### DIFF
--- a/examples/api/amp-form.js
+++ b/examples/api/amp-form.js
@@ -1,0 +1,76 @@
+/**
+ * Copyright 2019 The AMP HTML Authors. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS-IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+'use strict';
+
+const express = require('express');
+const multer = require('multer');
+const upload = multer();
+
+// eslint-disable-next-line new-cap
+const examples = express.Router();
+
+const ERROR_CASE_AMP_FORM = 'error';
+
+examples.all('/submit-form', submitForm);
+examples.post('/submit-form-xhr', upload.none(), submitFormXHR);
+examples.post('/submit-form-input-text-xhr', upload.none(), submitFormXHRInputText);
+examples.post('/verify-form-input-text-xhr', upload.none(), verifyFormXHRInputText);
+
+function submitForm(request, response) {
+  response.redirect(303, '/static/samples/files/amp-form-success.html');
+}
+
+function submitFormXHR(request, response) {
+  response.json({
+    result: 'ok',
+  });
+}
+
+function submitFormXHRInputText(request, response) {
+  const email = request.body ? request.body.email : '';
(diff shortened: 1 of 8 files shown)


examples/api/amp-form.js
```diff
@@ -0,0 +1,76 @@
...
+const express = require('express');
+const multer = require('multer');
+const upload = multer();
+
+// eslint-disable-next-line new-cap
+const examples = express.Router();
+
+const ERROR_CASE_AMP_FORM = 'error';
+
+examples.all('/submit-form', submitForm);
```
sebastianbenz: I think this should only use `get`

examples/static/samples/files/amp-form-success.html
```diff
@@ -0,0 +1,28 @@
...
+  <style amp-boilerplate>body{-webkit-animation:-amp-start 8s steps(1,end) 0s 1 normal both;-moz-animation:-amp-start 8s steps(1,end) 0s 1 normal both;-ms-animation:-amp-start 8s steps(1,end) 0s 1 normal both;animation:-amp-start 8s steps(1,end) 0s 1 normal both}@-webkit-keyframes -amp-start{from{visibility:hidden}to{visibility:visible}}@-moz-keyframes -amp-start{from{visibility:hidden}to{visibility:visible}}@-ms-keyframes -amp-start{from{visibility:hidden}to{visibility:visible}}@-o-keyframes -amp-start{from{visibility:hidden}to{visibility:visible}}@keyframes -amp-start{from{visibility:hidden}to{visibility:visible}}</style><noscript><style amp-boilerplate>body{-webkit-animation:none;-moz-animation:none;-ms-animation:none;animation:none}</style></noscript>
+</head>
+<body> 
+  <!--
+  ## Success page for amp-form
+
+  Success! Thanks for trying the amp-form demo.
+  -->
+  <!--
+  Back to [amp-form](/components/amp-form) sample.
```
sebastianbenz: Does this link work?

lluerich: No, nothing is rendered on the page. Since the example is only to show, that a page reload can be triggered via the `action` attribute, we could change the redirect in backend to just `response.redirect('back')`. This will just reload the page in the same location.

examples/api/amp-form.js
lluerich: PTAL

examples/static/samples/files/amp-form-success.html
sebastianbenz: In that case we should fix this and find a better way. I'm not really happy about simply reloading the page as users might miss it (we should change the sample phrasing as well, it's not about reloading). The simplest fix is probably to change the comment into HTML with a link back to the sample.

In the future, feel free to directly fix things like this that are obviously broken.

lluerich: Alright. For now I implemented your suggestion with a link back to the example.

examples/static/samples/files/amp-form-success.html
```diff
@@ -0,0 +1,25 @@
+<!---
```
sebastianbenz: nit: you can remove all the comments as they're no longer needed

lluerich: PTAL

sebastianbenz:
Thanks!

PR #2250, opened 2019-06-04
Disallow /tests/ for robots

tharders:
The tests folder that contain files to test the grow page generation should not end up in a search index


sebastianbenz:
Are they currently being deployed? If yes, we should fix that.

tharders:
Ok, we will not deploy on prod then.
Closing this PR then.

PR #2252, opened 2019-06-04
✨ Add credentials util.

matthiasrohmer:
This adds a credential helper to get API credentials from either an environment variable or a Google Cloud Datastore. Usage as follows:

```js
(async () => {
  console.log('Trying to get environment credential ...');
  await get('GOOGLE_MAPS')
      .then((credential) => { console.log(`GOOGLE_MAPS: ${credential}`); })
      .catch(() => { console.log('No key named GOOGLE_MAPS.'); });
})();
```

When exporting credentials as environment variables they need to be prefixed with `AMP_DEV_CREDENTIAL_` - in the above example the credential would have been exported by

```bash
$ export AMP_DEV_CREDENTIAL_GOOGLE_MAPS="secure-api-key"
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,7 +18,7 @@
     "build:local": "gulp build --env local",
     "build:staging": "gulp build --env staging",
     "build:prod": "gulp build --env production",
-		"build:grow-tests" : "gulp buildForGrowTests --env test --locales en --include_paths tests",
+    "build:grow-tests": "gulp buildForGrowTests --env test --locales en --include_paths tests",
     "log:staging": "gcloud app logs tail -s default --project amp-dev-staging",
     "posttest": "npm run lint",
     "start:local": "cd platform && NODE_ENV=local node serve.js",
@@ -43,12 +43,14 @@
     "@lib": "platform/lib"
   },
   "dependencies": {
+    "@google-cloud/datastore": "^4.0.0",
     "amp-toolbox-cors": "0.2.5",
     "casual": "1.6.2",
     "cheerio": "1.0.0-rc.3",
     "cookie-parser": "1.4.4",
     "cors": "2.8.5",
     "express": "4.17.1",
+    "feedparser": "^2.2.9",
     "http-proxy": "1.17.0",
     "iltorb": "2.4.3",
     "js-yaml": "3.13.1",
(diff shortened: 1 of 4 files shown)


platform/lib/utils/credentials.js
```diff
@@ -0,0 +1,65 @@
...
+        }
+      });
+    }
+
+    if (credential) {
+      resolve(credential);
+      return;
+    }
+
+    reject(Error(`No credential with key ${key}`));
```
sebastianbenz: Maybe mention that this can be provided via an environment variable.

platform/lib/utils/credentials.js
```diff
@@ -0,0 +1,65 @@
...
+
+/**
+ * Tries to get a credential by name either from an environment variable
+ * or if on app engine from the GCloud Datastore
+ * @param  {String} key A named Credential entity entry in the datastore
+ * @return {Promise}
+ */
+function get(key) {
+  return new Promise((resolve, reject) => {
+    let credential = process.env[`${ENV_PREFIX}${key.toUpperCase()}`];
```
sebastianbenz: suggestion: simplify and improve error handling

```
function get(key) {
  let credential = process.env[`${ENV_PREFIX}${key.toUpperCase()}`];
  if (credential) {
    return Promise.resolve(credential);
  }
  if (!config.isProdMode() && !config.isStageMode()) {
    return Promise.error('ENV variable missing ...');
  }
  const datastoreKey = datastore.key([CREDENTIAL_ENTITY, key]);
  ....
  // return a different error if datastore does not contain key
```

platform/lib/utils/credentials.js
```diff
@@ -0,0 +1,65 @@
...
+    if (!credential && (config.isProdMode() || config.isStageMode())) {
+      const datastoreKey = datastore.key([CREDENTIAL_ENTITY, key]);
+      datastore.get(datastoreKey, (e, entity) => {
+        if (!e) {
+          credential = entity.credential;
+        }
+      });
+    }
+
+    if (credential) {
```
sebastianbenz: Does this work as datastore.get is an async call?

package.json
```diff
@@ -43,12 +43,14 @@
     "@lib": "platform/lib"
   },
   "dependencies": {
+    "@google-cloud/datastore": "^4.0.0",
```
sebastianbenz: pin the version

platform/lib/utils/credentials.js
matthiasrohmer: Of course not, thanks.

matthiasrohmer:
Thanks for your good input!

Issue #2195, opened 2019-05-28
Fix sidebar scrolling behaviour


matthiasrohmer:
The sidebar scrolling/nesting behaviour is really off currently and needs an update. See #1652 for history.

matthiasrohmer:
@sebil, just noticed the TOC sidebar has a similar problem as the one on the left. What do you think? Is this also fixable?

sebil:
This is already fixed with the same PR for the sidebar.

sebil:
I just added a new PR #2266 for more improvements.

PR #2253, opened 2019-06-04
Fix/sidebar

sebil:
With this PR the sidebar becomes much more usable again.
Both, the format-toggle and the sidebar-toggle stay sticky at the same height.
If you scroll inside the sidebar the format-toggle stays at its position.
Autoscrolling to an active nav-item is still working with this modification.
Also there are some bug-fixes and code and styling improvements.

Relies to #2194 and #2195

### DIFF
--- a/frontend/scss/components/atoms/sidebar-toggle.scss
+++ b/frontend/scss/components/atoms/sidebar-toggle.scss
@@ -131,9 +131,12 @@
     display: inline-block;
 
     &-label {
+      position: sticky;
+      top: 86px;
       display: flex;
       margin-top: -52px;
       margin-left: -40px;
+      padding-bottom: 26px;
     }
 
     &-input {
@@ -141,9 +144,7 @@
         &~ .ap--sidebar {
           grid-column: 1 / 2;
           nav {
-            ul {
-              display: none;
-            }
+            display: none;
           }
           svg {
             transform: rotate(180deg);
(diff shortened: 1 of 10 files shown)


matthiasrohmer:
@sebastianbenz you want to have look before merge?

sebil:
Thanks for the feedback.
I'm going to tweak this a little now.

In the meantime, can you tell me if you still want the component-sidebar letters to be collapsable when inactive?
I thought about this, and I'm not sure if it really improves the usability of the sidebar.

Here is a test screenshot how it would look like (arrows missing).
component-sidebar-test

sebastianbenz:
You're right. Doesn't work too well. Thanks for flagging this. How about
showing letters as shortcuts on the right-hand side? Similar to the iOS
contacts app: find-fix-missing-contacts-ios-7-1-2.w1456.jpg
<https://img.gadgethacks.com/img/35/14/63541628772516/0/find-fix-missing-contacts-ios-7-1-2.w1456.jpg>
.

>

matthiasrohmer:
@sebastianbenz, would you be fine approving this and let @sebil deal with improving the animation tomorrow? I know the sidebar has been ugly for a while but that's our chance to have it broken one day less 😉

sebil:
> You're right. Doesn't work too well. Thanks for flagging this. How about showing letters as shortcuts on the right-hand side? Similar to the iOS contacts app: find-fix-missing-contacts-ios-7-1-2.w1456.jpg <https://img.gadgethacks.com/img/35/14/63541628772516/0/find-fix-missing-contacts-ios-7-1-2.w1456.jpg> .
> […](#)

I opened a new issue to keep this alive #2269

PR #2258, opened 2019-06-05
Specific robots.txt for subdomains and redirect on 404 for go.amp.dev

tharders:
Fix https://github.com/ampproject/docs/issues/2251
While looking at the robots.txt problem I found that go.amp.dev used the 404 page from amp.dev which has relative links in the menu that will not work on go.amp.dev
https://go.amp.dev/robots.txt
To fix this we redirect to amp.dev when the path is not found (like all the other subdomains do)


platform/lib/routers/robots.js
```diff
@@ -0,0 +1,36 @@
...
+const express = require('express');
+const {setMaxAge} = require('@lib/utils/cacheHelpers');
+const project = require('@lib/utils/project');
+const config = require('@lib/config');
+
+function createRobotsHandler(robotsFile) {
+  // eslint-disable-next-line new-cap
+  const router = express.Router();
+  const robotsFilePath = config.isProdMode() ? 'robots/' + robotsFile : 'robots/disallow_all.txt';
+  router.get('/robots.txt', (request, response) => {
```
sebastianbenz: nit: add a line break before

platform/lib/routers/robots.js
```diff
@@ -0,0 +1,36 @@
...
+
+const express = require('express');
+const {setMaxAge} = require('@lib/utils/cacheHelpers');
+const project = require('@lib/utils/project');
+const config = require('@lib/config');
+
+function createRobotsHandler(robotsFile) {
+  // eslint-disable-next-line new-cap
+  const router = express.Router();
+  const robotsFilePath = config.isProdMode() ? 'robots/' + robotsFile : 'robots/disallow_all.txt';
```
sebastianbenz: nit: using `path.join` makes this a bit cleaner

```
const robotsFilePath = path.join('robots', config.isProdMode() ? robotsFile : 'disallow_all.txt');
```

PR #2260, opened 2019-06-05
Let topic link 'Track engagement with analytics' open index page

tharders:
Currently an index page will only get a link in the navigation topic if the filename is index.md.
(It is not enough that the path is configured to be index.html)
So in order to have a link in the navigation for the following page I renamed it to index.md
https://amp.dev/documentation/guides-and-tutorials/optimize-measure/configure-analytics/?format=websites


sebastianbenz:
Is this so that when I click on "Track engagement with analytics" that the first guide is displayed automatically?

tharders:
@sebastianbenz Yes. Since the navigation template can only check for the "index" in the source file name and not in the resulting path I had to rename the file and all links to it.

tharders:
Yes, there are no other pages with this problem.
- No page with path: index.html but not named index.md
- Also no page exists that is hidden:true and not an index.md

PR #2262, opened 2019-06-05
Fix sitemap 404

tharders:
The sitemap.xml which is mentioned in the robots.txt results in a 404
https://amp.dev/sitemap.xml
Also the grow generated sitemap is not served.
https://amp.dev/sitemap_grow.xml

This PR fixes those issues.

### DIFF
--- a/platform/lib/routers/pages.js
+++ b/platform/lib/routers/pages.js
@@ -20,8 +20,9 @@ const express = require('express');
 const config = require('@lib/config');
 const {Signale} = require('signale');
 const {isFilterableRoute} = require('@lib/common/filteredPage');
-const {project} = require('@lib/utils');
+const project = require('@lib/utils/project');
 
+const PATH_HAS_FILE_EXTENSION_PATTERN = /\/[^\/]+\.[a-z0-9]{1,5}$/i;
 
 // eslint-disable-next-line new-cap
 const pages = express.Router();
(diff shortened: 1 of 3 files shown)


platform/lib/routers/pages.js
```diff
@@ -56,7 +56,7 @@ function ensureFileExtension(path) {
     return path += 'index.html';
   }
 
-  if (!path.endsWith('.html')) {
+  if (!path.endsWith('.html') && !path.endsWith('.xml')) {
```
sebastianbenz: Can we fix this more generically by only appending `.html` if there's no other file extension?

tharders: Changed that and added unit tests for the logic.

platform/lib/routers/pages.js
```diff
@@ -56,7 +57,7 @@ function ensureFileExtension(path) {
     return path += 'index.html';
   }
 
-  if (!path.endsWith('.html')) {
+  if (!PATH_HAS_FILE_EXTENSION_PATTERN.test(path)) {
```
sebastianbenz: I'd prefer to use `path.extname` here instead of a regex.

```
const extension = path.extname(filePath)
if (!extension) {
   ...
}
```

tharders: Done

PR #2263, opened 2019-06-05
Do not build grow tests for production

tharders:
The grow test contents should not be deployed on prod.
We simply exclude the path /tests/ in the grow build.

On staging the files will still be available:
https://amp-dev-staging.appspot.com/tests/examples/import

### DIFF
--- a/platform/lib/config.js
+++ b/platform/lib/config.js
@@ -28,6 +28,7 @@ const GROW_CONFIG_DEST = utils.project.absolute('pages/podspec.yaml');
 const ENV_DEV = 'development';
 const ENV_STAGE = 'staging';
 const ENV_PROD = 'production';
+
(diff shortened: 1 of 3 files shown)


platform/lib/config.js
```diff
@@ -128,9 +129,16 @@ class Config {
...
   /**
    * Builds a podspec for the current environment and writes it to the Grow pod
+   * @param {Object} settings Options to filter grow pages (optional). Can be overwritten by command line options.
    * @return {undefined}
    */
-  configureGrow() {
+  configureGrow(settings) {
+    const options = {};
+    if (settings) {
+      Object.assign(options, settings);
```
matthiasrohmer: I find options vs. settings a little confusing 😁

PR #2264, opened 2019-06-05
Fix margin problems in important-tip and help section

dlemm:
The p-tag of the important-tip gets a "margin-top 0" in order to let the tip appear in accordance with other tips.

the help organism gets a "margin-top: 2rem" in order to add space between it and the content section.

### DIFF
--- a/frontend/scss/components/organisms/help.scss
+++ b/frontend/scss/components/organisms/help.scss
@@ -22,6 +22,8 @@
 
 .#{organism('help')} {
 
+  margin-top: 2rem; 
+
   &-hl {
     @include hl;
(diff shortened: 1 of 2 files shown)


frontend/scss/components/organisms/help.scss
```diff
@@ -22,6 +22,8 @@
 
 .#{organism('help')} {
 
+  margin-top: 2rem;
```
matthiasrohmer: Can we not mix units here? The other properties in this component seem to use `px` instead of `rem`.

matthiasrohmer:
Closed in favor of #2318.

PR #2265, opened 2019-06-05
Align path "optimize-and-measure" and improve redirect logic

tharders:
Fixes https://github.com/ampproject/docs/issues/2261

**Improve redirect logic:**
Before the redirect mapping did only look for the complete path
and redirected to the target when a complete match was found.
Now we also look for the path without the locale path and use the locale for the redirect.
We also append the original query parameters to support the format=websites param

That means the shortcuts amp.dev/es/docs would work

### DIFF
--- a/platform/lib/middleware/redirects.js
+++ b/platform/lib/middleware/redirects.js
@@ -50,8 +50,9 @@ function getRedirectLink(req) {
       }
     }
   }
-  if (result && req.query) {
-    result = result + '?' + url.parse(req.url).query;
+  const queryString = url.parse(req.url).query;
+  if (result && queryString) {
+    result = result + '?' + queryString;
   }
   return result;
 }
--- a/platform/lib/middleware/redirects.js
+++ b/platform/lib/middleware/redirects.js
@@ -17,6 +17,7 @@
 'use strict';
 
 const {readFileSync} = require('fs');
+const url = require('url');
 const {join} = require('path');
 const yaml = require('js-yaml');
 const config = require('@lib/config.js');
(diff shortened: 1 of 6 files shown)


platform/lib/middleware/redirects.js
```diff
@@ -28,6 +29,33 @@ const WWW_PREFIX = 'www.';
...
 
+const AVAILABLE_LOCALES = config.getAvailableLocales();
+const LANGUAGE_PATH_PATTERN = /^\/([a-z]{2}(_[a-z]{2})?)\//;
+
+/**
+ *
+ * @param req
+ * @returns The redirect link or undefined if none was found.
+ */
+function getRedirectLink(req) {
```
sebastianbenz: It'd be good to have a test for this

tharders: So true.
I hesitated yesterday, because I wanted to have specific unit test rules and it was not obvious how to do that.
Will not happen again :-)

platform/lib/middleware/redirects.test.js
```diff
@@ -0,0 +1,85 @@
+const express = require('express');
+const request = require('supertest');
+
+// The languages used here have to be in the config.js
+// We do not mock the config, since it is not likely to change
+const VALID_LANGUAGE_SHORT = 'en';
+const VALID_LANGUAGE_LONG = 'pt_br';
+
+const UNIT_TEST_REDIRECTS = '/shortcut: /some/deep/link/\n' +
```
sebastianbenz: nit: JS template strings work better here

```
const UNIT_TEST_REDIRECTS = `
/shortcut: /some/deep/link/
/${VALID_LANGUAGE_SHORT}/old/link: /new/link/target
`
```

tharders: Good idea!
I changed that.

PR #2266, opened 2019-06-05
Fix/sidebar

sebil:
This PR brings in a lot of fine tuning and fixes for the sidebar, toc and target styling when using the toc.
The main feature is the animating sidebar-toggle.
It slides along the edge back and forth with the original animation speed and easing.

### DIFF
--- a/frontend/scss/components/organisms/toc.scss
+++ b/frontend/scss/components/organisms/toc.scss
@@ -74,6 +74,13 @@
     align-items: center;
     justify-content: space-between;
     outline: 0;
+
+    @media (min-width: 1024px) {
+      position: -webkit-sticky;
+      position: sticky;
+      top: 0;
+      background: #fafafc;
+    }
   }
 
   ul {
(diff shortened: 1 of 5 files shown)


sebil:
@sebastianbenz could you review this again?

> [...]
> One nit: ideally, the button should slide out with the sidebar and not appear before.
> [...]

https://github.com/ampproject/docs/pull/2253#pullrequestreview-245447984

frontend/scss/components/organisms/toc.scss
```diff
@@ -48,13 +48,14 @@
...
+    margin: 0;
     height: auto;
-    top: 114px;
+    top: 90px;
     right: auto;
     background: none;
     box-shadow: none;
     overflow-y: scroll;
-    max-height: calc(100vh - 214px);
+    max-height: calc(100vh - 90px);
```
sebastianbenz: nit: I assume the 90px is related to the top value, in that case it'd be good to extract the value into a variable.

sebil: Thanks for the feedback!
That's right, I also thought about extracting all that spacing calculations into variables, but this seems to me like an own new task.
It depends on individual font-sizes and other metrics.
I'll create a new issue for this.

sebil: There it is: #2272

PR #2275, opened 2019-06-05
Update allowed form input elements

westonruter:
Remove `<input type=password>` and `<input type=file>` from disallowed list, since they are now allowed.

See https://amp.dev/documentation/components/amp-form#inputs-and-fields


sebastianbenz:
Thanks for the fix Weston! Unfortunately, these are imported from the ampproject/amphtml repo (`spec/amp-html-format.md`). Can you please create a PR there. Thanks!

westonruter:
Thanks. Opened PR: https://github.com/ampproject/amphtml/pull/22725

PR #2299, opened 2019-06-11
♻️ Update echo API endpoint

lluerich:
Since the echo API has already bin migrated, this just updates the references in [this](https://amp.dev/documentation/examples/interactivity-dynamic-content/advanced_server_request/?format=email) example.


fstanis:
Hi @lluerich, could you put `<% hosts.backend %>` back? Emails must use absolute URLs, so this broke the sample.

Issue #2094, opened 2019-05-15
Restore missing features on roadshows map


morsssss:
As per #2085 : 

The Roadshow Cities map on https://amp.dev/events/amp-roadshow lacks the dots I added on ampproject.org for Africa. Would love to get those back.

We used to add in new cities by creating a dot for each city that was absolutely positioned to resemble its actual location. This system was imperfect but worked. Can we get that, or something like it, back?

I also liked that dots for upcoming cities used to flash... 😄

matthiasrohmer:
@keevkeev, is that something you could take care of?

keevkeev:
@matthiasrohmer yeah, I'll have a look at a nice solution

sebil:
Hey @morsssss I've created a PR for this #2310 
I've tried to find the same spots for the cities as before and added the missing ones for africa freestyle.
Feel free to adjust the coordinates of all cities how you'd like it the best.

PR #2310, opened 2019-06-13
Re-add city markers to roadshow map

sebil:
With this PR, the roadshow map is easy to maintain again.

For each location (next and previous stop) a value for the coordinates can be added (eg 57x11).
If the 'coords' value exist, a point is displayed on the map with a small tooltip that is displayed on hover.
'past' stops are small and light gray, with 'nextstop' points being larger and whiter and having a vibrant glossy animation.

In addition, there is a new feature that makes it easy to find the exact coordinates of the desired position for a particular point.
This feature is only visible in the development environment.

I think this new solution makes it much more intuitive to maintain the roadshow map than playing around with absolute position values.

@morsssss please let me know how you like it :)

#2094


pages/content/amp-dev/events/amp-roadshow.html
```diff
@@ -63,13 +63,78 @@ <h2 class="ap-o-stage-content-subline">Come join us in your city for a free, inc
...
+        <button
+          on="tap:AMP.setState({coordsGridToggled: !coordsGridToggled})"
+          [text]="coordsGridToggled ? 'Hide grid' : 'Show grid'"
+          type="button">
+          Show grid
+        </button>
+        </h4>
+        <div hidden [hidden]="!coordsGridToggled">
+          <strong><code>Coordinates: <span [text]="coords"></span></code></strong>
+          <p>You can copy this coordinates and use it in the <strong>roadshow.yaml</strong> <em>(pages/shared/data/roadshow.yaml)</em></p>
```
morsssss: Little nit: "You can copy this coordinates and use it in the" -> "You can copy these coordinates and use them in"

pages/shared/data/roadshow.yaml
```diff
@@ -16,70 +18,103 @@ nextstop:
...
+    coords: 8x13
     date: JUN 12, 2018
   - title: AUSTIN
+    coords: 16x19
     date: JUL 11, 2018
   - title: NEW YORK
+    coords: 26x15
     date: JUL 24, 2018
   - title: CAMBRIDGE
+    coords: 50x11
```
morsssss: Heh heh.. I was surprised to see two dots in the British Isles. And then I realized... you'd chosen Cambridge, England! As opposed to Cambridge, Massachusetts, USA, where we actually had the roadshow.

Cambridge is basically part of Boston (though don't tell them that I said that). Why don't we just change this to Boston - and change the coordinates accordingly?

pages/shared/data/roadshow.yaml
```diff
@@ -16,70 +18,103 @@ nextstop:
...
   - title: HO CHI MINH CITY
+    coords: 88x24
     date: DEC 14, 2018
   - title: TORONTO
+    coords: 24x12
     date: AUG 30, 2017
   - title: SINGAPORE
+    coords: 86x29
     date: OCT 6, 2017
   - title: JAKARTA
```
morsssss: Singapore and Jakarta appear to both be on the same island on this map. I'm not sure what all the little stylized islands are supposed to be, but it would be nice if Singapore were somehow part of the mainland. As it's on a peninsula.

pages/shared/data/roadshow.yaml
```diff
@@ -16,70 +18,103 @@ nextstop:
...
+    coords: 15x22
     date: MAR 21, 2018
   - title: CHICAGO
+    coords: 20x13
     date: MAR 27, 2018
   - title: SUNNYVALE
+    coords: 8x16
     date: APR 4, 2018
   - title: TAIPEI
+    coords: 92x21
```
morsssss: Taipei's in the water now. Can we get it on to one of the little cute blue stylized islands?

morsssss:
I like this. This is great!
In particular, it seems much more logical to add our coordinates to the YAML instead of adding CSS for each city.

One request: would you consider making the big flashing dots for upcoming cities a color that's a bit more distinct from the color of the ocean around them?

Also, this is not part of your PR, but as long as you're at it... it would be easier to read the past cities if they were ordered by rows, not columns. Don't know whether that's easy to implement...

Thanks so much - I'll be so glad to have this map back!

sebil:
Nice, I'm glad you liked it!
Like I said, feel free to change the coordinates as you want.
And please don't mind the wrong grammar, I'm actually happy to have native english speakers on the other side who see those nits :)

Sure, I'll apply the latest changes and update this PR.

pages/shared/data/roadshow.yaml
sebil: At least I know now what I better not say, if I travel there one day… 🙊

pages/content/amp-dev/events/amp-roadshow.html
sebil: Thanks for this. Fixed it.

pages/shared/data/roadshow.yaml
sebil: I changed it to Boston now and set the dot slightly above NYC.

pages/shared/data/roadshow.yaml
sebil: Singapore is now on the mainland

pages/shared/data/roadshow.yaml
sebil: I moved Taipei down to the next island, but it seems it is in Indonesia now and not Taiwan. Maybe we should just update the cities-map.svg itself and add more island?

sebil:
I've applied all requested changes except of the flashing dot color. I don't really understand what you mean. Should the white dot be more grayish? Or just the flash? Or should the dot have an overall darker look?

morsssss:
Well, I'm not a designer... but I was thinking something more bright yellow would stand out more and look thrilling. Of course this may violate all your site design principles, in which case, please ignore me 😎 

Thanks for making all those changes. Actually, now that the past cities progress horizontally instead of vertically, I can tell for the first time that they're in two different chronological orders! As in, at first they start from the most recent and progress backward in time, and then suddenly we jump to the very first one (Toronto) and progress forward.

I feel bad asking you to fix that since it's not something you did, so if you want, I can do it after we merge this.

sebil:
Hey Ben, our designer has giving me a new suggestion for the next cities pin, let yourself be surprised ;)
Of course I'll take care of the order of the cities, bad enough, that I didn't recognized it -.-

sebil:
I've pushed the changes now, so you can have a look at them. The only thing that is missing, is the flashing animation, I've commented that out and will take care of it next week.
The visited cities are now correctly sorted.
Let me know if I missed anything :)

morsssss:
Cool! I'll look forward to the flashing animation too :)

Once this is in prod, I'll use it to add four new cities! LGTM

sebil:
Hey Ben @morsssss,
the next-stop animation is now pushed.
Also I've made some more improvements in collaboration with our designer Longo.
I know this is already merged, but you can still have a look at the changes and let me know if it's ready for prod.
If yes, I will move Berin and Atlanta to past stops.

morsssss:
Oo, cool animation!

It looks ready for prod to me.

You can move those cities to the past if you want - but I'd planned to follow your PR with one of my own that would move those to the past and add the four next cities. I also want to add to this page a link to a form people can fill out if they want to either request that we come to their city or run a roadshow of their own.

I could of course add the new cities to your PR if you want...

sebil:
Hi @morsssss, I've created a new PR for this, see #2394

PR #2311, opened 2019-06-13
:bug: Set header's sticky top value to zero

sebil:
Since the banner is gone, the header scroll's outside of the viewport


sebil:
Hotfix

PR #2312, opened 2019-06-13
Hotfix/header without banner

sebil:
Since the banner is gone, the header scroll's outside of the viewport


sebil:
Mistakely splitted from future. Closing here.

PR #2314, opened 2019-06-13
Correct autocomplete redirect path

caroqliu:


nainar:
Approving and merging since @sebastianbenz and @CrystalOnScript are out.

Issue #2234, opened 2019-06-03
Tools page: support logos and improve description handling


sebastianbenz:
We need a flag in the template/data do define wether the image is a screenshot or logo (and update the margins accordingly). 

@pbakaus can you elaborate what you mean by description handling?

![](https://user-images.githubusercontent.com/43004/58677177-d4d52f80-830f-11e9-97ba-7bfe8b25cbf1.png)

See #2217 for more context.

@sebil can you take this on?

pbakaus:
re: description handling: Right now the cards look pretty stupid without a description (imho). The email team has requested to not have descriptions in the email service provider cards (so they can be fairly compared), but in other sections further below, the descriptions add value. It would be nice to tweak the design so that certain cards look great with just a logo and title.

Also: The email team requested "Email service providers" as overall category, as opposed to slotting these into "Creation tools". @sebil feel free to take a stab at this as well, I think it might also then be a good time to move the filtering to server-side, so that we can link to pre-filtered pages.

Thanks!

sebil:
WIP

PR #2316, opened 2019-06-14
Tweak tools page styling and structure

sebil:
With this PR the tools page will have a better look & feel for those cards that only have a logo and a headline.
When an image has the 'fit: contain' property it now automatically gets a padding, so I don't think we need an extra flag here. @sebastianbenz 

Also I've separated 'Sparkpost' and 'Twilio Sendgrid' to a new category called 'Email service providers' as requested, but in my opinion now the filtering makes no sense anymore.
This new category could only display cards with the email label, which means that it is always empty as another filter is active.

Our suggestion to prevent this is to rename the new category to just 'Service providers' and eg. add 'Signed exchange' service by Cloudflare to the section, to have cards of multiple formats again.

What do you think about that @pbakaus ?

FYI: Don't worry about the 3 commits related to the roadshow-map-grid. I've worked mistakenly on the wrong branch and reverted them again.

#2234


pbakaus:
Thanks!

"Service providers" category seems decent, let's try that. Also, could you paste a screenshot here so anyone reviewing could take a quick glimpse? I recommend doing this for all PRs with big visual impact.

Thanks!
Paul

sebil:
I'm glad you like it.
Sure, here is a recent screenshot of the whole tools page (excluding the footer):
screenshot-tools

Issue #2308, opened 2019-06-13
Help section position problem in small screen sizes


tharders:
The help section at the bottom of the component and example pages is not positioned well when viewing the page with a lower resolution.
Below 1024 pixel the margin to the content above is to small.
Below 768 pixel the alignment to the left and right is not the same as the content above (but in a different way for components and examples)
https://amp.dev/documentation/components/amp-yotpo?format=websites
https://amp.dev/documentation/examples/guides/internationalization/?format=websites

![image](https://user-images.githubusercontent.com/6458408/59425049-73b06000-8dd5-11e9-92eb-fc8874d428e7.png)

Seen on Mac with latest Chrome.

tharders:
The alignment to the left and right is still not the same as the content above for both pages on small screen widths.
The alignment to the top is also not the same on both pages. That is because on the example pages the help section uses "ap--content" as class while the components uses the "ap--help".
With "ap-content" the space above looks a little bit too much, but it matches the space from the divider. I do not know how much space is correct, but both pages should have the same.

![image](https://user-images.githubusercontent.com/6458408/59912736-f27d4c80-9416-11e9-9a04-73bd2979cd39.png)

PR #2318, opened 2019-06-14
🐛 Add and edit padding to the help section

### LINKED ISSUE #2112: Component docs: add margin below validation section
See screenshot:

Screen Shot 2019-05-16 at 20 16 34

dlemm:
fix #2112 and #2308 add padding-top to the help-Section and on mobile remove side-padding from help section on component-detail sites

### DIFF
--- a/frontend/scss/components/organisms/help.scss
+++ b/frontend/scss/components/organisms/help.scss
@@ -22,7 +22,7 @@
 
 .#{organism('help')} {
 
-  padding-top: 2rem;
+  padding-top: 32px;
 
   &-hl {
     @include hl;
--- a/frontend/scss/components/organisms/help.scss
+++ b/frontend/scss/components/organisms/help.scss
@@ -22,6 +22,8 @@
 
 .#{organism('help')} {
 
+  padding-top: 2rem;
+
   &-hl {
     @include hl;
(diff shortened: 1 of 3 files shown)


frontend/scss/components/organisms/help.scss
```diff
@@ -22,6 +22,8 @@
 
 .#{organism('help')} {
 
+  padding-top: 2rem;
```
matthiasrohmer: Can we not mix units here? The other properties in this component seem to use px instead of rem.

frontend/scss/components/templates/component-detail.scss
```diff
@@ -33,6 +33,8 @@ body {
   .#{utility('help')} {
     @include content;
     @include content-white-bg;
+    padding-left: 0px;
```
matthiasrohmer: Omit `px` as zero is zero no matter the unit.

frontend/scss/components/organisms/help.scss
dlemm: changed

frontend/scss/components/templates/component-detail.scss
dlemm: corrected

Issue #2309, opened 2019-06-13
Hides menu on support page because footer comes up


Amanrazdan:
- Page link: https://amp.dev/support/
-  when going to the support page because the content is less it collapses it up and when we navigate menu they are under them

Please provide the steps to reproduce the issue:

1. visit the link https://amp.dev/support/
2. hover over other menu options

browser - chrome
os- windows 10

this can be easily fixed by sticking footer to bottom or increasing z-index

sebil:
Thanks for reporting this, I'll take care of this

Issue #2272, opened 2019-06-05
Extract position/calculation values into variables


sebil:
There are a lot of spacing calculations and fixed position values to fit things like sidebar, toc, toggles, etc. perfectly to the page.
All this values could be extracted into variables and functions to keep the code dry and better maintainable.

> nit: I assume the 90px is related to the top value, in that case it'd be good to extract the value into a variable.

_Originally posted by @sebastianbenz in https://github.com/ampproject/docs/pull/2266_

sebil:
I will take care of this as soon as no higher prioritised issues existing anymore.

PR #2322, opened 2019-06-14
Fix header and navigation styling issues

sebil:
- Definitly fixes one bug where the footer overlays the dropdown navigation on pages with less content, see #2309 
- Also fixes/improves some issues that happens after temporally hiding the banner
- Furthermore starting with extracting specific values into variables, see #2272

### DIFF
--- a/frontend/scss/_variables.scss
+++ b/frontend/scss/_variables.scss
@@ -20,6 +20,9 @@ $sizes: (
   xl: '5',
 );
 
+$mobileHeaderHeight: 55px;
+$bannerHeight: 31px;
+
 /* The following map contains every color that is used inside the project. The
 names are identified by using the http://chir.ag/projects/name-that-color to
 prevent double colors and/or hirarchy dicussions */
--- a/frontend/scss/_variables.scss
+++ b/frontend/scss/_variables.scss
@@ -20,7 +20,7 @@ $sizes: (
   xl: '5',
 );
 
-/* This variables where used in the following files:
+/* This variables are used in the following files:
 - frontend/scss/components/molecules/banner.scss
 - frontend/scss/components/organisms/header.scss
 - frontend/scss/components/organisms/burger-menu.scss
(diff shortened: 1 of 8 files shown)


frontend/scss/_variables.scss
```diff
@@ -20,6 +20,9 @@ $sizes: (
   xl: '5',
 );
 
+$mobileHeaderHeight: 55px;
```
matthiasrohmer: Please add a comment where and why those variables are used.

frontend/templates/views/partials/header.j2
```diff
@@ -8,9 +8,15 @@
 {% if doc.pod_path.startswith(search_path) %}ap-m-nav-link-active{% endif %}
 {%- endmacro %}
 
+{# Simply set to False to disable banner #}
+{% set show_banner = True %}
+{% set banner %}
+{% include '/views/partials/banner.j2' %}
```
matthiasrohmer: This always includes (and loads) the banner template even if it will not be shown.

sebil: You absolutely right, fixed this with an if condition.
Also I forgot changing show_banner back to False.

frontend/scss/_variables.scss
sebil: Please check if the suggested comment will do it, otherwise let me know.

frontend/scss/_variables.scss
```diff
@@ -20,6 +20,16 @@ $sizes: (
   xl: '5',
 );
 
+/* This variables where used in the following files:
```
matthiasrohmer:
```suggestion
/* This variables are used in the following files:
```

frontend/scss/_variables.scss
```diff
@@ -20,6 +20,16 @@ $sizes: (
...
+- frontend/scss/components/molecules/banner.scss
+- frontend/scss/components/organisms/header.scss
+- frontend/scss/components/organisms/burger-menu.scss
+
+The calculations of the height and top values of banner, header and burger-menu
+are based on this variables. If you change them, these it automatically fits. */
```
matthiasrohmer:
```suggestion
are based on this variables. If you change them, they automatically adjust. */
```

Issue #2295, opened 2019-06-10
Design to toggle "compressed/experienced developer" view of courses


morsssss:
Now the courses are live on amp.dev. Yay!

These courses assume the reader may not have much experience with webdev beyond knowing some HTML and CSS. There's been some demand for a compressed version of the courses for more experienced webdevs, so these folks can take the courses more quickly and won't feel turned off. We're making the compressed versions now. They'll omit the bits that explain what an API is, what web components are, etc., Another example: they'll define the cache more quickly.

The question is: how can these be presented on amp.dev?

Here's my thought: what if each page loads up both versions, and readers can use a little button to toggle between them? Ideally we could use a cookie or query param to persist the state between pages.

Normally I like cookies, but a query param might be ideal because someone might really want to share a specific view of a specific page. I don't know which is simpler to implement for us.

Or is there an easier/prettier way to do this?

Thanks!

Looking for comments here from @pbakaus , @matthiasrohmer , @sebastianbenz, @CrystalOnScript , etc.!

pbakaus:
I like this! We run into this issue sometimes in the rest of the docs too. We could trial this functionality in the courses, and then expand it to other docs if it works well, potentially.

@sebastianbenz @matthiasrohmer thoughts on how we'd best implement this?

matthiasrohmer:
Good idea! The easiest way to implement this is when both versions of the document would still be maintained in the same markdown file and the differentiation between easy and advanced would only happen by some kind of tagging. Similar to what we enabled for formats like with `[filter formats="stories"]This bit is only visible for stories![/filter]`.

This exact shortcode is easily extendable with a `level` option for example: `[filter level="easy"]This won't show for pro-programmers[/filter]`.

Server-side filtering via GET-parameter would of course then work best. Saves us from all the AMP cache hurdles and make things shareable as you mentioned.

Not necessarily blocked by #1694 but this would better be implemented first as the filtering during build-time is already kinda fragile and I'd like to hold back from adding any more logic into that. But I'll make this my next task so we can also continue on this one here.

morsssss:
That sgtm too!

I was thinking that each page would load up *both* versions, and that the user could toggle between them by clicking a button that would change state... almost like tabs. And that we'd update the URL accordingly using `pushState()`. This should be reasonable performance-wise for the courses, since both versions will probably use all the same images.

[...]

PR #2325, opened 2019-06-16
♻️ Rewrite pages router to enable SSR.

### LINKED ISSUE #2180: Trailing slash woes
I realized that forgetting a trailing slash causes 404, e.g.

Valid: https://amp.dev/documentation/examples/components/amp-accordion/
Invalid: https://amp.dev/documentation/examples/components/amp-accordion

Would be good to make it instead redirect to the correct (canonical) page.

matthiasrohmer:
Fixes #2180 and (build) performance issues related to format-filtering via cheerio.js, foundation for #2295, #1694.


platform/lib/templates/index.js
```diff
@@ -97,8 +118,14 @@ class Templates {
   async fetchTemplate_(templatePath) {
     const templateUrl = new URL(templatePath, config.hosts.pages.base);
     const fetchResponse = await fetch(templateUrl);
-    return fetchResponse.text();
+
+    if (fetchResponse.status && fetchResponse.status !== 404) {
+      return fetchResponse.text();
+    }
```
tharders: Shouldn't there be a return or an error if the condition does not match

platform/lib/templates/index.js
```diff
@@ -27,11 +27,32 @@ const {pagePath, paths} = require('../utils/project');
...
+ * Builds a Object that adds mandatory variables to the render context
+ * that are needed to successfully SSR the pages
+ * @param  {expressjs.Request} request
+ * @return {Object}
+ */
+function context(request, base) {
+  request = request || {'query': {}};
+  const context = {} || base;
+
+  const ALLOWED_FORMATS = ['websites', 'stories', 'ads', 'email'];
```
sebastianbenz: Templates.js feels like the wrong place for this. Can we move this to the caller?

platform/lib/templates/index.js
```diff
@@ -97,8 +118,14 @@ class Templates {
   async fetchTemplate_(templatePath) {
     const templateUrl = new URL(templatePath, config.hosts.pages.base);
     const fetchResponse = await fetch(templateUrl);
-    return fetchResponse.text();
+
+    if (fetchResponse.status && fetchResponse.status !== 404) {
```
sebastianbenz: use `fetchResponse.ok` instead

platform/lib/routers/pages.bak.js
```diff
@@ -0,0 +1,189 @@
+/**
```
sebastianbenz: Remove this file as it's stored in the revision history.

platform/lib/templates/index.js
matthiasrohmer: My reasoning was that potentially any template that is loaded here could rely on having `format` defined. If we move it to the caller we'd need to define it in both: 

https://github.com/ampproject/docs/blob/7c944d4f14e44d8e74282e117a086ed96931edea/examples/source/1.components/amp-live-list/api.js#L22

and 

https://github.com/ampproject/docs/blob/7c944d4f14e44d8e74282e117a086ed96931edea/platform/lib/routers/pages.js#L162

platform/lib/templates/index.js
matthiasrohmer: `Response.ok` would be `false` if Grow returned a error debug page which would have a 500 status but you'd like to return to the user. That's why I went for the more verbose status checking.

sebastianbenz: add a comment in that case

matthiasrohmer:
@sebastianbenz, accomodated your requested changes, only waiting for feedback on one. Additionally  I made some small last changes: when ensuring the URL scheme the redirects are now happening by 301s instead of 302s, all tests pass now and the canonical URL is correct.

pages/content/amp-dev/documentation/guides-and-tutorials/index.html
```diff
@@ -1,26 +1,41 @@
...
 $titles:
   navigation: Guides & Tutorials
 $view: /views/overview/guides.j2
-$path: /documentation/guides-and-tutorials/index.html
-$localization:
-  path: /{locale}/documentation/guides-and-tutorials/index.html
 $order: 1
 $hidden: true
-description: 'All resources you need to build stunning websites with AMP: Guides, tutorials, reference documentation and examples.'
+description: 'All resources you need to build stunning [[ format ]] with AMP: Guides, tutorials, reference documentation and examples.'
```
sebastianbenz: This won't work for email: `build stunning email` with AMP

pages/content/amp-dev/documentation/guides-and-tutorials/index.html
```diff
@@ -1,26 +1,41 @@
...
 flyout:
   description: Get started with AMP
   icon: guides-documents
 ---
 <section class="ap--section">
   <h1>{{ doc.title }}</h1>
-  <p>Here you can find everything you need to build AMP experiences. The sidebar on the left shows you all resources for building <strong>AMP websites.</strong></p>
+  <p>Here you can find everything you need to build AMP experiences. The sidebar on the left shows you all resources for building <strong>AMP [[ format ]].</strong></p>
```
sebastianbenz: same as above

matthiasrohmer:
Thanks for the review, @sebastianbenz! Especially a good catch with the email plural thing.

matthiasrohmer:
Something I thought about: should we use LRU here as well over a Map, as the later one would need to hold 2GB in the worst case.

https://github.com/ampproject/docs/blob/dd2b357d8aaee6ed25413953783b5a7498a195da/platform/lib/templates/index.js#L78

sebastianbenz:
@matthiasrohmer good catch, we definitely should use an LRU cache here (not sure what a good cache size would be though). It's ironic that we now run into these kinds of problems we wouldn't have with a purely server-side-rendered or static site...

matthiasrohmer:
Ran a test build to check: one locale with 862 pages weighs 82MB in total, making 984MB in total for all locales. Our VMs seem to be n1-standard-2 which have a total of 7,5GB memory though Node.js won't get it all and might be [limited at maybe 1,4GB]. 

So I assume fitting three full locales (246MB) into the LRU will not cause any problem. That would mean capping it at 2586 entries. WDYT?

matthiasrohmer:
@sebastianbenz, please take another look.

matthiasrohmer:
Added another context variable `forceFiltered` that is true if no valid format is given as GET parameter. That's needed for the tools page for example to be able to show all tools if no explicit format is given.

pages/content/amp-dev/documentation/guides-and-tutorials/index.html
```diff
@@ -1,26 +1,41 @@
...
 $titles:
   navigation: Guides & Tutorials
 $view: /views/overview/guides.j2
-$path: /documentation/guides-and-tutorials/index.html
-$localization:
-  path: /{locale}/documentation/guides-and-tutorials/index.html
 $order: 1
 $hidden: true
-description: 'All resources you need to build stunning websites with AMP: Guides, tutorials, reference documentation and examples.'
+description: "All resources you need to build stunning [[ format if format !== 'email' else 'emails' ]] with AMP: Guides, tutorials, reference documentation and examples."
```
sebastianbenz: Maybe use a helper `pluralize` instead.

sebastianbenz: s/build/create

platform/lib/routers/pages.js
```diff
@@ -17,173 +17,169 @@
...
-  });
+  // As the request path is not the actual path to the template it is somehow
+  // guessed by testing all of AVAILABLE_STUBS ...
+  for (const stub of AVAILABLE_STUBS) {
+    // Othwerwise try the first stub or the already resolved path if there is one
+    const searchPath = resolvedPath || `${templatePath}${stub}`;
+    try {
+      template = await Templates.get(searchPath);
+    } catch (e) {
+      continue;
```
sebastianbenz: add a comment why this is expected.

platform/lib/routers/pages.js
```diff
@@ -17,173 +17,169 @@
...
 
-  // Also create a logger during development since you want to know
-  // what's going on
-  const log = new Signale({
-    'interactive': false,
-    'scope': 'Grow (Proxy)',
-  });
+  // As the request path is not the actual path to the template it is somehow
+  // guessed by testing all of AVAILABLE_STUBS ...
+  for (const stub of AVAILABLE_STUBS) {
```
sebastianbenz: Moving this part into a method makes it easy to implement this without `continue` and `break` making it a lot easier to follow the control flow.

PR #2337, opened 2019-06-18
♻️ Migrate amp-access API to amp.dev

lluerich:
Migrated the backend/amp-access.go from ampbyexample.com to amp.dev.
This updates the example [here](https://amp.dev/documentation/examples/components/amp-access/?format=websites) and [here](https://amp.dev/documentation/examples/interactivity-dynamic-content/comment_section/?format=websites).

### DIFF
--- a/examples/api/amp-access.js
+++ b/examples/api/amp-access.js
@@ -36,9 +36,9 @@ const POWER_USERS = {
 };
 const EXPIRATION_DATE = 24*60*60*1000; // 1 day in ms
 
-examples.all('/amp-access/authorization', handleAuthorization);
-examples.all('/amp-access/login', handleLogin);
-examples.all('/amp-access/logout', handleLogout);
+examples.get('/amp-access/authorization', handleAuthorization);
+examples.get('/amp-access/login', handleLogin);
+examples.get('/amp-access/logout', handleLogout);
 examples.post('/amp-access/submit', handleSubmit);
 
 function handleAuthorization(request, response) {
--- a/examples/api/amp-access.js
+++ b/examples/api/amp-access.js
@@ -35,6 +35,7 @@ const VALID_USERS = {
 const POWER_USERS = {
   'Jane@gmail.com': true,
 };
+const EXPIRATION_DATE = 24*60*60*1000; // 1 day in ms
 
 examples.all('/amp-access/authorization', handleAuthorization);
 examples.all('/amp-access/login', handleLogin);
@@ -89,7 +90,7 @@ function handleSubmit(request, response) {
     response.status(401).send('Invalid email');
     return;
   }
-  response.cookie(AMP_ACCESS_COOKIE, {email}, {expires: new Date(Date.now() + 24*60*60*1000)});
+  response.cookie(AMP_ACCESS_COOKIE, {email}, {expires: new Date(Date.now() + EXPIRATION_DATE)});
   let returnUrl = request.body.returnurl;
   if (!isValidURL(returnUrl)) {
     response.status(500).send('Invalid return URL');
--- a/examples/api/amp-access.js
+++ b/examples/api/amp-access.js
@@ -36,9 +36,9 @@ const POWER_USERS = {
   'Jane@gmail.com': true,
 };
 
-examples.get('/amp-access/authorization', handleAuthorization);
-examples.get('/amp-access/login', handleLogin);
-examples.get('/amp-access/logout', handleLogout);
+examples.all('/amp-access/authorization', handleAuthorization);
+examples.all('/amp-access/login', handleLogin);
+examples.all('/amp-access/logout', handleLogout);
 examples.post('/amp-access/submit', upload.none(), handleSubmit);
 
 function handleAuthorization(request, response) {
--- a/examples/api/amp-access.js
+++ b/examples/api/amp-access.js
@@ -15,11 +15,11 @@
  */
 'use strict';
 
-const path = require('path');
 const express = require('express');
 const cookieParser = require('cookie-parser');
 const nunjucks = require('nunjucks');
 const URL = require('url');
+const utils = require('@lib/utils');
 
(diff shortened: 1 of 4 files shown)


examples/api/amp-access.js
```diff
@@ -0,0 +1,107 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+'use strict';
+
+const path = require('path');
+const express = require('express');
+const cookieParser = require('cookie-parser');
+const multer = require('multer');
+const upload = multer();
```
matthiasrohmer: 😁 As always: is it needed?

examples/api/amp-access.js
```diff
@@ -0,0 +1,107 @@
...
+  'mark@gmail.com': true,
+  'jane@gmail.com': true,
+};
+const POWER_USERS = {
+  'Jane@gmail.com': true,
+};
+
+examples.all('/amp-access/authorization', handleAuthorization);
+examples.all('/amp-access/login', handleLogin);
+examples.all('/amp-access/logout', handleLogout);
```
matthiasrohmer: Do all of these really need to respond to GET and POST requests?

examples/api/amp-access.js
```diff
@@ -0,0 +1,107 @@
...
+examples.post('/amp-access/submit', upload.none(), handleSubmit);
```
matthiasrohmer: I'd remove this whole block (all route registrations) down to the file as they are relying on the functions that should be defined beforehand.

examples/api/amp-access.js
```diff
@@ -0,0 +1,107 @@
...
+  });
+}
+
+function handleLogin(request, response) {
+  const returnUrl = request.query ? request.query.return : '';
+  if (!isValidURL(returnUrl)) {
+    response.status(500).send('Invalid return URL');
+    return;
+  }
+  const filePath = path.join(__dirname, '../static/samples/files/login.html');
```
matthiasrohmer: Please use `utils.project.absolute` here 🙂 This way the sample is save to be moved around.

examples/api/amp-access.js
```diff
@@ -0,0 +1,107 @@
...
+  response.json({
+    loggedIn: true,
+    powerUser,
+    email,
+    name: email.split('@')[0],
+  });
+}
+
+function handleLogin(request, response) {
+  const returnUrl = request.query ? request.query.return : '';
```
matthiasrohmer: How could an invalid URL ever land here?

examples/api/amp-access.js
```diff
@@ -0,0 +1,107 @@
...
+  response.status(303).redirect(returnUrl);
+}
+
+function handleSubmit(request, response) {
+  const email = request.body ? request.body.email : '';
+  if (!VALID_USERS[email]) {
+    response.status(401).send('Invalid email');
+    return;
+  }
+  response.cookie(AMP_ACCESS_COOKIE, {email}, {expires: new Date(Date.now() + 24*60*60*1000)});
```
matthiasrohmer: Add a comment how long the cookie is valid.

examples/source/interactivity-dynamic-content/Comment_Section/index.html
```diff
@@ -136,11 +136,11 @@
     This sample allows an user to login and logout using an email and a password. Logout is implemented by configuring a second endpoint in the login property `sign-out`, find more [here]({{g.doc('/content/amp-dev/documentation/components/reference/amp-access.md', locale=doc.locale).url.path}}#login-page).  -->
     <script id="amp-access" type="application/json">
     {
-        "authorization": "<% hosts.backend %>/components/amp-access/authorization?rid=READER_ID&url=CANONICAL_URL&ref=DOCUMENT_REFERRER&_=RANDOM",
+        "authorization": "/documentation/examples/api/amp-access/authorization?rid=READER_ID&url=CANONICAL_URL&ref=DOCUMENT_REFERRER&_=RANDOM",
         "noPingback": "true",
         "login": {
-          "sign-in": "<% hosts.backend %>/components/amp-access/login?rid=READER_ID",
-          "sign-out": "<% hosts.backend %>/components/amp-access/logout?rid=READER_ID"
+          "sign-in": "/documentation/examples/api/amp-access/login?rid=READER_ID",
```
matthiasrohmer: What I find a bit confusing now: some component APIs ended up in `examples/source/1.components/amp-live-list/api.js` for example some in `examples/api/amp-consent.js`. I personally like the former one better as it couples logic and markup - what do you think?

lluerich: Since multiple examples (two atm) are using this backend, the general way of implementing this in the other cases was to put it somewhere central. Example specific APIs are located in the example folder.

examples/api/amp-access.js
lluerich: Moved it to a constant and added a comment.

examples/api/amp-access.js
lluerich: Acutally no, thanks for pointing that out to me. Express has its own parser since v4.16.0.

examples/api/amp-access.js
lluerich: I wasn't sure, but I guess only `/authorization` needs to be POST.

examples/api/amp-access.js
lluerich: Thanks for pointing that out. Consider it done.

examples/api/amp-access.js
lluerich: I just copied that part of the logic from the original GO file, but thinking about it, you are correct. Removed it now.

examples/api/amp-access.js
lluerich: I see your point here, but since the functions are hoisted and it looks cleaner (imo), I wouldn't change it.

PR #2356, opened 2019-06-21
💄 Add styling to static DayPicker in examples

### LINKED ISSUE #2351: amp-date-picker example appears to have mildly broken css?
### Current Behavior
https://amp.dev/documentation/examples/components/amp-date-picker/

Screenshot from my laptop, red annotations mine:

jeYFsQbnVNS

It looks like the Saturday column header has been moved into the Sunday column. The other column headers look a bit awkwardly merged with the date numbers as well.

Google Chrome | 75.0.3770.80 | Linux | Desktop

dlemm:
fixes #2351 add styling (remove padding-left) for the static dayPicker in order to let the days appear in one row above the dates.

### DIFF
--- a/examples/source/1.components/amp-date-picker.html
+++ b/examples/source/1.components/amp-date-picker.html
@@ -85,15 +85,6 @@ author: kul3r4
     margin-top: var(--space-1);
     width: 27px;
   }
-  
-  /* edit inline-code-preview */
-  .ap-o-code-preview .DayPicker_weekHeader {
-    top: 60px;
-  }
-
-  ul.DayPicker_weekHeader_ul.DayPicker_weekHeader_ul_1 {
-    padding-left: 0;
-  }
   </style>
 </head>
 <body>
--- a/examples/source/1.components/amp-date-picker.html
+++ b/examples/source/1.components/amp-date-picker.html
@@ -69,14 +69,6 @@ author: kul3r4
     margin-left: var(--space-2);
   }
 
-  .DayPicker_weekHeader {
-    top: 60px;
-  }
-
-  ul.DayPicker_weekHeader_ul.DayPicker_weekHeader_ul_1 {
-    padding-left: 0;
-  }
-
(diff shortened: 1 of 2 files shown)


examples/source/1.components/amp-date-picker.html
```diff
@@ -69,6 +69,14 @@
     margin-left: var(--space-2);
   }
 
+  .DayPicker_weekHeader {
+    top: 60px;
```
tharders: Please check the stand alone preview and the playground preview.
It's only together with the styles from amp.dev that it was messed up.
The changes fixes the inline preview, but introduces problems in the stand alone example.
When we introduce styles that fix the inline preview, we should add a comment, to make clear that it is only needed for the inline preview on amp.dev
If possible it would be better to change the site styles so that the effects on the inline preview is removed.

dlemm: should be fixed now

examples/source/1.components/amp-date-picker.html
```diff
@@ -85,6 +85,15 @@
     margin-top: var(--space-1);
     width: 27px;
   }
+  
+  /* edit inline-code-preview */
```
matthiasrohmer: Some more details would help 😉What behaviour does that rule prevent specifically - basically try to explain what you would have liked to know when you were starting with this problem.

sebil:
With the last [commit](https://github.com/ampproject/docs/pull/2356/commits/7a2e5146a2f1f9787e5495c644b554c00b445e7d) we prevent our global list stylings to affect any list element that is no direct child of ap--content or ap--section, which fixes the issue of overriding the <ul> inside the datePicker extension.

sebil:
@pbakaus could you please review and merge these changes.
We're stuck because of the 'verify author consent' check from the google bot.
I made 2 commits on Dennis PR but forgot to assign him as co-author.

Issue #1860, opened 2019-04-12
Short URLs for AMP.dev


machal:
I'm writing book/e-book about AMP and thinking about shorter URLs for docs pages.

For example:

```
https://amp.dev/documentation/components/amp-pixel
```

Should be (or have redirected variant) similar to the following:

```
https://amp.dev/docs/c/amp-pixel
```

…or even better:

```
https://amp.dev/amp-pixel
```

The problem with long URLs is that they are hard to read and impossible to print. Shorter version (`amp.dev/docs/c/amp-pixel`) looks fine in the printed page.

Of course, I can use `bit.ly` or a similar service, but I think AMP docs should be more friendly by default. After moving to `amp.dev` it is better, but still a problem.

sebastianbenz:
That's a great idea. We've just launched the option to explicitly define short links via go.amp.dev (configured [here](https://github.com/ampproject/docs/blob/future/platform/config/go-links.yaml). 

However, it'd be great to have something generic for compoent docs. For example: go.amp.dev/c/amp-pixel.

//cc @pbakaus for thoughts.

sebastianbenz:
...and awesome that you're writing a book about AMP ⚡️⚡️⚡️

pbakaus:
Indeed great to hear about your book!

`amp.dev/amp-*` is my favorite, should not have any naming conflicts with existing pages. But if we want to be safe, I'm OK with `amp.dev/c/amp-*` as well.

fstanis:
To echo some thoughts from #2182: I think it's better to separate this from `go.amp.dev` and make a new subdomain, e.g. `component.amp.dev/amp-form` or `c.amp.dev/amp-form`.

That said, if this is infeasible on infrastructure side, `go.amp.dev` works, but it would be good to also have a scheme to go to the component example rather than the reference doc. Maybe `go.amp.dev/sample/amp-form`?

machal:
@fstanis Thanks. For me (and my book), `go.amp.dev/c/amp-img` is better, because it is shorter and consistent with other `go.amp.dev` links.

I don't think links to samples of the components are needed. It is simple to navigate there from main component page.

fstanis:
I think there are many cases where short-linking to a sample is very useful, e.g. [in this SO answer](https://stackoverflow.com/a/56178123/2676845) linking to the `amp-lightbox-gallery` sample directly was better than linking to the component, since the sample very clearly demonstrates the feature.

[...]

PR #2358, opened 2019-06-21
Allows go-links.yaml to have regex based redirects

### LINKED ISSUE #2182: Allow go.amp.dev to append the rest of the path to target
Edit: moved part of suggestion to #1860

Would be great to extend `go.amp.dev` to append suffixes, so that e.g. `go.amp.dev/docs/start/create/` would redirect to `https://amp.dev/documentation/guides-and-tutorials/start/create/` if `go.amp.dev/docs` redirects to `https://amp.dev/documentation/guides-and-tutorials`.

This would then allow us to create e.g. `go.amp.dev/component/amp-form` if we only add `go.amp.dev/component` to [go-links.yaml](https://github.com/ampproject/docs/blob/future/platform/config/go-links.yaml).

fstanis:
This PR allows `go-links.yaml` to have this syntax for "catch-all" redirects:

```yaml
/c/amp-([a-z-]+):
  url: /documentation/components/amp-$1
  useRegex: true
```

It also adds redirects for components (e.g. `go.amp.dev/c/amp-form`) and samples (e.g. `go.amp.dev/s/amp-form`) and closes #2182 and #1860.

### DIFF
--- a/platform/config/go-links.yaml
+++ b/platform/config/go-links.yaml
@@ -24,9 +24,9 @@
 /websites: /about/websites
 
 # Regex-based entries
-/c/amp-([a-z-]+):
+^/c/amp-([a-z-]+)$:
   url: /documentation/components/amp-$1
   useRegex: true
-/s/amp-([a-z-]+):
+^/s/amp-([a-z-]+)$:
   url: /documentation/examples/components/amp-$1
   useRegex: true
--- a/platform/config/go-links.yaml
+++ b/platform/config/go-links.yaml
@@ -22,3 +22,11 @@
 /tools: /documentation/tools
 /vision-mission: /about/mission-and-vision
 /websites: /about/websites
+
+# Regex-based entries
+/c/amp-([a-z-]+):
+  url: /documentation/components/amp-$1
+  useRegex: true
+/s/amp-([a-z-]+):
+  url: /documentation/examples/components/amp-$1
+  useRegex: true
--- a/platform/lib/routers/go.test.js
+++ b/platform/lib/routers/go.test.js
@@ -1,0 +1,13 @@
+const express = require('express');
+const request = require('supertest');
+
+const app = express();
+const router = require('./go.js');
+app.use(router);
+
+test('redirects simple golink', (done) => {
+  request(app)
+      .get('/tools')
+      .expect(302)
+      .expect('Location', /\/documentation\/tools$/, done);
+});
--- a/platform/lib/routers/go.test.js
+++ b/platform/lib/routers/go.test.js
@@ -11,3 +11,16 @@ test('redirects simple golink', (done) => {
       .expect(302)
       .expect('Location', /\/documentation\/tools$/, done);
 });
+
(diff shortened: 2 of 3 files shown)


platform/config/go-links.yaml
```diff
@@ -22,3 +22,11 @@
 /tools: /documentation/tools
 /vision-mission: /about/mission-and-vision
 /websites: /about/websites
+
+# Regex-based entries
+/c/amp-([a-z-]+):
+  url: /documentation/components/amp-$1
+  useRegex: true
```
tharders: Seems to complicated to me.
Why don't we simply put the matches in two categories like the code uses them:
```yaml
simple-matches:
   /ads: /about/ads
regex-matches:
  /c/amp-([a-z-]+): /documentation/components/amp-$1
```

We could also simply use regex for all the matches.
The existing go links would even work without any escaping since they do not have special characters like the dot.
The downside is of course, that its easy to forget special escaping might be needed, since it is not obvious that the matches are regex when most of them do not contain regex specific matching.
So I would prefer the categories, but I am open for discussions.

platform/lib/routers/go.test.js
```diff
@@ -0,0 +1,26 @@
+const express = require('express');
+const request = require('supertest');
+
+const app = express();
+const router = require('./go.js');
+app.use(router);
+
+test('redirects simple golink', (done) => {
+  request(app)
+      .get('/tools')
```
tharders: Super that unit tests are included! :-D
But it would be better not to rely on the content in the regular config file, since this should be able to change without adapting the unit tests.
We already have an example where we mocked the yaml here: platform/lib/middleware/redirects.test.js

platform/config/go-links.yaml
fstanis: Some prior discussion in #2182 - I prefer the object notation since this allows us to add additional features / flags in the future.

I don't see it as more / less complicated than splitting into `simple-matches` and `regex-matches`, since users can easily be confused by what "simple" means. More importantly, I feel there's very limited need for regex matches, so I'm not too concerned about it being slightly more verbose.

> We could also simply use regex for all the matches.

I'm strongly against this for several reasons:

-   It should remain easy for people with no technical background to edit this file and add new links. Regex raises that bar significantly.

-   It's very error-prone: in fact, I just realized that the regex I used is incorrect (lacks `$` at the end).

-   What happens if two regexes match? This can easily happen by accident and it's very difficult to test.

-   Once again, I feel the need for regex matches in redirects is limited, definitely not something most redirects would need.

platform/lib/routers/go.test.js
fstanis: Done - thanks, I didn't know it was that easy to inject a mock config.

platform/config/go-links.yaml
tharders: Ok, you convinced me.
Additional flags is a strong point for your config structure.
And also the very obvious difference between direct matches and the few special regexp cases is not so bad.

Issue #1809, opened 2019-04-09
Fix validation and formatting errors in imported documents


matthiasrohmer:
Follow-up to #1802. Imported docs are sometimes badly formatted and break when imported to amp.dev. Good sample is the [`amp-selector` reference docs](https://amp.dev/documentation/components/amp-selector.html?format=websites). 

One problem is python-markdown not escaping the angle brackets in the term `<amp-selector>` and the dependency injector therefore wrongly injecting the `amp-sidebar` extension. 

Another one is that the documents have HTML with nested markdown syntax which is not parsed by python-markdown.

matthiasrohmer:
#1795 is somehow related and should be taken care of when extending `MarkdownDocument`.

westonruter:
I ran across what I believe to be an example of this today: https://amp.dev/documentation/components/amp-recaptcha-input.html?format=websites

![image](https://user-images.githubusercontent.com/134745/58989658-2f034400-8799-11e9-872a-6e913eecabd2.png)

matthiasrohmer:
@dlemm , #2367 will fix some validation errors but the formatting errors would still exist. It'd be great if you could take care of `amp-selector` for example and probably take Miri & Timo along with you to help you fix them and then also look at the other components. Especially they've been using Markdown in tables again which need to be fixed.

dlemm:
I'll take care of that.

dlemm:
What puzzled me is that the code is exactly the same for ```amp-accordion``` (for example) in the "Events" and "Actions" table. In the amphtml files there is
```<td width="40%"><strong><code>expand</code></strong></td>```
```<td width="40%"><strong><code>expand</code></strong></td>```
which render in the doc to 
```<td width="40%"><strong>`expand`</strong></td>```
and
```<td width="40%"><strong>`expand`</strong></td>``` 
but only the latter in the Actions table got rendered out right. The first one becomes ``` éxpand` ``

tharders:
It is a problem with Python-Markdown that sometimes the markdown inside an html table is evaluated and sometimes not. According to the documentation it should normally not be evaluated per default.
So I will enable the `markdown.extensions.extra` that will evaluate markdown in the table when the attribute `markdown="1"` is set for the table tag. I will also modify the page importer so that this attribute is set for all imported markdown documents. For documents inside the docs project we then have to add the attribute `markdown="1"` manually.

tharders:
Problems are fixed with PR #2633

PR #2367, opened 2019-06-24
Use manual dependency hints instead of automatic injection.

### LINKED ISSUE #1747: Use explicit hints for AMP extensions
The string-lookup based dependency injection is not 100% reliable and might create invalid AMP pages. Therefore this solution should be replaced by manual hints similar to `{% do doc.amp.use('amp-animation') %}`. Follow-up issue to #1670.

matthiasrohmer:
Fixes #1747 and fixes errors related to #1809.


frontend/templates/layouts/default.j2
```diff
@@ -1,4 +1,5 @@
 {% macro filter(formats='websites, ads, stories, email') -%}
+{% do doc.amp_dependencies.add('amp-anim', '0.1') %}
```
matthiasrohmer: Huh? Why would `amp-anim` be needed here?

frontend/templates/layouts/example-pages/amp-stories-page.j2
```diff
@@ -1,4 +1,5 @@
 {% set body_contains_story = example.body and example.has_tag(example.body, 'amp-story') %}
+{% do doc.amp_dependencies.add('amp-story', '0.1') %}
```
matthiasrohmer: Sorry, could have mentioned that: all layouts in `frontend/templates/layouts/example-pages` have another dependency mechanism - I guess. How do they get their deps actually, @tharders?

frontend/templates/views/detail/success-story-detail.j2
```diff
@@ -1,4 +1,6 @@
 {% extends '/layouts/default.j2' %}
+{% do doc.amp_dependencies.add('amp-vimeo', '0.1') %}
```
matthiasrohmer: Are always `amp-vimeo` and `amp-video` are used both? Otherwise they need to be inside the corresponding `{% if %}` clause.

frontend/templates/views/examples/documentation.j2
```diff
@@ -1,5 +1,7 @@
 {% extends '/layouts/default.j2' %}
 {% set inline_preview = not doc.example.document.metadata.hidePreview %}
+{% do doc.amp_dependencies.add('amp-iframe', '0.1') %}
```
matthiasrohmer: `amp-iframe` on the example documentation pages is only needed for story preview.

frontend/templates/views/partials/format-visuals/ads.j2
```diff
@@ -1,6 +1,7 @@
 {% do doc.styles.addCssFile('css/components/molecules/format-visual.css') %}
 {% do doc.styles.addCssFile('css/components/molecules/format-visual-ads.css') %}
 {% do doc.styles.addCssFile('css/components/molecules/code-snippet.css') %}
+{% do doc.amp_dependencies.add('amp-ad', '0.1') %}
```
matthiasrohmer: `amp-ad` is actually not used here.

frontend/templates/views/partials/teaser.j2
```diff
@@ -1,5 +1,6 @@
 {# Success story teaser #}
 {% if teaser_doc.view == '/views/detail/success-story-detail.j2' %}
+{% do doc.amp_dependencies.add('amp-video', '0.1') %}
```
matthiasrohmer: I guess the success story teaser does not always show a video.

frontend/templates/views/partials/video.j2
```diff
@@ -1,4 +1,6 @@
 {% do doc.styles.addCssFile('css/components/atoms/video.css') %}
+{% do doc.amp_dependencies.add('amp-video', '0.1') %}
+{% do doc.amp_dependencies.add('amp-youtube', '0.1') %}
```
matthiasrohmer: Only one of `amp-video` or `amp-youtube` is needed, put them in the corresponding if conditions.

pages/content/amp-dev/documentation/templates/index.html
```diff
@@ -41,6 +41,9 @@
       alt: image
 ---
 {% do doc.styles.addCssFile('css/components/templates/template.css') %}
+{% do doc.amp_dependencies.add('amp-bind', '0.1') %}
+{% do doc.amp_dependencies.add('amp-video', '0.1') %}
+{% do doc.amp_dependencies.add('amp-stories', '0.1') %}
```
matthiasrohmer: There is no component `amp-stories` and for what would it be used here?

frontend/templates/layouts/example-pages/amp-stories-page.j2
tharders: @matthiasrohmer @dlemm The imports are directly in these templates:
```html
{% if not example.has_tag_in_head('script', 'custom-element', 'amp-story') %}
  <script async custom-element="amp-story" src="https://cdn.ampproject.org/v0/amp-story-1.0.js"></script>
{% endif %}
{% for dependency in example.imports if dependency %}
{% if not example.has_import_in_head(dependency) and dependency.name != 'amp-story' %}
  <script custom-element="{{dependency.name}}" src="https://cdn.ampproject.org/v0/{{dependency.name}}-{{dependency.version}}.js" async></script>
{% endif %}
{% endfor %}
```
The example tags have the needed imports as an attribute. The version number is taken from https://playground.amp.dev/api/amp-component-versions

dlemm: Ok, I'll remove the added line.

matthiasrohmer:
@tharders, could you please have another look? Except the one comment I had it looks good from my side.

pages/extensions/amp_dependencies/__init__.py
```diff
@@ -0,0 +1,100 @@
...
+        self._pod = doc.pod
+        # Used to determine where to print all dependencies
+        self._placeholder = '{}'.format(uuid.uuid4())
+        # Stores registered paths
+        self._dependencies = []
+
+    def __repr__(self):
+        return '<AmpDependencies({})>'.format(self._placeholder)
+
+    def add(self, name, version='0.1'):
```
tharders: Would it be better to use the latest version as default instead of 0.1?
The old logic used the latest version from https://playground.amp.dev/api/amp-component-versions'
Or should we always require an explicit version.
The changes in the files all use an explicit version.

pages/extensions/amp_dependencies/__init__.py
```diff
@@ -0,0 +1,100 @@
...
+from grow.extensions import hooks
+
+class AmpDependencies(object):
+    def __init__(self, doc):
+        self._doc = doc
+        self._pod = doc.pod
+        # Used to determine where to print all dependencies
+        self._placeholder = '{}'.format(uuid.uuid4())
+        # Stores registered paths
+        self._dependencies = []
```
tharders: Should this not be a set, add may be called multiple times for the same dependency
Still the question would remain what to do if different versions are requested of the same dependency

pages/extensions/amp_dependencies/__init__.py
```diff
@@ -0,0 +1,100 @@
...
+
+        return previous_result if previous_result else raw_content
+
+
+class AmpDependenciesPostRenderHook(hooks.PostRenderHook):
+    """Handle the post-render hook."""
+
+    def should_trigger(self, previous_result, doc, original_body, *_args,
+                       **_kwargs):
+        """Only needs to trigger if pre-render hook added a stylesheet
```
tharders: 'stylesheet'?

pages/extensions/amp_dependencies/__init__.py
```diff
@@ -0,0 +1,100 @@
...
+        self._placeholder = '{}'.format(uuid.uuid4())
+        # Stores registered paths
+        self._dependencies = []
+
+    def __repr__(self):
+        return '<AmpDependencies({})>'.format(self._placeholder)
+
+    def add(self, name, version='0.1'):
+        src = 'https://cdn.ampproject.org/v0/{name}-{version}.js'.format(name=name, version=version)
+        type = 'element' if name != 'amp-mustache' else 'template'
```
tharders: When we are changing things to an explicit mode I think it would be better to explicitly define the type.
I would prefer to have different methods for templates and elements.
We already have that for [example] tags (different attributes 'imports' and 'template')
New template implementations would be supported out of the box then.

pages/extensions/amp_dependencies/__init__.py
```diff
@@ -0,0 +1,100 @@
+# -*- coding: utf-8 -*-
+import uuid
+from sets import Set
+
+from grow import extensions
+from grow.documents import document, static_document
+from grow.extensions import hooks
+
+class AmpDependencies(object):
+    def __init__(self, doc):
```
tharders: For new files please use the agreed indention width of 2

pages/extensions/amp_dependencies/__init__.py
```diff
@@ -0,0 +1,100 @@
...
+        self._doc = doc
+        self._pod = doc.pod
+        # Used to determine where to print all dependencies
+        self._placeholder = '{}'.format(uuid.uuid4())
```
tharders: There will be only one instance of AmpDependencies for a document.
So we do not need a dynamic placeholder

pages/extensions/amp_dependencies/__init__.py
matthiasrohmer: If no explicit version is set only then the extension falls back to 0.1 now.

pages/extensions/amp_dependencies/__init__.py
matthiasrohmer: It has been converted to a set later but is now one initially 🙂

pages/extensions/amp_dependencies/__init__.py
matthiasrohmer: As mentioned the whole extension is a copy-cat of the inline-text-assets extension 😉

pages/extensions/amp_dependencies/__init__.py
matthiasrohmer: Added the optional kwarg `type='element'`.

pages/extensions/amp_dependencies/__init__.py
matthiasrohmer: Done!

pages/extensions/amp_dependencies/__init__.py
matthiasrohmer: Correct! Changed it to a static placeholder.

matthiasrohmer:
@tharders, thanks for taking such a thorough look! Implemented all your comments.

frontend/templates/views/examples/documentation.j2
```diff
@@ -1,44 +1,43 @@
...
 {% if 'amp-consent' not in doc.used_components or not inline_preview %}
-<script async custom-element="amp-consent" src="https://cdn.ampproject.org/v0/amp-consent-0.1.js"></script>
+{% do doc.amp_dependencies.add('amp-consent', '0.1') %}
 {% if 'amp-geo' not in doc.used_components or not inline_preview %}
-<script async custom-element="amp-geo" src="https://cdn.ampproject.org/v0/amp-geo-0.1.js"></script>
+{% do doc.amp_dependencies.add('amp-geo', '0.1') %}
 {% endif %}
 {% endif %}
 {% if 'amp-analytics' not in doc.used_components or not inline_preview %}
 <script async custom-element="amp-analytics" src="https://cdn.ampproject.org/v0/amp-analytics-0.1.js"></script>
```
tharders: I think this should also be changed

frontend/templates/views/examples/documentation.j2
```diff
@@ -1,44 +1,43 @@
...
+{% do doc.amp_dependencies.add('amp-install-serviceworker', '0.1') %}
 {% endif %}
 
 {# Include amp-iframe for story previews #}
 {% if doc.example.document.isAmpStory %}
-<script async custom-element="amp-iframe" src="https://cdn.ampproject.org/v0/amp-iframe-0.1.js"></script>
+{% do doc.amp_dependencies.add('amp-iframe', '0.1') %}
 {% endif %}
 
 {# Include other AMP dependencies if needed for preview #}
```
tharders: And the loop below also has to change

matthiasrohmer:
Ah indeed #1809 goes further than I remembered - only validation errors coming from wrongly injected dependencies would be solved but not the formatting issues. Updated the PR description.

The CLA issue is fine - the Google Bot just can't determine if all PRs contributors are actually fine with their commits being part of this PR, but as it's just the three of us we can safely assume so 😉

matthiasrohmer:
Also looks good to me. Looking forward to an in-depth explanation of the updated RegEx 😁Thanks for taking care of this and enjoy your time off!

Will run a test build later and then merge.

PR #2381, opened 2019-06-26
Fix mustache escaping in markdown and fix unit test lookup

tharders:
The old logic to escape mustache expressions in markdown code blocks did not work well.
It could happen that the end of one code block and the start of an other where considered a code block, which caused problems by placing the {% raw %} tags wrong.
For example on the following page the link 'common attributes' is wrong:
https://amp.dev/documentation/examples/components/amp-img/#offline-fallback

This PR will fix this problem by searching for blocks and only in a second step escape the markup.
Also existing {% raw %} blocks are taken into account.
Additionally source blocks with [sourcecode] tags are supported.
Therefore the escaping logic in the `rewriteCodeBlocks` function could be removed.

Together with this the jest library and test lookup was fixed.

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -79,6 +79,7 @@ module.exports = {
   moduleNameMapper: {
     // TODO: improve: jest will not work with 'module-alias', so we have define the alias here again!
     // see https://github.com/ilearnio/module-alias/issues/46
+    '^@lib/utils$': '<rootDir>/platform/lib/utils/index.js',
     '^@lib/(.*?)(\.js)?$': '<rootDir>/platform/lib/$1.js',
   },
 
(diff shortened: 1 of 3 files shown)


jest.config.js
```diff
@@ -79,6 +79,7 @@ module.exports = {
   moduleNameMapper: {
     // TODO: improve: jest will not work with 'module-alias', so we have define the alias here again!
     // see https://github.com/ilearnio/module-alias/issues/46
+    '^@lib/utils$': '<rootDir>/platform/lib/utils/index.js',
```
matthiasrohmer: Just wondering: wouldn't it help to just add `require('module-alias/register');` at the top of this file? My guess would be that this file is the entry point for jest and all other files are only evaluated after.

tharders: @matthiasrohmer This does not work. I also tries to import the module-alias in a globalSetup module without success.

PR #2389, opened 2019-06-27
Add module for transforming samples from one format to another

fstanis:
This adds a `formatTransform` util that allows transforming from AMP for web to another format, specifically email.

The idea is to transform existing samples, e.g. [amp-form](https://amp.dev/documentation/examples/components/amp-form/?format=websites) "on the fly" to valid AMP for Email format. This is still work in progress, so the class is not currently used in the website.


platform/lib/format-transform/index.js
```diff
@@ -0,0 +1,92 @@
...
+  supportsFormat(target) {
+    return target in this.transforms;
+  }
+
+  getSupportedFormats() {
+    return Object.keys(this.transforms);
+  }
+
+  transform(input, target) {
+    if (!this.transforms[target]) {
```
sebastianbenz: why not use supportsFormat?

platform/lib/format-transform/transforms.js
```diff
@@ -0,0 +1,66 @@
...
+const host = config.getHost(config.hosts.platform);
+
+function absoluteUrl(url) {
+  if (!url.match(/^https?:\/\//)) {
+    url = host + url;
+  }
+  return url;
+}
+
+const amp4emailTransform = {
```
sebastianbenz: Please move this into it's own file.

platform/lib/format-transform/transforms.js
```diff
@@ -0,0 +1,66 @@
...
+  'html': (el) => {
+    el.removeAttr('amp');
+    el.removeAttr('⚡');
+    el.attr('⚡4email', '');
+  },
+  'form, a': (el) => {
+    el.removeAttr('target');
+  },
+  'amp-list': (el) => {
+    el.removeAttr('reset-on-refresh');
```
sebastianbenz: These transformations are dangerous as they break functionality. Instead, I'd throw an exception and mark the file as not transformable.

platform/lib/format-transform/transforms.js
```diff
@@ -0,0 +1,66 @@
...
+    el.attr('src', absoluteUrl(el.attr('src')));
+  },
+  'head > style[amp-boilerplate]': '<style amp4email-boilerplate>body{visibility:hidden}</style>',
+
+  'form[method=get i][action]': null,
+  'form[verify-xhr]': null,
+  'amp-state[src]': null,
+  'amp-img[srcset]': null,
+  '[\\[src\\]]': null,
+  'input[type=password]': null,
```
sebastianbenz: same as above, this is functionality breaking.

platform/lib/utils/cheerio.js
```diff
@@ -0,0 +1,41 @@
+/**
```
sebastianbenz: rename this file to cheerioHelper.js

platform/lib/format-transform/index.js
```diff
@@ -0,0 +1,92 @@
...
+    if (transform === null) {
+      this.removeElement_(el);
+      return;
+    }
+
+    if (typeof transform === 'function') {
+      transform(el);
+    } else if (typeof transform === 'string') {
+      el.replaceWith(transform);
+    }
```
sebastianbenz: add an else block throwing an exception

platform/lib/format-transform/index.js
```diff
@@ -0,0 +1,92 @@
...
+  }
+
+  transformElement_(el, transform) {
+    if (transform === null) {
+      this.removeElement_(el);
+      return;
+    }
+
+    if (typeof transform === 'function') {
+      transform(el);
```
sebastianbenz: nit:  early exit

platform/lib/format-transform/index.js
```diff
@@ -0,0 +1,92 @@
...
+    } else if (typeof transform === 'string') {
+      el.replaceWith(transform);
+    }
+  }
+
+  removeElement_(el) {
+    if (el.parent().is('div')) {
```
sebastianbenz: Why do we need to check for a parent div?

fstanis:
Addressed comments, most notably, removed the logic for element removal which was functionality breaking. We'll take a different approach here in another PR, to add conditional blocks to examples.

Also changed transformation format to include validatorRuntime that should be used to validate the code post-transform.

platform/lib/format-transform/index.js
```diff
@@ -0,0 +1,71 @@
...
+  getValidatorRuntime(format) {
+    if (!this.supportsFormat(format)) {
+      throw new Error(`Unsupported format: ${format}`);
+    }
+    return this.formats[format].validatorRuntime || null;
+  }
+
+  transform(input, target) {
+    if (!this.supportsFormat(target)) {
+      return input;
```
sebastianbenz: Throw an exception instead. Otherwise it's not possible for the caller to figure out if the transformation worked.

platform/lib/format-transform/index.js
```diff
@@ -0,0 +1,71 @@
...
+      const elements = $(selector);
+      elements.each((i, el) => {
+        this.transformElement_($(el), transforms[selector]);
+      });
+    }
+    return htmlContent($);
+  }
+
+  transformElement_(el, transform) {
+    if (transform === null) {
```
sebastianbenz: add a comment explaining that this is the delete element case

platform/lib/format-transform/formats/amp4email.js
```diff
@@ -0,0 +1,54 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+const config = require('@lib/config.js');
+const host = config.getHost(config.hosts.platform);
+
+function absoluteUrl(url) {
+  if (!url.match(/^https?:\/\//)) {
```
sebastianbenz: Use `new URL(url, host).toString()` instead

platform/lib/format-transform/formats/amp4email.js
```diff
@@ -0,0 +1,54 @@
...
+  transforms: {
+    'html': (el) => {
+      el.removeAttr('amp');
+      el.removeAttr('⚡');
+      el.attr('⚡4email', '');
+    },
+    'form, a': (el) => {
+      el.removeAttr('target');
+    },
+    'amp-list': (el) => {
```
sebastianbenz: this is also a functionality breaking transformation I would not support.

platform/lib/format-transform/formats/amp4email.js
fstanis: Well TIL. I moved this method to config, let me know if that doesn't make sense.

fstanis:
@sebastianbenz can we merge this?

PR #2394, opened 2019-06-28
Add pulsing animation to upcoming cities on roadshow map

sebil:
With this PR we can finally complete the roadshow map changes.
It adds an animation for the upcoming cities and also moves Berlin and Atlanta to the past cities list.

@morsssss I've added placeholder content to the nextstop list and commented it out, so it is easier to fill out the needed values.

### DIFF
--- a/frontend/scss/components/organisms/cities.scss
+++ b/frontend/scss/components/organisms/cities.scss
@@ -13,6 +13,7 @@
 
 @import 'components/atoms/_headline.scss';
 @import 'components/atoms/_text.scss';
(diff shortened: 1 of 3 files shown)


morsssss:
Sadly I can't build the site today, I think because of #2410 .

For that matter, I'm ready to submit my own PR in which I remove old cities and add new ones using your very helpful coordinates technique! But I made some more changes to the roadshows page and don't want to submit that until I can test... but... #2410 ...

cc @fstanis

PR #2404, opened 2019-06-29
Add Japanese localization on "Add Translations" page

hogashi:
issue: https://github.com/ampproject/docs/issues/2378


hogashi:
will retry in another p-r

PR #2407, opened 2019-07-01
Add trailing slash for example redirect

fstanis:
Temporary fix until #2180 is resolved.

Also changed from `go.amp.dev/s/...` to `go.amp.dev/e/...` since "**e**xample" makes more sense than "**s**ample" based on how we call them on the website. Open to discussion about that though.


matthiasrohmer:
Makes sense, thanks!

PR #2409, opened 2019-07-01
Add Mail.ru playground to tools

fstanis:


fstanis:
CC @8coon and @latrekc

PR #2411, opened 2019-07-01
Add example transforms for email for supported components

fstanis:
This PR requires both #2389 and ampproject/amp-by-example#2028 to be merged first.

### DIFF
--- a/examples/source/1.components/amp-accordion.html
+++ b/examples/source/1.components/amp-accordion.html
@@ -3,6 +3,8 @@
 standaloneSnippets: true
 preview: true
 author: sebastianbenz
+transforms:
+  - email
 
 --->
 
--- a/examples/source/1.components/amp-accordion.html
+++ b/examples/source/1.components/amp-accordion.html
@@ -3,8 +3,6 @@
 standaloneSnippets: true
 preview: true
 author: sebastianbenz
-transforms:
-  - email
 
 --->
 
--- a/platform/lib/build/samplesBuilder.js
+++ b/platform/lib/build/samplesBuilder.js
@@ -253,8 +253,8 @@ class SamplesBuilder {
     }, sample.contents.toString());
 
     // Transformed sample files end with ".<format>", e.g. "amp-list.email".
-    // Change this into " (format)" instead, e.g. "amp-list (email)".
-    parsedSample.document.title = parsedSample.document.title.replace(/\.([^.]+)$/, ' ($1)');
+    // This removes that suffix to have the same title as the original.
+    parsedSample.document.title = parsedSample.document.title.replace(/\.[a-z]+$/, '');
 
     // parsedSample.filePath is absolute but needs to be relative in order
     // to use it to build a URL to GitHub
--- a/platform/lib/build/samplesBuilder.js
+++ b/platform/lib/build/samplesBuilder.js
@@ -253,8 +253,11 @@ class SamplesBuilder {
     }, sample.contents.toString());
 
     // Transformed sample files end with ".<format>", e.g. "amp-list.email".
-    // This removes that suffix to have the same title as the original.
-    parsedSample.document.title = parsedSample.document.title.replace(/\.[a-z]+$/, '');
+    if (parsedSample.document.title.match(/\.[a-z]+$/)) {
+      parsedSample.isTransformed = true;
+      // Strip the ".<format>" suffix to have the same title as the original.
+      parsedSample.document.title = parsedSample.document.title.replace(/\.[a-z]+$/, '');
+    }
 
     // parsedSample.filePath is absolute but needs to be relative in order
     // to use it to build a URL to GitHub
@@ -516,6 +519,11 @@ class SamplesBuilder {
     const data = manual.clone();
     data.contents = Buffer.from(JSON.stringify(parsedSample));
     data.extname = '.json';
+
+    // TODO: remove this when we add support for transformed samples in docs.
+    if (parsedSample.isTransformed) {
+      return [data];
+    }
 
     return [manual, data];
   }
--- a/platform/lib/build/samplesBuilder.js
+++ b/platform/lib/build/samplesBuilder.js
@@ -147,6 +147,7 @@ class SamplesBuilder {
 
           const isWebSample = String(document.formats()) === 'websites';
           const shouldTransform = isWebSample
+              && !document.metadata.disableTransform
               && !document.metadata.hideCode
               && !document.metadata.disablePlayground
               && !document.metadata.hidePreview;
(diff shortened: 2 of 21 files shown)


examples/source/1.components/amp-accordion.html
```diff
@@ -3,6 +3,8 @@
 standaloneSnippets: true
 preview: true
 author: sebastianbenz
+transforms:
+  - email
```
sebastianbenz: Can't we determine this automatically by transforming all samples and then running the validator?

platform/lib/build/samplesBuilder.js
```diff
@@ -160,7 +171,7 @@ class SamplesBuilder {
           for (const file of files) {
             stream.push(file);
           }
-
+        })).then(() => {
```
sebastianbenz: We shouldn't combine then and await. While you're fixing this, can you also change  `.catch(...)` into a `catch` block into a

fstanis:
As we discussed, we'll need to add format filtering to examples to make this work without functionality-breaking transforms from #2389.

Proposal is to allow examples to include `@format(format1,format2)` anywhere in the block comment. This comment signifies that the section that follows is only relevant for those formats and will be stripped prior to running formatTransform.

matthiasrohmer:
Sorry if I am missing context, just wanted to add this: wouldn't filtering the comments/documentation work with the already implemented syntax like `[filter formats="email"]This only shows for email[/format]`?

sebastianbenz:
Good idea. That could work. The closing tag might be a problem in some cases as you need to add a dummy empty comment before subsequent html elements without comment:
 
```
<!--
[filter formats="email"] 

This is amp-form ...
-->
<form class="sample-form"
  method="GET"
  action="/documentation/examples/api/submit-form"
  target="_top">
  ...
</form>
<!-- [/format] -->
<!-- -->
<div class="this-should-not-be-displayed-as-a-snippet"></div>
```

fstanis:
Sorry, I fear that adding logic like that (specifically, implementation that includes both start and end tag) would take significant time (with my knowledge of the samples parser, based on experience doing ampproject/amp-by-example#1338) and I have limited bandwidth for this.

It's significantly easier to implement IFF:

-   It's section level, i.e. no ending tag.
-   It can be caught with a (simple) regex.

sebastianbenz:
@fstanis please give Matthias's suggestion a try. It might already work as sample pages are also generated via grow (which understands the filter syntax).

fstanis:
Sorry @mathiasbynens, misunderstood your comment!

Anyway, this works, but only if both start and end are in the same comment. Your code sample results in this error which I'm not sure how to debug:

> Unexpected end of template. Jinja was looking for the following tags: 'endcall'. The innermost block that needs to be closed is 'call'.

(there are a few other issue, such as the fact that the current system assumes an example always has one format, but that could probably be fixed)

More importantly, this wouldn't work for the `preview.amp.dev` endpoint, which is my #1 priority with this feature. For that to work, we need the different formats to all be generated at build time rather than runtime. For this reason, I prefer the aforementioned `@format` block-level syntax.

sebastianbenz:
Good point. This won't work for the previews.

platform/lib/build/samplesBuilder.js
fstanis: Done in #2464.

mathiasbynens:
> Sorry @mathiasbynens, misunderstood your comment!

You meant @matthiasrohmer, not me. (GitHub’s autocomplete widget could use some love…)

sebastianbenz:
@fstanis what's the status here?

fstanis:
This should now be done. PTAL.

I added `@formats` logic, removed metadata (now it will attempt to transform every component sample) and validation after transform.

platform/lib/build/samplesBuilder.js
```diff
@@ -140,33 +143,45 @@ class SamplesBuilder {
...
-            ...this._createPreview(sample, parsedSample),
-            ...this._renderEmbed(sample, parsedSample),
-          ];
-
-          // Since stream.push doesn't allow to push multiple files at once
-          /* eslint-disable guard-for-in */
-          for (const file of files) {
-            stream.push(file);
+          const samples = [];
+          if (sample.relative.indexOf('1.components/') === -1) {
```
sebastianbenz: why not all samples?

platform/lib/build/samplesBuilder.js
```diff
@@ -228,6 +243,8 @@ class SamplesBuilder {
       },
     }, sample.contents.toString());
 
+    parsedSample.document.title = parsedSample.document.title.replace(/\.([^.]+)$/, ' ($1)');
```
sebastianbenz: what does this do? please add a comment

platform/lib/format-transform/index.js
```diff
@@ -67,6 +92,57 @@ class FormatTransform {
...
+  findCommentFormatFilters_(text) {
+    const match = text.match(FORMATS_REGEXP);
+    if (!match) {
+      return null;
+    }
+    return new Set(match[1].split(',').map(e => e.trim()));
+  }
+}
+
+async function getInstance() {
```
sebastianbenz: rename to `create`, `getInstance` implies a shared singleton.

platform/lib/build/samplesBuilder.js
fstanis: It pulls in some unexpected things, e.g. https://amp.dev/documentation/examples/guides/using_the_google_amp_cache/ is somehow an example.

In theory, this is temporary, but fixing it needs to either (1) make sure only examples are examples or (2) make an opt-out mechanism.

I'll add a comment to explain it.

sebastianbenz: You can check if any of these is true:

```
hideCode: true
disablePlayground: true
hidePreview: true
```

That should filter out the weird ones...

fstanis:
Addressed comments, updated tests for `formatTransform` (do we have tests for gulp too?) and fixed linter issues.

fstanis:
I changed the way validation results are returned and added some tests for it.

sebastianbenz:
Tried the PR locally. Two things:

* can you remove the `(email)` postfix. It's confusing as you're already in "email" mode
* we should fix the URLs as part of this PR

sebastianbenz:
@matthiasrohmer have you got a good idea, how we load `sample.$FORMAT.html` instead of `sample.html` if `/sample/?format=$FORMAT` is requested? If I remember correctly you implemented something similar for the normal doc pages?

fstanis:
> can you remove the `(email)` postfix. It's confusing as you're already in "email" mode

Done.

> we should fix the URLs as part of this PR

I added logic for the following:

1.  If the URL has a `format` parameter and the format is not `websites` (e.g. `?format=email`), then try to load template with `.${format}` first.
2.  If that fails, continue with normal flow.
3.  If the URL ends with `.<something>`, redirect to `?format=<something>` (`amp-img.email/` -> `amp-img/?format=email`).

fstanis:
As we discussed, I temporarily made it so generated samples are hidden from the main website and only visible in the playground and preview until we resolve the issues with URLs.

PR #2413, opened 2019-07-01
Feature/new roadshows

morsssss:
New roadshow cities, plus new text... and a little new styling for that new text.

Also made it possible to have new cities that don't have an event link.

Please feel free to tweak or totally change my styling, of course!

I also noticed that at screen widths from ~680px to ~768px, you see three new cities on one line, and one city taking up the entirety of another line. Ideally they'd stay in a 2x2 grid. But I didn't fix this.


morsssss:
Thanks, Matthias!

It's not urgent. Mostly I'm eager to get the links for the do-it-yourself
roadshow and roadshow request forms live soon, so if it doesn't look bad to
you, I'd just merge it tomorrow and wait for a fast follow from @sebil. If
you think it could use some serious work, let's wait!

pages/shared/data/roadshow.yaml
```diff
@@ -1,12 +1,36 @@
...
-  #   coords: 9x9
-  #   location: Menlo Park
-  #   time: 09:00–18:00
-  #   day: 4
-  #   month: sep
-  #   year: 1998
-  #   link: https://google.com/
+nextstop:
+  - title: tel aviv
+    coords: 64x21
```
sebil: I'd suggest changing Tel Aviv coords to 66x22 and Istanbul coords to 64x19 to fix the overlapping issue. This would fix the overlapping issue.
fix-coords

pages/shared/data/roadshow.yaml
```diff
@@ -1,12 +1,36 @@
...
+    location: tba
+    time: 09:00–18:00
+    day: 22
+    month: sep
+    year: 2019
+    link: #
+  - title: istanbul
+    coords: 64x19
```
sebil: See comment above

sebil:
@morsssss I found only 2 little things:
First, the coordinates of Tel Aviv and Istanbul should be adjusted a little bit to prevent them from overlapping. See my suggestion and screenshot above.

Second, a little alignment bug on the cards, see screenshot here:
box-not-aligned

I also could fix that second issue after merging. _I'm not sure how Github handles it, when I make changes to someone else's PR. The last time I did, we couldn't merge anymore..._ 🤦‍♂

The rest is fine for me 👍

morsssss:
@sebil - sure, I changed these coordinates in the way you suggest.

Also, I just switched our Cairo event to Karachi, so I made that change. I then moved Mumbai a bit so it wasn't so close to Karachi.

Generally I admit I miss the map we had on the old site, which was more detailed and where I could find more realistic spots for events!

Also, yes, the cards aren't quite the right size (see the comment I made above too). Thanks for offering to fix all of those!

After this goes through, I'm going to add even more new cities. We'll face a new issue, though, because some of those are cities we've been to before. Without some fancy footwork, we'll have some overlaps!

PR #2414, opened 2019-07-02
Load component versions locally instead of using ABE

### LINKED ISSUE #2410: Cache component versions locally instead of making an HTTP request
It seems that we're getting component versions from a remote endpoint:

https://github.com/ampproject/docs/blob/5fe8a1c60d3e1230022125b43b4e774be05c6ee4/pages/extensions/amp_example_preview/util/amp_component_versions.py#L3-L5

Since `https://playground.amp.dev/api/amp-component-versions` doesn't seem to work since ampbyexample.com doesn't work, this prevents `gulp develop` from working now.

Since component versions don't change often enough to justify making a request like this, we should consider caching component versions locally and statically generating them. Or at the very least hosting them outside of `ampbyexample.com`.

fstanis:
Fixes #2410

`pages/extensions/amp-component-versions.json` was just requested by using:

```bash
curl -k -X GET -H "x-requested-by: playground" "https://ampbyexample.com/playground/amp-component-versions"
```

This is temporary, a better way to generate is coming when AMP validator JSON gets hosted (related to ampproject/amphtml#22528).


fstanis:
I'd hold off on that - I think we can replace `amp-component-versions` with just a static file generated at build time.

This should be significantly easier to do when we get validator data as JSON (expecting this to be live soon).

PR #2434, opened 2019-07-05
Allow markdown in all table tags of imported documents

tharders:
This addresses problems discussed in https://github.com/ampproject/docs/issues/1809 that sometimes markdown is not evaluated in table tags.

### DIFF
--- a/platform/lib/pipeline/markdownDocument.js
+++ b/platform/lib/pipeline/markdownDocument.js
@@ -28,15 +28,20 @@ const LOG = new Signale({'scope': 'Markdown Documents'});
 // This expression matches a {% raw %}...{% endraw %} block
 const JINJA2_RAW_BLOCK = /\{%\s*raw\s*%\}(?:(?!\{%\s*endraw\s*%\})[\s\S])*\{%\s*endraw\s*%\}/;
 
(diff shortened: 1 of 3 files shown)


platform/lib/pipeline/markdownDocument.js
```diff
@@ -157,13 +171,21 @@ class MarkdownDocument {
...
+   * let the markdown.extensions.extra of grow evaluate markdown inside the table.
+   * Needed because often code blocks are inside the table.
+   * @param  {String} contents
+   * @return {String} The rewritten input
+   */
+  static enableMarkdownInHtmlTables(contents) {
+    return contents.replace(TABLE_MARKDOWN_ATTRIB_PATTERN, (match, p1, p2) => {
+      // only if we are not in a code block set the markdown attribute
+      if (p1) {
+        return p1 + ' markdown="1"' + p2;
```
sebastianbenz: Is `markdown="1"` going to be removed by the markdown extension? Otherwise, this would lead to invalid AMPHTML.

matthiasrohmer: Jep, the attribute is removed by python-markdown.

PR #2459, opened 2019-07-10
Migrate FAQ overview and detail pages from amproject.org with new design

sebil:
With this PR we finally get a cool FAQ section 😎

Some things changing:
- Menu Item "Support" gets a flyout menu, to easily reach the FAQ overview page
  - To beware the ability of reaching the support page itself, the first item in the flyout ist now "Help"
  - So it's `Support > Help` and `Support > FAQ`
- The page `/community/platform-and-vendor-partners` moved over to the FAQ's
- There is now one `faq.yaml` for all contents (for better maintenance for non-devs)
- All FAQ pages are markdown files (also for better maintenance for non-devs)
- For the typical accordion you can now simply include the accordion.j2 and pass over the array it should loop trough (see examples in the markdown files)
  - Optionally you can define a list of links in the faq.yaml
- All the contents are migrated from https://www.ampproject.org/support/faqs but with a new design

If something is not clear, feel free to ask.

Here come some screenshots:
amp_faq_flyout
amp_faq_index
amp_faq_overview
amp_faq_platform-and-vendor-partners
amp_faq_supported-browsers


frontend/scss/components/templates/faq-detail.scss
```diff
@@ -0,0 +1,54 @@
...
+@import 'components/templates/_default';
+
+.#{template('faq-detail')} {
+  top: 0;
+  left: 0;
+  background-repeat: no-repeat;
+  background-size: 100% auto;
+  background-position: top right;
+  background-attachment: fixed;
+  background-image: url('data:image/svg+xml;utf8,<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 1024 768"><defs><linearGradient id="a" x1="239.67" y1="855.23" x2="606.28" y2="-152.04" gradientUnits="userSpaceOnUse"><stop offset="0" stop-color="%23fff" stop-opacity="0"/><stop offset="1" stop-color="%23EDEDF1"/></linearGradient></defs><g><path fill="url(%23a)" d="M0 0h1024v214L786.37 768H0V0z"/></g></svg>');
```
matthiasrohmer: Why aren't we using a CSS gradient here?

frontend/templates/views/overview/faq-overview.j2
```diff
@@ -0,0 +1,13 @@
+{% extends '/layouts/default.j2' %}
```
matthiasrohmer: As the FAQ overview will only every exist once can we use a similar pattern to `pages/content/amp-dev/documentation/components/index.html` here (using `$view: /views/custom.j2`)?

matthiasrohmer:
... oh, and getting rid of the background for the browser logos or tinting them in the new grey would be nice ;)

frontend/scss/components/templates/faq-detail.scss
sebil: That's the background bevel we also use on the documentation pages

frontend/templates/views/overview/faq-overview.j2
sebil: Good point! I'll change it

sebil:
> ... oh, and getting rid of the background for the browser logos or tinting them in the new grey would be nice 😉

🦅👁!

sebil:
All done

sebil:
squooshed it to the limit

matthiasrohmer:
Finally the broken references check pays off 😅 @sebil, can you please check and fix?

sebil:
shame on me! 😖

PR #2464, opened 2019-07-11
Replace promise handling with `await` in samplesBuilder.js

fstanis:
-   Uses content override in abe added in ampproject/amp-by-example#2028 to speed up build
-   Replaces all calls to `then` and `catch` with `await`
-   Minor JSDoc fixes

### DIFF
--- a/platform/lib/build/samplesBuilder.js
+++ b/platform/lib/build/samplesBuilder.js
@@ -326,7 +326,7 @@ class SamplesBuilder {
         flag: 'wx+',
       });
       this._log.success('Wrote sample sitemap.');
-    } catch () {
+    } catch (_) {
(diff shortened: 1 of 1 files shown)


platform/lib/build/samplesBuilder.js
```diff
@@ -81,6 +81,12 @@ class SamplesBuilder {
     this._sitemap = {};
   }
 
+  /**
+   * Parse a sample source file into a JSON using the parser from the
+   * ampbyexample.com package and while doing so updates some fields
```
sebastianbenz: > updates some fields

This method does things ;-)

fstanis: Whoops 😁

PR #2468, opened 2019-07-11
Feature/teachers pixel

### LINKED ISSUE #2386: Add tracking pixel to "Get the courses" button
We'd like to know how many people are clicking on the "Get the courses" button at https://amp.dev/documentation/courses/teachers.

To do this, I'm planning to create a new event in GA. I'd then like to give that "Get the courses" button an ID and track it with a trigger as described [here](https://developers.google.com/analytics/devguides/collection/amp-analytics/).

I'm looking at the `analytics.j2` partial. I guess I'd need to add something to the podspec to let the config include triggers. At this point, I thought I should ask others for guidance!

/cc @matthiasrohmer @sebastianbenz @pbakaus @CrystalOnScript

morsssss:
Resolves #2386 .

I decided to use this method to check which page is in use... let me know if another method is preferred.

### DIFF
--- a/frontend/templates/views/partials/analytics.j2
+++ b/frontend/templates/views/partials/analytics.j2
@@ -9,7 +9,7 @@
         }
       }
     }
-{% if '/courses/' in doc.pod_path and doc.title == 'Teachers' %}
+{% if doc.pod_path == '/content/amp-dev/documentation/courses/teachers.md' %}
     ,
       "triggers": {
         "button": {
--- a/frontend/templates/views/partials/analytics.j2
+++ b/frontend/templates/views/partials/analytics.j2
@@ -16,9 +16,10 @@
           "selector": "#get-courses",
           "on": "click",
           "vars": {
-            "category": "courses",
-            "action": "download"
-          }
+            "event_name": "courses",
+            "event_category": "courses",
+            "event_action": "download"
+            }
         }
       }
 {% endif %}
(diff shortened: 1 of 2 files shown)


frontend/templates/views/partials/analytics.j2
```diff
@@ -9,6 +9,20 @@
         }
       }
     }
+{% if '/courses/' in doc.pod_path and doc.title == 'Teachers' %}
```
sebastianbenz: Will this work for translated versions? Why not only use the path?

matthiasrohmer: Sebastian was a little bit quicker 😁 Check for `doc.pod_path == '/content/amp-dev/documentation/courses/teachers.md'` instead.

frontend/templates/views/partials/analytics.j2
```diff
@@ -9,6 +9,20 @@
...
+    ,
```
matthiasrohmer: Does not needs to be done in this PR: we should refactor the analytics configuration to be a Python dict in the template which is then printed with `config|tojson` - that's a little less verbose.

morsssss:
Yeah, ok, well, that's the other way I was considering. I see now that it's better :)

To match your phenomenally fast reviews, I've rapidly changed this, checked it, and resubmitted.

PR #2470, opened 2019-07-11
update cache doc for amp for email

alabiaga:
snippet on amp4email docs being exempt from an amp cache


alabiaga:
@CrystalOnScript I wasn't sure on the proper way to communicate this. Pls advise and I can make the change.

alabiaga:
@CrystalOnScript Thanks. Yea i think it's worth noting so I will make the change and remove it from formats as well.

PR #2472, opened 2019-07-11
Adding Sydney and Melbourne to roadshow cities

morsssss:
I moved Sydney over a bit to accommodate Melbourne.

Note that this is the first case where we have a city twice... namely, Sydney. I hacked this by moving the old city Sydney up by a coordinate so that it would fit entirely beneath the new city Sydney. But we're going to have this problem a couple more times soon as we repeat two other cities.


sebastianbenz:
Happy pixel pushing! @matthiasrohmer can we handle this cases in the template? e.g by checking if there's an upcoming one, before placing a marker for an old one?

PR #2474, opened 2019-07-12
🍱 Add unpacked template zip-folders

### LINKED ISSUE #2466: Unpack Template ZIPs
Our infrastructure supports virtual ZIPs for a while now meaning a directory in one of the static directories with its name ending on `.zip` will be packed in to an archive during build time.

This allows the contents of the ZIP files to be version controlled. Only thing left to do is unpacking the existing template ZIPs.

dlemm:
Fixes #2466 

The template zip files are unpacked and saved as .zip-folders


sebastianbenz:
Can we please optimize the images while we're on it. 

@alankent FYI this makes it possible to:

1. add a link to browse on GitHub 
2. serve a demo version on amp.dev (we only need to add a static router)

dlemm:
@sebastianbenz Images should now be optimized

Issue #2374, opened 2019-06-25
404 page on amp documentation


Amanrazdan:
Got this 404 on  https://amp.dev/documentation/courses/?format=email 
![image](https://user-images.githubusercontent.com/28403318/60080783-31304100-974e-11e9-9020-255bb681efc3.png)

matthiasrohmer:
Thanks! That's indeed a bug. The course pages aren't format filterable although the sidebar shows the format toggle.

PR #2480, opened 2019-07-12
🐛 Do not show format toggle for courses

### LINKED ISSUE #2477: Non-website course return a 404
For example: https://amp-dev-staging.appspot.com/documentation/courses/?format=ads

matthiasrohmer:
Fixes #2477 and #2374.


morsssss:
Thanks for doing this!

I don't have your level of experience with grow.io, but is there an easy way to make this come from the podspec? So it could be configurable for any section of the site?

sebastianbenz:
Thanks Matthias!

morsssss:
Just realized - I still see the toggle [on the site](https://amp.dev/documentation/courses/).

Am I missing something?

Issue #2475, opened 2019-07-12
Make various pages format- or category filterable server- and client-side


matthiasrohmer:
Successor to #1694 and #2341. #2325 will allow SSR for pages based on one of the context variables `format = 'websites' || 'stories' || 'ads' || 'email'` and `category`.

We should use this to make the following pages filterable per either deep-link or via the client side:

- Success story overview
- Templates overview
- Tools
- Roadmap

dlemm:
The template overview, the Roadmap and the success story overview page are still open, right?

sebastianbenz:
Does Tools already support SSR?

>

fstanis:
Friendly ping - is this still on the roadmap? It would really help with the tools page.

matthiasrohmer:
I need to double check: AFAIK the tools page was one of the first pages we've implemented SSR on but needed to revert multiple times due to errors with our minifying pipeline. 

A quick check on https://amp.dev/documentation/tools/?format=stories does nothing so I guess we never merged it again after fixing the related bugs. See #2509 and #2508 and finally #2614 for more history.

If I didn't miss anything in this quick research we are good to reopen a PR to finally add SSR for the tools page. Given the branch with this work still exists somewhere.

fstanis:
Another friendly ping on this.

sebastianbenz:
It's live on staging for success-stories and tools:
https://amp-dev-staging.appspot.com/documentation/tools/?format=email

>

PR #2492, opened 2019-07-15
✨ Allow SSR for tools page

robinvanopstal:
Part of #2475


pages/content/amp-dev/documentation/tools.html
```diff
@@ -61,7 +61,7 @@ <h2 class="ap-o-stage-content-subline">for creation, design and development</h2>
 <amp-state id="activeFilter">
```
matthiasrohmer: That's a nit but can we rename "activeFilter" to just "filter" and "chosenFilter" to "format"?

robinvanopstal: okaay, done :)

PR #2494, opened 2019-07-15
Add filtering for AMP examples

fstanis:
-   Allows examples to use `filters:` frontmatter
-   Allows examples to filter sections using `@format` in the comments ([context](https://github.com/ampproject/docs/pull/2411#issuecomment-509748791))
-   Updates docs for frontmatter and formats

### DIFF
--- a/platform/lib/build/samplesBuilder.js
+++ b/platform/lib/build/samplesBuilder.js
@@ -242,12 +242,6 @@ class SamplesBuilder {
       markdown = MarkdownDocument.rewriteCodeBlocks(markdown);
       markdown = MarkdownDocument.escapeMustacheTags(markdown);
 
-      const filterMatch = markdown.match(/@filter\(([^)]+)\)/);
-      if (filterMatch) {
-        section.filters = filterMatch[1].split(',');
-        markdown = markdown.replace(filterMatch[0], '');
-      }
-
       // Splice out sourcecode blocks to preserve whitespace
       const codeBlocks = {};
       const CODE_BLOCK_PATTERN = /\[sourcecode.*?\[\/sourcecode]/gms;
--- a/platform/lib/build/samplesBuilder.js
+++ b/platform/lib/build/samplesBuilder.js
@@ -236,11 +236,17 @@ class SamplesBuilder {
     parsedSample.route = this._getDocumentationRoute(sample);
 
     // Rewrite some markdown to be consumable by Grow
-    for (const index in parsedSample.document.sections) {
+    for (const section of parsedSample.document.sections) {
       // Replace GitHub sourcecode syntax by python-markdown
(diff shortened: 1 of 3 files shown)


platform/lib/build/samplesBuilder.js
```diff
@@ -492,21 +499,26 @@ class SamplesBuilder {
   /**
    * Used to determine the sample format by string
    * @param  {Object} parsedSample
-   * @return {string}
+   * @return {Array<string>}
    */
-  _getSampleFormat(parsedSample) {
+  _getSampleFormats(parsedSample) {
+    if (parsedSample.document.metadata.formats) {
+      console.log(parsedSample.document.metadata.formats);
```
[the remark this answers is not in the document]
fstanis: Whoops, fixed

fstanis:
Moved `@filter` logic to ampproject/amp-by-example#2037

sebastianbenz:
Thanks - changes are live @0.2.15

platform/lib/build/samplesBuilder.js
```diff
@@ -657,7 +662,7 @@ class SamplesBuilder {
         '$localization': {
           'path': `/{locale}${this._getPreviewRoute(sample)}`,
         },
-        'formats': [this._getSampleFormat(parsedSample)],
+        'formats': this._getSampleFormats(parsedSample),
```
sebastianbenz: This is now redundant with filters. Can we somehow merge this with filters? //cc @matthiasrohmer

fstanis: Not sure I follow? Filters (`@filter`) are section-level and formats are article-level. This makes it aligned (more or less) with non-example articles which have both.

While we could possibly guess the formats in a smarter way, I'd prefer the less-eager model where the example author is explicit in which format the example is useful for. E.g. just because an example *works* in emails doesn't make it a good example of something to use in an email.

sebastianbenz: You're right, forgot that filters are on section level. We could even go as far as validating section filter against formats defined in the front matter.

fstanis: I think that makes sense.

Looking at the code, seems we have two options:

1.   Move the call to `_getSampleFormats` into `_parseSample`, e.g. `parsedSample.formats = this._getSampleFormats(parsedSample)` and then verify when traversing the sections in `_parseSample`.

2.   Move the whole `_getSampleFormats` login into ampproject/amp-by-example.

Any preference?

sebastianbenz: I think 2. would be good, this avoids also that the method is being called twice in this file.

sebastianbenz: @fstanis good with merging this branch?

PR #2514, opened 2019-07-16
update dependencies

sebastianbenz:
Reduces the number of vulnerabilities from 15938 to 1

### DIFF
--- a/package.json
+++ b/package.json
@@ -77,7 +77,7 @@
     "@babel/runtime": "7.5.4",
     "@octokit/rest": "16.28.4",
     "absolutify": "0.1.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -77,7 +77,7 @@
     "@babel/runtime": "7.5.4",
     "@octokit/rest": "16.28.4",
     "absolutify": "0.1.0",
-    "amp-by-example": "0.2.13",
+    "amp-by-example": "^0.2.15",
```
matthiasrohmer: Is this a non-fixed version intenionally?

sebastianbenz: nope. thanks for spotting. updated the version as well.

PR #2516, opened 2019-07-16
Fix roadshow maps duplicate locations and event box styling

sebil:
This fixes the duplicate city issue for the roadshow map.
A past-city will only be added to the map if it doesn't exist in the upcoming-city list.
Also this PR fixes the latest styling issue of the event boxes.

roadshow-map-events

### DIFF
--- a/frontend/scss/components/molecules/event.scss
+++ b/frontend/scss/components/molecules/event.scss
@@ -26,20 +26,22 @@ none
 @import 'components/atoms/_icon.scss';
 
 
-
 .#{molecule('event')} {
+
   @include roadshow-event-shadow;
+  flex: 0 0 calc(100% - 15px);
   display: flex;
   background-color: color('white');
   margin: 10px 0;
-  padding: 20px 30px;
-  min-width: 100%;
+  padding: 20px 15px 20px 30px;
 
   @media (min-width: 768px) {
-    max-width: 350px;
-    min-width: 350px;
-    margin: 20px;
-    margin-left: 0;
+    flex: 0 0 calc(50% - 10px);
+    margin: 20px 0;
+  }
+
+  @media (min-width: 1024px) {
+    flex: 0 0 calc(33.33333333% - 20px);
   }
 
 	&-date {
@@ -91,6 +93,7 @@ none
       @include txt-hl;
       @include color-blue-ribbon;
 
+      margin-top: 0;
       margin-bottom: 0.5em;
     }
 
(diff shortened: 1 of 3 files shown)


pages/content/amp-dev/events/amp-roadshow.html
```diff
@@ -109,7 +109,9 @@ <h4>Want to add cities?
       {% endif %}
 
       <aside class="ap-o-cities-map-grid">
+        {% set next_cities = [] %}
         {% for place in doc.roadshow.nextstop if doc.roadshow.nextstop %}
+          {% set push = next_cities.append(place.title|slug) %}
```
dlemm: Use `{% do %}` instead of `{% set %}`.

sebil: will `{% do %}`...

frontend/scss/components/molecules/event.scss
```diff
@@ -26,20 +26,22 @@ none
...
-    max-width: 350px;
-    min-width: 350px;
-    margin: 20px;
-    margin-left: 0;
+    flex: 0 0 calc(50% - 10px);
+    margin: 20px 0;
+  }
+
+  @media (min-width: 1024px) {
+    flex: 0 0 calc(33.33333333% - 20px);
```
matthiasrohmer: Don't overdo decimals here, they're capped anyways 😉Can you shorten down to two?

frontend/scss/components/molecules/event.scss
```diff
@@ -111,7 +115,7 @@ none
 
         display: inline-flex;
         align-items: center;
-        line-height: 1.4rem;
+        line-height: 1.4;
```
matthiasrohmer: Unit missing.

pages/content/amp-dev/events/amp-roadshow.html
```diff
@@ -109,7 +109,9 @@ <h4>Want to add cities?
       {% endif %}
 
       <aside class="ap-o-cities-map-grid">
+        {% set next_cities = [] %}
         {% for place in doc.roadshow.nextstop if doc.roadshow.nextstop %}
```
matthiasrohmer: Just curious: can't those two loops be unified? In a way the one that's also printing the cities is checking if it's already printed?

frontend/scss/components/molecules/event.scss
sebil: It's intended.

Issue #2512, opened 2019-07-16
Problem with build: grow.routing.routes.PathConflictError


fstanis:
### Current Behavior
`gulp develop` fails on second run with the following error:

```
Traceback (most recent call last):
  File "bin/grow", line 76, in <module>
  File "site-packages/click/core.py", line 764, in __call__
  File "site-packages/click/core.py", line 717, in main

  File "site-packages/click/core.py", line 1137, in invoke
  File "site-packages/click/core.py", line 956, in invoke
  File "site-packages/click/core.py", line 555, in invoke
  File "grow/commands/subcommands/run.py", line 54, in run
  File "grow/server/manager.py", line 77, in start
  File "site-packages/werkzeug/serving.py", line 1009, in run_simple
  File "site-packages/werkzeug/serving.py", line 962, in inner
  File "site-packages/werkzeug/serving.py", line 805, in make_server
  File "site-packages/werkzeug/serving.py", line 698, in __init__
  File "SocketServer.py", line 418, in __init__
  File "grow/server/manager.py", line 28, in server_activate
  File "grow/routing/router.py", line 66, in add_all
  File "grow/routing/router.py", line 100, in add_all_docs
  File "grow/routing/router.py", line 296, in add_docs
  File "grow/routing/router.py", line 54, in _add_to_routes
  File "grow/routing/routes.py", line 75, in add
  File "grow/routing/routes.py", line 255, in add
  File "grow/routing/routes.py", line 353, in add
  File "grow/routing/routes.py", line 372, in add
  File "grow/routing/routes.py", line 372, in add
  File "grow/routing/routes.py", line 372, in add
  File "grow/routing/routes.py", line 325, in add
grow.routing.routes.PathConflictError: Path already exists: /:locale/documentation/components/amp-story.html (<RouteInfo kind=doc meta={u'collection_path': u'/content/amp-dev/documentation/components/reference', u'pod_path': u'/content/amp-dev/documentation/components/reference/amp-story@fr.md'}> != <RouteInfo kind=doc meta={'pod_path': u'/content/amp-dev/documentation/components/reference/amp-story@pt_BR.md', 'collection_path': u'/content/amp-dev/documentation/components/reference'}>)
[201530] Failed to execute script grow
```

### Expected behavior/code
Build working. Even when ran more than once.

### How do we reproduce the issue?

```bash
git clone https://github.com/ampproject/docs
cd docs
npm i
gulp develop # wait for it to finish, then Ctrl+C
[...]

PR #2517, opened 2019-07-16
Add grow version check

fstanis:
To prevent dumb issues like #2512.

`gulp develop --skip-grow-version-check` disables it.

### DIFF
--- a/platform/config/shared.json
+++ b/platform/config/shared.json
@@ -3,5 +3,6 @@
 	 	"playground": "/#url=",
 		"repository": "https://github.com/ampproject/docs/blob/future/"
 	},
-  "gaTrackingId": "UA-67833617-1"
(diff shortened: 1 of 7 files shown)


platform/config/shared.json
```diff
@@ -3,5 +3,6 @@
 	 	"playground": "/#url=",
 		"repository": "https://github.com/ampproject/docs/blob/future/"
 	},
-  "gaTrackingId": "UA-67833617-1"
+	"gaTrackingId": "UA-67833617-1",
+	"growVersion": "0.8.8"
```
sebastianbenz: @matthiasrohmer Do we already support this version?

fstanis:
Can't we enforce it right now, at least ensure it's either 0.8.6 or 0.7.4?

matthiasrohmer:
Closed in favor of #2572.

PR #2519, opened 2019-07-16
Fix syntax highlighting

fstanis:


alabiaga:
Thanks Filip, can we also apply the change to the example under 

> Viewer server side template rendering

fstanis:
Done, fixed some identation too.

PR #2530, opened 2019-07-17
Make templates previewable

sebastianbenz:
* moves templates to a new root dir `templates
* adds a new build task to zip the templates
* adds a new static router serving the templates under http://localhost:8080/documentation/templates/preview/websites/gallery/templates/gallery.amp.html


sebastianbenz:
Being able to simply zip folders is still a good feature to have. Or is there a downside to keeping it?

matthiasrohmer:
Don't think so 🙂

PR #2531, opened 2019-07-17
🐛 Properly check for component version clash

### LINKED ISSUE #2488: Fix AMP same dependency error
Some example pages (e.g. /documentation/examples/components/amp-access/) throw the following error locally:

```
An error occurred building <grow.rendering.render_controller.RenderDocumentController object at 0x10fc14f90>

Using two versions (0.1, ) of the same dependency (amp-analytics) is not supported.
```

matthiasrohmer:
Fixes #2488. @lluerich, please check if that works for you 🙂

### DIFF
--- a/pages/extensions/amp_dependencies/__init__.py
+++ b/pages/extensions/amp_dependencies/__init__.py
@@ -34,16 +34,17 @@ class AmpDependencies(object):
     return '<AmpDependencies>'
 
   def add(self, name, version=None, type='element'):
-    tag, _version = self._dependencies.get(name, (None, version))
+    if version == None:
+        self._pod.logger.warning('Adding an AMP dependency ({}) without a specific version is not recommended.'.format(name))
+        version = COMPONENT_VERSIONS.get(name, DEFAULT_VERSION)
+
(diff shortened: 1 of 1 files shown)


lluerich:
Error still persists.

pages/extensions/amp_dependencies/__init__.py
```diff
@@ -34,17 +34,18 @@ def __repr__(self):
     return '<AmpDependencies>'
 
   def add(self, name, version=None, type='element'):
-    tag, _version = self._dependencies.get(name, (None, version))
+    if version == None:
+        self._pod.logger.warning('Adding an AMP dependency ({}) without a specific version is not recommended.'.format(name))
```
sebastianbenz: I'm not sure I agree. We should ideally always use the latest version. Having to explicitly define a version should be an exception (because it means there are breaking API changes).

matthiasrohmer: Oh we could also remove the warning, but this PR doesn't even add it. I was just rearranging code here.

matthiasrohmer:
Seems to have been an individual problem that doesn't ask for a fix.

PR #2544, opened 2019-07-18
♻️ Use different set of tags for SSR

matthiasrohmer:
We've been using tags that clashed with native JavaScript language patterns and therefore caused problems:

```js
{
        blockStart: '[%',
        blockEnd: '%]',
        variableStart: '[[', // Clashes with [[0, 2], [4, 5]] for example
        variableEnd: ]]',
        commentStart: '[#',
        commentEnd: '#]',
      }
```

The new set of tags is

```js
{
        blockStart: '<%',
        blockEnd: '%>',
        variableStart: '<=%',
        variableEnd: '%=>',
        commentStart: '<#',
        commentEnd: '#>',
      }
```

Which comes with the downside of being escaped by Grow in content documents which makes marking template string as safe necessary: 

```html
<p>Here you can find everything you need to build AMP experiences. The sidebar on the left shows you all resources for building <strong>AMP {{ "<=% format if format !== 'email' else 'emails' %=>"|safe }}.</strong></p>
```


sebastianbenz:
Are there any tags that don't not be escaped? Would be good it we could avoid it. Is it because of `<`?

matthiasrohmer:
Rolled back the changes and just replaced the variable tags with `[= variable =]` instead of `[[ variable ]]` to work around clashes.

PR #2548, opened 2019-07-18
♻️ Run amp-toolbox-optimizer on request

matthiasrohmer:
This moves optimizing out of the PageTransformer into the pages router.

### DIFF
--- a/platform/lib/routers/pages.js
+++ b/platform/lib/routers/pages.js
@@ -214,7 +214,7 @@ pages.get('/*', async (req, res, next) => {
   // Pipe the rendered template through the AMP optimizer
   try {
     renderedTemplate = await optimizer.transformHtml(renderedTemplate);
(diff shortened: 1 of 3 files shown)


platform/lib/routers/pages.js
```diff
@@ -208,6 +211,13 @@ pages.get('/*', async (req, res, next) => {
   // for this use case
   renderedTemplate = rewriteLinks(url.pathname, renderedTemplate, templateContext.format);
 
+  // Pipe the rendered template through the AMP optimizer
+  try {
```
matthiasrohmer: Not too sure if the optimizer's promise is ever rejected or if throws in general, you probably know better, @sebastianbenz 😉

sebastianbenz: I could fail so it's good to cache the error and return the original page.

PR #2551, opened 2019-07-18
♻️ Migrate poll backend

lluerich:
Migrated the backend/poll.go from ampbyexample.com to amp.dev.
It updates the example [here](https://amp.dev/documentation/examples/interactivity-dynamic-content/poll/?format=websites).

### DIFF
--- a/examples/source/interactivity-dynamic-content/Poll/index.html
+++ b/examples/source/interactivity-dynamic-content/Poll/index.html
@@ -80,7 +80,7 @@ Read more about [variable substitution]({{g.doc('/content/amp-dev/documentation/
       -->
       <form id="sample-poll"
             method="post"
-            action-xhr="submit"
+            action-xhr="./submit-poll"
             target="_blank">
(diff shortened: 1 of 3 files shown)


examples/source/interactivity-dynamic-content/Poll/index.html
```diff
@@ -80,7 +80,7 @@
       -->
       <form id="sample-poll"
             method="post"
-            action-xhr="<% hosts.backend %>/samples_templates/poll/submit"
+            action-xhr="submit"
```
sebastianbenz: nit: change to `./submit` to make it clear that this is an URL. Otherwise it looks a bit confusing. Even better: rename to endpoint to `submit-poll`.

lluerich: Makes it less confusing. Thanks for the feedback!

wassgha: This might've caused this #2671

PR #2558, opened 2019-07-19
Add Auckland and Porto Alegre, improve form links

morsssss:
Added two new cities.
Provided better links for Google Forms, so they don't show an option to request edit access.

Note that the cards no longer align that well :0 I'd love to get these new cities up, but we need to tighten the cards CSS so their width is more consistent.


morsssss:
I don't know why the Sydney/Melbourne commits are showing here... sorry!

PR #2570, opened 2019-07-21
Add gulp extract task, extract new strings, improve JA UI translation

pbakaus:
I noticed that our UI is still embarassingly untranslated so I:

1. Created a "gulp extract" task to be able to extract new strings into messages.pot regularly
2. Made a lot of untranslated strings translatable
3. Ported some untranslated strings over from the original Japanese messages.po file to the new one (this also needs to be done for all other languages, but is outside of the scope of this PR).

### DIFF
--- a/frontend/templates/views/partials/breadcrumbs.j2
+++ b/frontend/templates/views/partials/breadcrumbs.j2
@@ -20,7 +20,7 @@ and the current page title #}
     <svg class="ap-a-ico ap-m-breadcrumbs-angle"><use xmlns:xlink="http://www.w3.org/1999/xlink" xlink:href="#angle-down-solid"></use></svg>
   </span>
 
-  <div class="ap-m-breadcrumbs-crumb">{{ doc.collection.title }}</div>
+  <div class="ap-m-breadcrumbs-crumb">{{_(doc.collection.title)}}</div>
   {% endif %}
 </nav>
 {% elif doc.pod_path.startswith('/content/amp-dev/documentation/examples') %}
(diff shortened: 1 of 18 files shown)


pbakaus:
@matthiasrohmer I noticed that we have translated titles in files like https://github.com/ampproject/amp.dev/blob/future/pages/content/amp-dev/documentation/guides-and-tutorials/start/create/_blueprint.yaml using syntax that isn't documented on https://grow.io/reference/translations/.

Unfortunately they don't seem to work – I tested and confirmed that `{{ doc.collection.title }}` does not show the translated string (this happens in the breadcrumbs, for instance).

Interestingly that seems to be in line with Grow's documentation, which features the following example:

```
{{doc.title()}}      <!-- Untranslated -->
{{_(doc.title())}}   <!-- Translated -->
```

For now, I went with `{{_(doc.collection.title)}}` for that reason in breadcrumbs.j2, but very unfortunately, it does not pick up the translation from _blueprint.yaml, and rather picks it up from the messages.po file.

Not sure what the solution here is..

matthiasrohmer:
Thanks for flagging, @pbakaus! Actually that's a problem we've been putting off for a while: Grow doesn't support translated collection titles (or category names) natively in the same manner as it does for other fields, that's why I hacked around it for showing translated titles in the sidebar with

```jinja2
{% set collection_title = collection.tagged_fields.get('$title@' + doc.locale.language, collection.title) %}
```

I'll file an issue for the Grow team. In other respects your workaround should be fine for now.

Issue #2532, opened 2019-07-17
Edge Browser navbar


RedhatPH8:
## What's the issue?

Amp.Dev Website: Only the blog link is functional in the NavBar when using Microsoft Edge 44.17763.1.0; Microsoft EdgeHTML 18.17763.

I know Microsoft had plans to rework the edge browser using the Chromium project which might cause this issue to be null. My concern is the development of a site using this framework that would cause it to render incorrectly for my clients.

## How do we reproduce the issue?

1. Goto https://Amp.dev when utilizing Edge Browser

## What browsers are affected?

Microsoft Surface Book; Microsoft Surface Pro 4

## Which AMP version is affected?

AMP version is 1907152257550

matthiasrohmer:
@sebil, can you please investigate? Could be that the flyouts never show on Edge on a touch-device.

sebil:
I'm struggling getting this fully tested, but my guess is the `:focus-within` pseudo-selector causes that issue.

This is all the styling, which is responsible for revealing the flyout menus:
```scss
&-main-link:hover ~ &-flyout,
&-main-link:focus ~ &-flyout,
&-flyout:hover,
&-flyout:focus-within {
  opacity: 1;
  pointer-events: all;
  transition: opacity 0.2s cubic-bezier(0.25, 0.1, 0.25, 1);
}
```

You can [see here](https://caniuse.com/#feat=css-focus-within), that `:focus-within` is not supported in the current edge version.
I guess that edge ignores the whole css block, because of one _invalid_ property.

Splitting them up could fix that issue, like so:
```scss
%reveal-flyout {
  opacity: 1;
  pointer-events: all;
  transition: opacity 0.2s cubic-bezier(0.25, 0.1, 0.25, 1);
}

&-main-link:hover ~ &-flyout,
&-main-link:focus ~ &-flyout,
&-flyout:hover {
  @extend %reveal-flyout;
}

&-flyout:focus-within {
  @extend %reveal-flyout;
}
```

sebil:
Could finally test it. It is fixed now. I couldn't do it we the scss %placeholder like in the example above, because it ends up by combining the block again.

PR #2575, opened 2019-07-22
🐛 Fix not opening navigation dropdown on latest edge browsers

sebil:
I splitted the css block that handles all :hover, :focus and :focus-within properties of the flyout navigation, because edge doesn't supports the `:focus-within` prop.
By splitting this in two, edge stops ignoring the whole block and still reveal the flyout on :hover and :focus #2532


sebil:
The diff view is a bit confusing, you should check the raw file.

sebil:
Tested with build, everything still fine. I merge now.

Issue #2585, opened 2019-07-22
Fix grow unit test


sebastianbenz:
Current ` gulp buildForGrowTests --env test` fails with:

```
Building: 11/11 (in 0:00:01.0)

Error rendering /sitemap_grow.xml

/sitemap_grow.xml is an ignored path.
  File "/Users/sbenz/Library/Python/2.7/lib/python/site-packages/grow/rendering/render_batch.py", line 62, in render_func

    jinja_env=item['jinja_env']))
  File "/Users/sbenz/Library/Python/2.7/lib/python/site-packages/grow/rendering/render_controller.py", line 378, in render

    self.validate_path()
  File "/Users/sbenz/Library/Python/2.7/lib/python/site-packages/grow/rendering/render_controller.py", line 103, in validate_path

    raise errors.RouteNotFoundError(text.format(self.serving_path))

There were 1 errors during rendering.

Aborted!

```

PR #2591, opened 2019-07-23
Fix grow_sitemap.xml delivery and skip sitemap for grow tests

tharders:
Fixes https://github.com/ampproject/amp.dev/issues/2585 and also the sitemap delivery itself https://amp.dev/sitemap_grow.xml


platform/lib/common/growPageLoader.js
```diff
@@ -0,0 +1,62 @@
...
+ */
+async function fetchPage(pagePath) {
+  if (config.isTestMode()) {
+    // fetch original doc page from filesystem for testing
+    return readFileAsync(join(paths.PAGES_SRC, pagePath), 'utf-8');
+  } else if (config.isDevMode()) {
+    // fetch doc from proxy
+    return fetchPageFromGrowServer(pagePath);
+  } else {
+    // fetch comiled doc page from filesystem
```
sebastianbenz: nit: s/comiled/compiled

sebastianbenz: I know, I probably made that spelling mistake...

platform/lib/routers/pages.js
```diff
@@ -171,6 +172,13 @@ pages.get('/*', async (req, res, next) => {
     return;
   }
 
+  // the grow sitemap.xml must not be processed as template
+  if (/\.xml/.test(req.path)) {
```
sebastianbenz: Why not define this as a different route?

tharders:
I introduced a router for the grow sitemap as a generic grow xml router. @sebastianbenz  PTAL

platform/lib/routers/pages.js
```diff
@@ -165,7 +165,7 @@ pages.get('/*', async (req, res, next) => {
   // Let known file extensions automatically fallthrough as if they could not
```
sebastianbenz: Following the `growXml` naming scheme, it would make sense to rename this file into `growPages`.

platform/lib/routers/pages.js
```diff
@@ -165,7 +165,7 @@ pages.get('/*', async (req, res, next) => {
...
   // been resolved by the preceeding middleware the pages router can't
   // resolve them either
-  const KNOWN_FILE_EXTENSIONS = /\.(jpg|png|css|js|map)$/;
+  const KNOWN_FILE_EXTENSIONS = /\.(jpg|png|css|js|xml|map)$/;
```
sebastianbenz: Maintaining an explicit filter feels a bit fragile. Wouldn't it make more sense to check if there's no file extension or `.html`?

sebastianbenz: //cc @matthiasrohmer

matthiasrohmer: Jup, totally fine to turn it around to check for `/`, `$` and `.html` instead!

tharders: Implemented a matching pattern, so that the handler method is only called when the path does not point to a file with an extension other than html

tharders:
Ooops, @sebastianbenz approval was so fast I accidently requested the review again.

sebastianbenz:
@tharders np - please rebase and merge

platform/lib/routers/growPages.js
```diff
@@ -157,20 +157,12 @@ function rewriteLinks(canonical, html, format) {
...
-  // been resolved by the preceeding middleware the pages router can't
-  // resolve them either
-  const KNOWN_FILE_EXTENSIONS = /\.(jpg|png|css|js|map)$/;
-  if (KNOWN_FILE_EXTENSIONS.test(req.path)) {
-    next();
-    return;
-  }
-
+// only match urls with slash at the end or html extension or no extension
+growPages.get(/^(.*\/)?([^\/\.]+|.+\.html|.+\/)$/, async (req, res, next) => {
```
matthiasrohmer: This doesn't seem to work for the homepage: https://amp-dev-staging.appspot.com/

platform/lib/templates/index.js
```diff
@@ -91,50 +86,14 @@ class Templates {
...
-      }
-    });
-  }
-
-  /**
-   * Loads a template from cache (if not in dev mode). If the template
-   * is not cached, it will use the provided callback to retrieve the
-   * template string.
-   */
-  async compile(key, fn) {
```
matthiasrohmer: `compile` is still called in 

https://github.com/ampproject/amp.dev/blob/0281a56be55127312ed7af3887f33fcc9e792b2d/platform/lib/templates/index.js#L61

tharders: Fixed with https://github.com/ampproject/amp.dev/pull/2613

platform/lib/routers/growPages.js
tharders: fixed with https://github.com/ampproject/amp.dev/pull/2612

PR #2600, opened 2019-07-23
Remove binding from email sample

fstanis:
`binding` was removed from email spec so removing it from sample.


choumx:
Why was `binding` removed from the spec?

PR #2603, opened 2019-07-24
🐛 Switch relative to absolute paths

### LINKED ISSUE #2599: Use absolute URLs in email samples
It seems a few samples now use relative URLs as of #2249, #2299 and #2320.

Since emails require absolute URLs, it would be good to ensure we use those in our samples.

@lluerich, assigning to you since I'm unsure about the context behind this change.

lluerich:
fixes #2599

### DIFF
--- a/examples/source/interactivity-dynamic-content/Advanced_Server_Request.html
+++ b/examples/source/interactivity-dynamic-content/Advanced_Server_Request.html
@@ -49,7 +49,7 @@ When the form is submitted for the first time, the `amp-list` is hidden and the 
     Note: Use a placeholder to make the loading state look the same as the submitting state for the form.
   -->
   <div>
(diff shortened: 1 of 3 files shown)


fstanis:
Thanks!

Should we also update images to not use `https://amp.dev` directly, like e.g.
https://github.com/ampproject/amp.dev/blob/9f17b0e71f7da9edfdf3b5ec4c8b4a70233d53b1/examples/source/interactivity-dynamic-content/Conference_Survey_Email.html#L52

[examples/static/samples/standalone/amp-email/index.html](https://github.com/ampproject/amp.dev/blob/future/examples/static/samples/standalone/amp-email/index.html) also has a few "hardcoded" domain links, not sure what that file is used for.

PR #2606, opened 2019-07-24
Fix import of templates for inline example preview

tharders:
Fix stupid copy paste error that prevents the import of the amp-mustache template here:
https://amp-dev-staging.appspot.com/tests/examples/template/


tharders:
@matthiasrohmer PTAL

PR #2611, opened 2019-07-25
Fix roadshow dates

morsssss:
Some dates need to be updated!


morsssss:
Looks like I need to make a new branch next time around, since all sorts of old commits are being included here 😭

PR #2594, opened 2019-07-23
✨ Add SSR for success story overview page

robinvanopstal:

PR #2593, opened 2019-07-23
✨ Update template overview page including SSR

robinvanopstal:

PR #2614, opened 2019-07-25
♻️ Move pageTransformer logic into gulp task

### LINKED ISSUE #2508: SSR rendered tools page fails with amp-toolbox-optimizer
### Current Behavior
https://amp.dev/documentation/tools/ shows an error page (response code is 200, but the page shows "500")

### Expected behavior/code
A list of tools that support AMP

### How do we reproduce the issue?
Navigate to the URL

1. Open https://amp.dev/documentation/tools/ in a browser

What browser are you using?
Chrome 75

What O/S are you using?
MacOS 10.15

matthiasrohmer:
Unblocks #2601, #2594 and #2593. Fixes #2508. Additionally this fixes an unreported bug: `buildPages` was marked as `async` and was therefore not properly handled by gulp.

In summary this PR removes the separate PageTransformer class and integrates its remaining logic (only minifying, as optimizing and filtering were moved to the server side) into a gulp task called `minifyPages`. At the same time sharding was removed from the transforming process to reduce complexity and as the executed operations are not that costly anymore.


gulpfile.js/build.js
```diff
@@ -321,56 +353,57 @@ async function buildPages(done) {
...
-    return Promise.all(shards);
-  }
+  // Only shorten selectors that are actually safe to be rewritten (i.e. are
+  // not used inside `amp-bind` statements for example)
+  rcs.selectorLibrary.setExclude(
+      new RegExp('^(?!' + SELECTOR_REWRITE_SETTINGS.safeSelectors.join('|') + ').*$')
+  );
+
+  return gulp.src(`${project.paths.GROW_BUILD_DEST}/**/*.html`)
+      .pipe(through.obj(async function(page, encoding, callback) {
```
sebastianbenz: does this need to be async?

gulpfile.js/build.js
```diff
@@ -321,56 +353,57 @@ async function buildPages(done) {
...
+        let html = page.contents.toString();
+
+        // Compress multiple spaces down to only one
+        html = html.replace(/\s{2,}/gm, ' ');
+
+        // Extract the CSS in order to minify it
+        const css = html.match(/<style amp-custom>(.*?)<\/style>/ms);
+        if (css) {
+          const minifiedCss = cleanCss.minify(css[1]).styles;
+          html = html.replace(css[1], minifiedCss);
```
sebastianbenz: this probably would be more efficient using indexes from the match instead of performing another string comparison.

matthiasrohmer: A little more verbose, but yes good idea!

gulpfile.js/build.js
matthiasrohmer: Good catch, thanks!

PR #2636, opened 2019-07-29
✨Add Toggle for advanced and beginner mode on course sites

### LINKED ISSUE #2295: Design to toggle "compressed/experienced developer" view of courses

dlemm:
Fixes #2295 Adding a toggle to courses sites in order to display advanced content via Filter

```
[filter level="advanced"]
"advanced content"
[/filter]
```

Bildschirmfoto 2019-07-29 um 16 43 43

Bildschirmfoto 2019-07-29 um 16 47 28


morsssss:
Thanks for doing this! And... it looks quite nice!

Is there a simpler way to portray the control? I'd imagined a toggle between "Beginner" and "Advanced". If you prefer a way to toggle "Expert mode" on and off, maybe we can shorten it to just "Expert mode", or "Expert mode on" / "Expert mode off"?

I haven't gotten to try the code out yet, but I will do that too :) thanks!

dlemm:
Hey Ben, thanks for the feedback. Yes, I think the wording is not final and can be adjusted. Like this:

Bildschirmfoto 2019-07-30 um 10 41 27

Bildschirmfoto 2019-07-30 um 10 40 50


I haven't included the toggle in the files yet. I think it is easier to add them to the final courses files? While testing just add
```
{% include '/views/partials/level-toggle.j2' %}
```
to the md-files.

morsssss:
I'm seeing the toggle... but for some reason I'm also seeing template conditionals output to the page... any idea what I'm missing here?

Screen Shot 2019-07-31 at 12 48 16 PM

dlemm:
Hm, I'll check on that. But it looks like the Server Site Rendering is not working. Are there perhaps any dependencies missing or is it maybe an older grow (0.8.8) version?

morsssss:
It's definitely an older grow version! I think I once updated and broke my build, so I've avoided updating ever since.

Sounds like it's time to upgrade :)

morsssss:
Upgraded, and futzed around some more, and I still get these SSR errors...

I have done `npm install` too.

Maybe @matthiasrohmer would have an idea about what I'm missing here?

dlemm:
I am sorry that it hasn't worked out. I've just tried it on a different computer and it worked. 🤔 So yeah, maybe Matthias has an answer.

morsssss:
I'm about to head out for a week of vacation. So if we can't figure this out today, then don't wait for my approval :)

Just please change the text in your tooltip to something like this:

     <h4>Expert mode</h4>
     <p>Use expert mode to hide our web development guidance for beginners.</p>

matthiasrohmer:
Only thing I can think of is that you have been accessing the wrong host on the other machine 🙂So localhost:808**1** instead of localhost:808**0**.

morsssss:
Oh. Um... well, yeah, maybe I was.

If I use port 8080, all works well 💩

Trying this out now!

morsssss:
Yay, it works!

My favorite text is:

      <h4>Expert mode</h4>
      <p>Use expert mode to web development guidance that's intended for beginners.</p>

I wish of course that this toggle didn't involve a page refresh, but I know that's how it has to be. LGTM!

matthiasrohmer:
Oh not necessarily. The problem is that sharing would get confusing if we do it with `amp-bind` as there's no chance for an AMP site to update the URL with the correct GET parameter. As `History.pushState` would in comparison to `AMP.pushState`.

morsssss:
Wow, good point. That's all that's stopping us from avoiding a page refresh!

Do I see a FR for AMP?

matthiasrohmer:
At least I think so. Looping in @sebastianbenz as we've discussed this for the tools page as well and I can't quite recall if he explained to me if there's a specific reason that that isn't a feature already 😉

dlemm:
I copied the changes of PR #2640 for the courses sites and added the frontmatter parameter leveled: true in order to show the toggle button for sites that need the button.

sebastianbenz:
Is this ready for review?

dlemm:
Sorry, yes the toggle is ready for review.

frontend/scss/components/molecules/level-toggle.scss
```diff
@@ -0,0 +1,224 @@
...
+  @media (min-width: 768px){
+    width: 500px;
+    height: auto;
+
+    h5 {
+      font-size: 0.875em;
+    }
+  }
+
+  h5 {
```
sebastianbenz: this will override the media query specific rule above

frontend/scss/components/molecules/level-toggle.scss
```diff
@@ -0,0 +1,224 @@
...
+        opacity: 0;
+        z-index: 10;
+      }
+
+      &-switch {
+        display: flex;
+        width: 70px;
+        height: 35px;
+        border-radius: 2px;
+        background-color: #2773F1;
```
sebastianbenz: have we got a color var for this?

frontend/scss/components/molecules/level-toggle.scss
```diff
@@ -0,0 +1,224 @@
...
+            height: 7px;
+            width: 7px;
+            margin: 9px 0 0 9px;
+          }
+        }
+      }
+    }
+  }
+
+  .show-tip {
```
sebastianbenz: this is usually better handled by AMP's built-in `hidden` attribute

frontend/templates/views/partials/level-toggle.j2
```diff
@@ -0,0 +1,55 @@
+{% do doc.styles.addCssFile('css/components/molecules/level-toggle.css') %}
+{% set level = doc.level %}
+{% set url = doc.url.path %}
+
+
+
+<amp-state id="filter" class="" hidden="" aria-hidden="true">
```
sebastianbenz: are these needed? `class="" hidden="" aria-hidden="true"`

frontend/templates/views/partials/level-toggle.j2
```diff
@@ -0,0 +1,55 @@
...
+  }
+  </script>
+</amp-state>
+
+
+<div class="ap-m-level-toggle">
+  <div class="ap-m-level-toggle-elements">
+    <div class="toggle-button">
+      <input
+      [% if level == 'advanced' %]
```
sebastianbenz: nit: I'd prefer to only switch on the content that changes and avoid duplicating strings

pages/content/amp-dev/documentation/courses/advanced-course/appendix.md
```diff
@@ -70,3 +72,73 @@ Throughout these courses, you will need various images to complete the exercises
...
+[/sourcecode]
+
+If you want a full specification, you can look at the [official documentation](https://www.json.org/).
+
+## What is Deep Merge?
+
+We explained that when `AMP.setState()` is called, `<amp-bind>` will "deep-merge the provided object literal with the current state." The deep-merge functionality will not overwrite anything, but will rather merge the two objects. [In the example]({{g.doc('/content/amp-dev/documentation/components/reference/amp-bind.md', locale=doc.locale).url.path}}#deep-merge-with-ampsetstate) on the `<amp-bind>` documentation, we see that providing only an age for the employee will just update the value of the age of the employee in the state.
+
+[sourcecode:html]
+{% raw %}<button on="tap:AMP.setState({employee: {name: 'John Smith', age: 47, vehicle: 'Car'}})"...></button>
```
sebastianbenz: //cc @morsssss it's better to pretty print bindings and actions to make them easier to understand:

```
<button on="tap:AMP.setState({
  employee: {
    name: 'John Smith', 
    age: 47, vehicle: 'Car'
  }
})". ...>
</button>
```

platform/lib/templates/index.js
```diff
@@ -39,6 +40,12 @@ function createRequestContext(request={'query': {}}, context={}) {
     context.format = request.query.format;
   }
 
+  if (!ALLOWED_LEVEL.includes(request.query.level)) {
```
sebastianbenz: nit: negated boolean expressions are harder to understand. In this case it's simple to avoid this by switching if and else.

sebastianbenz:
The courses also don't render locally for me: 

```
An error occurred building <RenderDocumentController(/content/amp-dev/documentation/courses/advanced-course/index.md)>
Encountered unknown tag 'endcall'.
```

frontend/scss/components/molecules/level-toggle.scss
sebastianbenz: ping

frontend/scss/components/molecules/level-toggle.scss
sebastianbenz: ping

frontend/scss/components/molecules/level-toggle.scss
sebastianbenz: my fault

frontend/scss/components/molecules/level-toggle.scss
sebastianbenz: my fault

sebastianbenz:
The courses still don't render for my. Am I missing something?

pages/content/amp-dev/documentation/courses/advanced-course/appendix.md
morsssss: Thanks!

//cc @robocel

morsssss:
I'm seeing the same issue...

morsssss:
I'm hoping that we can get this in soon... because I'd like to use the compressed courses at our BrazilJS learning event on August 22!

Just saying...

sebastianbenz:
@dlemm please make sure to get the issues get fixed asap so that we can get this merged. Or are there any blockers that I'm missing?

Issue #2053, opened 2019-05-10
Migrate inline doc samples to preview.amp.dev


sebastianbenz:
Documentation pages still use the old inline samples with preview, e.g. here: https://amp.dev/documentation/guides-and-tutorials/develop/style_and_layout/responsive_design

We should improve these by directly rendering the source code into the doc page and serve the preview from preview.amp.dev.

You can find the existing samples here: https://github.com/ampproject/docs/tree/master/examples/src

@matthiasrohmer @lluerich @tharders does anyone of you want to look into this?

CrystalOnScript:
I'm unsure if this issue is related or not, but while sandboxing at I/O I noticed none of the inline examples are working with the component documentation either, such as the `amp-carousel`

https://amp.dev/documentation/components/amp-carousel?format=websites

sebil:
I'm reopening this issue, because of an unfinished styling task.

PR #2637, opened 2019-07-29
✨ Implement example code blocks with preview in iframe

sebil:
We now support example code blocks in markdown files with preview in iframe: `[example preview="top-frame" orientation="portrait"]`

This PR contains:
* Changed grow [example] plugin to support new mode and 'orientation' attribute
* Changed Templates to render the preview
* Guides and Tutorials with examples that did not work in `inline` preview mode now use `top-frame` (see attachments)
* Information in the documentation for editors
* Fix for standalone example code generation from inline code snippets
* Fix for reload button of standalone example (ampbyexample) pages

Closes: #2053 

top-frame-example-video
top-frame-example-landscape
top-frame-example-portrait


contributing/documentation.md
```diff
@@ -179,15 +179,32 @@ You can let a code sample have a preview or a link to open the code sample in th
   </code>
 </pre>
 
-An inline preview is only possible if the code does not contain any head elements.
-For email content with resource links use the placeholder `{{server_for_email}}`.
+Use the `preview` attribute to define how the preview is generated:
+- **none**: No preview will be generated
+
+- **inline**: The example is included directly in the page above the source code display.
+  An inline preview is only possible for normal website examples if the code does not contain any head elements.
```
sebastianbenz: Mark `head` as code

```
s/head/`head`/g
```

contributing/documentation.md
```diff
@@ -179,15 +179,32 @@ You can let a code sample have a preview or a link to open the code sample in th
...
+  Use this option for small examples that do not need any styling or other head elements
+  (imports do not count, since they are specified via the `imports` attribute).
+  
+- **top-frame**: The preview is shown above the example in an iframe and supports switching the orientation.
```
sebastianbenz: Rephrase: "The preview is shown above the example in an iframe and supports switching the orientation."

to

"The example preview is displayed above the source code inside an iframe. The orientation can be toggled between `portrait` and `landscape` mode."

contributing/documentation.md
```diff
@@ -179,15 +179,32 @@ You can let a code sample have a preview or a link to open the code sample in th
   </code>
 </pre>
 
-An inline preview is only possible if the code does not contain any head elements.
-For email content with resource links use the placeholder `{{server_for_email}}`.
+Use the `preview` attribute to define how the preview is generated:
+- **none**: No preview will be generated
+
+- **inline**: The example is included directly in the page above the source code display.
```
sebastianbenz: rephrase: "The example is included directly in the page above the source code display"

to 

"The example preview is displayed above the source code"

contributing/documentation.md
```diff
@@ -179,15 +179,32 @@ You can let a code sample have a preview or a link to open the code sample in th
...
+  An inline preview is only possible for normal website examples if the code does not contain any head elements.
+  Use this option for small examples that do not need any styling or other head elements
+  (imports do not count, since they are specified via the `imports` attribute).
+  
+- **top-frame**: The preview is shown above the example in an iframe and supports switching the orientation.
+  You can select the default orientation by specifying the additional attribute...
```
sebastianbenz: s/.../:/

contributing/documentation.md
```diff
@@ -179,15 +179,32 @@ You can let a code sample have a preview or a link to open the code sample in th
...
+  - **orientation**: `default: landscape|portrait`
+  
+    The `responsive` mode shows the preview over the full width of the content.
```
sebastianbenz: The responsive mode is not mentioned before.

sebastianbenz:
Good idea with putting the tool icons in the middle. Few high-level comments:

* it'd be super cool if we could animate the preview rotation 🤪
* on mobile, the preview does not rotate, but expand to the bottom. We should keep the preview dimensions during rotation.

contributing/documentation.md
tharders: Thanks for catching this

sebil:
Hi @sebastianbenz ,

after a long try getting the preview rotate animation as intuitive and 'realistic' as on real-life devices, we decided sticking with this solution. (See screen capture below)

What I basically did is fading and rotating the `<amp-iframe>` (child) while morphing the `<div>` (parent) to it's new aspect ratio. Additionally I gave the parent `background-color: black` and set `overflow: hidden` to make it feel more like a real device.

**IMPORTANT NOTE:** To morph the size of the preview I had to transition `padding` and `max-width` of child and parent element. I know that these are not part of the **white listed properties for keyframes**, but in this case it is necessary to get the desired effect and since it causes no amp validation error, I think we can shut one eye on this... what do you think?

ezgif com-video-to-gif

P.S. the other thing is also fixed now, so the preview should keep it's dimensions anytime on any resolution.

P.P.S. I also got a fallback solution, if you don't like the rotation. It's just morphing the size of the preview, without fading and rotating anything.

But I hope you like it 👍 

> Good idea with putting the tool icons in the middle. Few high-level comments:
> 
> * it'd be super cool if we could animate the preview rotation 🤪
> * on mobile, the preview does not rotate, but expand to the bottom. We should keep the preview dimensions during rotation.

sebastianbenz:
Impressive! Thanks for trying! However, we shouldn't use non-hardware accelerated properties for the animation (we have to be a good example :-) ). What does the fallback look like?

sebil:
The fallback looks like this, but unfortunately it also uses non-hardware accelerated properties:

ezgif com-video-to-gif (1)

The `padding-top` of `<amp-iframe>` is changing from `56.25%` to `177.78%` (16:9 to 9:16) and the `max-width` of the parent `<div>` is changing from `56.25%` to `100%`.

Both elements have the `transition` property, which I'll remove now.

I'll try a new solution with just transform and opacity.

sebil:
Okay, 2 new solutions:

1. I removed the `transition` properties and left the rotating animation as it was. This causes that the code block is jumping up and down, because of different dimension heights.

2. I gave the landscape frame a dynamic margin, so it takes the same height as in portrait mode. This fixes the jumping code block issue, but waste space in landscape mode.

Screen captures:

First: code block is jumping up and down
ezgif com-video-to-gif (2)

Second: code block stands still, but space is wasted
ezgif com-video-to-gif (4)

sebastianbenz:
Awesome! I think the first one is perfect! The jumping code block is not a problem.

sebil:
I'm glad you like it!

sebil:
@sebastianbenz This PR is ready for official review now. Don't be suprised, I just renamed it.

LongoOfficial:
> Awesome! I think the first one is perfect! The jumping code block is not a problem.

@sebastianbenz, I wouldn't recommend that compromise. without the animation of the iframe container between the two states, the rotation of only the content makes the example feels broken and not meaningful anymore. Although we put some effort to it, wouldn't it be better for the user if we do not have an animation at all?

sebil:
Hey @sebastianbenz 
like we discusses I tried to give it a last shot based on your suggestion and here are my results:

I could achieve a morphing effect for the iframe container by adding a background div and change its `transform: scale()` values to fit in the correct dimensions.

This method brings two major issues with it: Sometimes the iframe is flickering and if you do a stress-test by rapidly clicking the button, it doesn't behave like expected. It looks broken, see screen captures below

1. On **white** background color (your suggestion)
ezgif com-video-to-gif (6)

2. On **black** background color (as before)
ezgif com-video-to-gif (7)


So @LongoOfficial and I wouldn't recommend this solution and we should rather stick without an animation then.

What do you think?

sebil:
I decreased the `rotation angle` from `75deg` to `45deg` and the `animation duration` from `650ms` to `350ms` so it's quite faster and snappier now compared to the old version without morphing.

I like it better now, what do you say?

### **New version:** `350ms`, `45deg`

ezgif com-video-to-gif (8)

### **Old version:** `650ms`, `75deg`

ezgif com-video-to-gif (2)

sebastianbenz:
I think `350ms, 45deg` works best so far, maybe even only `250ms`. Did you try the version without morph and animated placeholder?

sebil:
As discussed I removed the rotation and speed up the morphing background container to `250ms`.

Here is the result:
ezgif com-video-to-gif (1)

sebastianbenz:
Are we good to merge or is anything missing?

sebil:
From my side we are ready. The only things that are missing are planned for the next steps, eg. the 'side-frame' preview, update for the sampler, etc.

PR #2670, opened 2019-07-31
♿️ Support small and low-end devices with 240px width

sebil:
This PR optimizes the amp.dev for low-end and small devices up to 240px width.

Most things look accurate now and are at least consumable/accessible, even though some things don't look 100% perfect, because they used to be designed for bigger devices. But nothing looks broken.

**TBD** if you are fine with decreasing the base `font-size` from 16 to 14px. See commits.

Here is a screen capture of the home page:

ezgif com-optimize

Closes: #2670


frontend/scss/base.scss
```diff
@@ -11,14 +11,18 @@
 html, body {
   padding: 0;
   margin: 0;
-  min-width: 320px;
+  min-width: 240px;
```
sebastianbenz: Can we extract the 240px into a variable?

sebil: Good point. Done

PR #2703, opened 2019-08-05
Update signed-exchange.md

ithinkihaveacat:
Two substantive changes + wording cleanup. The substantive changes:

1. Change the `amp-cache-transform` header to include the version parameter `v`. This is always issued, and some developers thought it was sufficient to do a strict string match on `google`, which is not sufficient. A "large" version range is used in an attempt to ensure the doc remains accurate for as long as possible. It's still possible for this test to succeed but for indexing to fail if the version of the response isn't acceptable to the indexing system, but I'm not sure how to fix this without some laborious explanation. (Also, I'm not sure where the currently acceptable versions are documented.)
2. Add the test that `vary: accept,amp-cache-transform` should always be returned. This is mentioned in e.g. https://github.com/ampproject/amppackager/blob/releases/README.md but didn't make it to this doc.

Also some separate wording changes to the first paragraph that I think are improvements, but happy to drop if you disagree.

/cc @sebastianbenz @twifkak

### DIFF
--- a/pages/content/amp-dev/documentation/guides-and-tutorials/optimize-measure/signed-exchange.md
+++ b/pages/content/amp-dev/documentation/guides-and-tutorials/optimize-measure/signed-exchange.md
@@ -130,7 +130,11 @@ content-type: application/signed-exchange;v=b3
 ```
 
 [tip type="important"]
-The `v=b3` version string is the current version. This version will change. 
+The `v="1..100"` in the request will change. Do not match on this exact string; instead split on `;` and look for `google`.
+[/tip]
+
+[tip type="important"]
+The `v=b3` version string in the response is the current version as of August 2019. This version will change. 
 [/tip]
 
 The bulk of the response should be your AMP page (in plaintext). There's a small binary header, and, if the page is >16kb, a few binary bytes sprinkled throughout.
--- a/pages/content/amp-dev/documentation/guides-and-tutorials/optimize-measure/signed-exchange.md
+++ b/pages/content/amp-dev/documentation/guides-and-tutorials/optimize-measure/signed-exchange.md
@@ -117,10 +117,10 @@ See [`packager.js`](https://github.com/ampproject/docs/blob/future/platform/lib/
 
 ### Testing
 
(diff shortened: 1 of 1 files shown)


pages/content/amp-dev/documentation/guides-and-tutorials/optimize-measure/signed-exchange.md
```diff
@@ -130,27 +130,37 @@ content-type: application/signed-exchange;v=b3
 ```
 
 [tip type="important"]
-The `v=b3` version string is the current version. This version will change. 
+The `v="1..100"` in the request will change. Do not match on this exact string; instead split on `;` and look for `google`.
```
twifkak: My advice, for folks running `amppackager`, is look only for presence of the `amp-cache-transform` header, and delegate value-parsing to `amppkg`. This is mainly so it can later be extended easily to support more AMP SXG caches (pending https://github.com/ampproject/amppackager/issues/156), and have publishers pick that change up automatically. Totally open to discussion on this point, e.g. if that's not feasible or advisable.

Also, this is under the "Testing" section. I'm not sure if there is a better section, so just a FYI there.

ithinkihaveacat: Adjusted the text in https://github.com/ampproject/amp.dev/pull/2703/commits/22b87660af02130fdcbef9d5003bb09a07cac8b2 to suggest checking for the existence of `amp-cache-transform` only (i.e. ignore value). My original thought was that it's technically possible for a non-google packaging solution to exist (also why the `google` value exists at all) but yeah, this doc is probably not a helpful place to make this point...

twifkak: Ah I see; I wasn't considering that angle; yeah, agreed this is probably not the doc for that.

(There's the other angle of somebody running a non-amppkg solution but meant for Google; in that case, though, they'll want to pay attention to target _and_ version #s in that case.)

PR #2704, opened 2019-08-05
♻️ Migrate/hotel

lluerich:
Migrated the hotel example backend from [glitch.com](https://glitch.com/~amp-by-example-live) to amp.dev.
It updates the example [here](https://amp.dev/documentation/examples/e-commerce/hotel/?format=websites).

I added some dates in the ```rooms.json``` so that the backend will check if those are available on selected dates.

### DIFF
--- a/examples/source/e-commerce/Hotel/api.js
+++ b/examples/source/e-commerce/Hotel/api.js
@@ -22,9 +22,6 @@ const rooms = require(utils.project.absolute('/examples/static/samples/json/room
 
 // eslint-disable-next-line new-cap
 const examples = express.Router();
-examples.use(express.urlencoded({
-  extended: true,
-}));
 
 examples.all('/rooms', (request, response) => {
   setMaxAge(response, 0);
--- a/examples/source/e-commerce/Hotel/api.js
+++ b/examples/source/e-commerce/Hotel/api.js
@@ -1,0 +1,47 @@
+/**
+ * Copyright 2019 The AMP HTML Authors. All Rights Reserved.
+ *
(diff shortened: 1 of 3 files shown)


examples/source/e-commerce/Hotel/api.js
```diff
@@ -0,0 +1,44 @@
...
+
+const express = require('express');
+const {setMaxAge} = require('@lib/utils/cacheHelpers');
+const utils = require('@lib/utils');
+const rooms = require(utils.project.absolute('/examples/static/samples/json/rooms.json'));
+
+// eslint-disable-next-line new-cap
+const examples = express.Router();
+
+examples.all('/rooms', (request, response) => {
```
sebastianbenz: s/all/get

examples/source/e-commerce/Hotel/api.js
```diff
@@ -0,0 +1,44 @@
...
+  setMaxAge(response, 0);
```
sebastianbenz: max-age can be long as results will always be the same for the given query parameters

lluerich: Do you have any particular duration in mind? A year?

sebastianbenz: make it 1d in case we want to change the sample

lluerich: Alright, PTAL

examples/source/e-commerce/Hotel/api.js
lluerich: PTAL

sebastianbenz:
Thanks!

Issue #2412, opened 2019-06-30
Implement a site search


virgiltu:
For the love of god, can we please get a from or maybe a link to the google forum that matches amp. Also would be nice to have something like codepen where people can put their work. If i build a nice amp module, i can showcase it so other can use it. And also a god dam search on the documentation. I know this was requested before. 

These have to be the top must have to get this more popular. I know so many people that got frustrated and just gave up on amp because it was to hard to find help or took hours to find something.

pbakaus:
Hi there, we definitely hear you, and understand your frustration.

We've recently redesigned/relaunched our site, and the Search feature has not been ported over yet. It's obviously critical as you pointed out, and it's the next high priority item for our devs. Looping in @matthiasrohmer, who is I believe starting work on it in the next days.

As for the others:

1. Codepen works with AMP! There's also https://playground.amp.dev/, but the Playground doesn't have save functionality. You make a good point, it'd be awesome to make it easier to contribute to our examples, and offer more CodePen like functionality..
2. For AMP on Google, consider the webmasters forum: https://support.google.com/webmasters/community?hl=en. We have nothing to do with it though.

Stay tuned for more updates, thanks for the report!

virgiltu:
This is amazing news. No worries with all the frustration I keep my spirits up to deliver amazing content.

matthiasrohmer:
Thanks for your feedback, @virgiltu! I can only second what Paul already said. We are working everyday to make AMP more accessible to developers - especially by improving amp.dev

As search is indeed one of the next top priority features and there's no publicly tracked issue for it yet, I am going to rephrase the title of this one here. Hope that's fine for you!

tharders:
I will close this, since the search is live now.
Problems and improvements for the search should be addressed in a new issue.

PR #2705, opened 2019-08-05
Search functionality

tharders:
This is the basic implementation for #2412 using Google Custom Search with JSON API.


tharders:
Notes to the current state: 
* Search form implemented using amp-form with amp-autocomplete and amp-list
* Components are shown at the top when they are in the first 10 search results
  * When a component example exists, the example link and a playground link is shown below the component. The example itself might show up again in the normal search results 
* Language filtering is currently done by Googles language parameters (hl, lr) and not by path. With this sometimes results in other languages are shown.
* Google will find pages where the keyword is only in the navigation. Sometimes the result is quite high in the result list. Would be good if we can improve this. 
* The search is currently implemented as an extra page.

The search is currently not linked in the navigation an can be found under /search/

Test search words:
* positive
  * responsive images: Example shows a nice result list for normal text queries. It is nice that also the amp-img component is shown.
  *  amp-access: The components listed at the top make sense for amp-access
* negative
  * amp-accordion: amp-lightbox is shown, although amp-accordion is only in the navigation
  * amp-3d-gltf: Chinese content and also lots of other components where only the navigation has the keyword

sebastianbenz:
Woohoo! This is amazing!

Few comments from a quick look:

* Works really well for the first results!
* As you said, due to navigation, way to many pages get included (e.g. query 'hotel'). This is a problem we need to fix. 
* Playground is included in the normal results
* Example Demos is listed separately, maybe these should be merged into the sample results.

tharders:
A few more comments:
* The autosuggest list has all components taken from the amp component version list, the build in components and some elements that are part of a component
* The Google Custom Search JSON API is limited to a maximum of 100 results. No real problem, but I have added a text that is shown to the user if not all results are shown.
* Currently the first search request is send twice (once by form submit and once by amp-list).
Using the amp-list with the result from amp-form in a state prevented the load-more.

pages/content/amp-dev/search.html
```diff
@@ -0,0 +1,155 @@
...
+      </amp-autocomplete>
+      <input name="page" type="hidden" value="1">
+      <input name="locale" type="hidden" value="{{ doc.locale }}">
+      <button type="submit">Search</button>
+    </form>
+    <div [hidden]="showResults != true" hidden>
+      <p [text]="search.result.totalResults + ' results found'" class="info"></p>
+
+      <amp-list [height]="(100) * search.result.components.length + (80) * search.result.pages.length"
+                [src]="'/search/do?q=' + query + '&page=' + pageNumber + '&locale={{ doc.locale }}'"
```
sebastianbenz: Why not bind to `[src]=search`?

tharders: `load-more-bookmark="nextUrl"` did not work with that :-(

matthiasrohmer:
This is not document too well but would `<!-- googleoff:all -->Not indexed.<!-- googleon:all -->` work to exclude sidebars, header and footer?

sebastianbenz:
@matthiasrohmer I was thinking the same. I've asked the CSE team to confirm.

tharders:
@sebastianbenz 

> * Playground is included in the normal results

I removed all playground urls except the start page.
I like the idea that I can find the playground with the search.
Also the blog content is found.


> * Example Demos is listed separately, maybe these should be merged into the sample results.

Currently we only separate Components, and there we only show the link to the basic example.
We could add the demo link for this basic example there.
And maybe we should remove the demos from the rest of the search?

sebastianbenz:
* +1 to adding only playground home
* perfect that blog content is found as well
* I agree, let's remove demos from the search

tharders:
> This is not documented too well but would `<!-- googleoff:all -->Not indexed.<!-- googleon:all -->` work to exclude sidebars, header and footer?

I fear this was only working in Google Site Search.
I found a `class="NOCONTENT"` notation that should have worked once in CSE, but I cannot find any recent documentation for that.

pages/content/amp-dev/search.html
sebastianbenz: ah - too bad. One thing you can do is trigger the amp-list request straight away: `on="submit: AMP.setState({'query': tempQuery, 'pageNumber': 1, 'showResults': true})`. You can then set the action to a dummy endpoint, e.g.: `action-xhr="https://amp.dev/documentation/examples/api/echo"`.

pages/content/amp-dev/search.html
```diff
@@ -0,0 +1,155 @@
...
+        <input name="q" on="change:AMP.setState({ 'tempQuery': event.value })"
+               placeholder="What are you looking for?"
+               required>
+      </amp-autocomplete>
+      <input name="page" type="hidden" value="1">
+      <input name="locale" type="hidden" value="{{ doc.locale }}">
+      <button type="submit">Search</button>
+    </form>
+    <div [hidden]="showResults != true" hidden>
+      <p [text]="search.result.totalResults + ' results found'" class="info"></p>
```
sebastianbenz: You can move this into the amp-list if you use `single-item` mode.

sebastianbenz: (if that's supported with infinite scroll)

pages/content/amp-dev/search.html
```diff
@@ -0,0 +1,155 @@
...
+    <h1>Search</h1>
+    <form action-xhr="/search/do" class="search-form" id="searchForm" method="GET" on="submit-success: AMP.setState({'query': tempQuery, 'pageNumber': 1, 'showResults': true, 'search': event.response})"
+          target="_top">
+      <amp-autocomplete filter="substring"
+                        min-characters="0"
+                        on="select:AMP.setState({ 'tempQuery': event.value }),searchForm.submit"
+                        submit-on-enter="false"
+                        src="/search/autosuggest"
+      >
+        <input name="q" on="change:AMP.setState({ 'tempQuery': event.value })"
```
sebastianbenz: Have you tried using `input-debounced` instead of `change`?

platform/lib/routers/search.js
```diff
@@ -0,0 +1,268 @@
...
+
+const IMPORTANT_INCLUDED_ELEMENTS = [
+  'amp-state',
+  'amp-list-load-more',
+  'amp-story-page',
+  'amp-story-grid-layer',
+  'amp-story-bookend',
+];
+
+const COMPONENT_VERSIONS_PATH = path.join(project.paths.GROW_POD,
```
sebastianbenz: We should migrate this to: https://github.com/ampproject/amp-toolbox/tree/master/packages/validator-rules

```
const validatorRules = require('@ampproject/toolbox-validator-rules');

const fetchAllComponents = async () => {
  const rules = await validatorRules.fetch();
  console.log(rules.extensions.map(e => e.name));
}

fetchAllComponents();
```

platform/lib/routers/search.js
```diff
@@ -0,0 +1,268 @@
...
+    if (components.hasOwnProperty(component)) {
+      result.push(component);
+    }
+  }
+  result = result.concat(BUILD_IN_COMPONENTS);
+  result = result.concat(IMPORTANT_INCLUDED_ELEMENTS);
+  return result.sort();
+}
+
+const AUTOSUGGEST_COMPONENTS = getAutosuggestComponents();
```
sebastianbenz: nit: rename to `fetchAutosuggestComponents` to indicate that this involves a network request.

platform/lib/routers/search.js
```diff
@@ -0,0 +1,268 @@
...
+
+examples.get('/search/autosuggest', handleAutosuggestRequest);
+examples.get('/search/do', handleSearchRequest);
+
+function handleAutosuggestRequest(request, response) {
+  const items = ({
+    items: AUTOSUGGEST_COMPONENTS,
```
sebastianbenz: nit: move items into the AUTOSUGGEST_COMPONENTS constant to avoid creating a new object every time.

platform/lib/routers/search.js
```diff
@@ -0,0 +1,268 @@
...
+  });
+  response.json(items);
+}
+
+function hasExample(component) {
+  const filePath = path.join(EXAMPLE_FOLDER_PATH, component + '.html');
+  return fs.existsSync(filePath);
```
sebastianbenz: We should avoid making a FS access. A better approach would be to write a list of samples during the build in samples builder.

platform/lib/routers/search.js
```diff
@@ -0,0 +1,268 @@
...
+}
+
+function getCseItemMetaTagValue(item, metaTag) {
+  // since pagemap has always key:array the metatags dictionary is always the first element in the array
+  if (item.pagemap && item.pagemap.metatags && item.pagemap.metatags.length > 0
```
sebastianbenz: nit: extract `item.pagemap` into a variable

platform/lib/routers/search.js
```diff
@@ -0,0 +1,268 @@
...
+  if (description) {
+    page.description = description;
+  }
+
+  if (page.url) {
+    const componentName = page.url.match(COMPONENT_REFERENCE_DOC_PATTERN)[1];
+    if (hasExample(componentName)) {
+      page.exampleUrl = EXAMPLE_URL_PATH + componentName + '/';
+      page.playgroundUrl = PLAYGROUND_URL + '/?url='
+        + encodeURIComponent(PREVIEW_HOST + page.exampleUrl);
```
sebastianbenz: we should extract this into a helper method.

platform/lib/routers/search.js
```diff
@@ -0,0 +1,268 @@
...
+      // the preview link must not have the locale, but the example doc page has to have it:
+      if (locale != config.getDefaultLocale()) {
+        page.exampleUrl = '/' + encodeURIComponent(locale.toLowerCase()) + page.exampleUrl;
+      }
+    }
+  }
+}
+
+async function handleSearchRequest(request, response, next) {
+  const query = request.query && request.query.q ? request.query.q : '';
```
sebastianbenz: Afaik request.query is always defined

platform/lib/routers/search.js
```diff
@@ -0,0 +1,268 @@
...
+  const page = request.query && request.query.page ? parseInt(request.query.page) : 1;
```
sebastianbenz: nit: extract request.query into a variable

platform/lib/utils/googleSearch.js
```diff
@@ -0,0 +1,60 @@
...
+  console.error('ERROR: Google site search will not be available!',
+      err.message ? err.message : err);
+});
+
+async function search(query, locale, page) {
+  if (!API_KEY) {
+    throw Error('Custom search api key not initialized! Check log for errors on startup.');
+  }
+
+  const startIndex = (page - 1) * 10 + 1;
```
sebastianbenz: extract magic number 10 into a constant

platform/lib/utils/googleSearch.js
```diff
@@ -0,0 +1,60 @@
...
+  }
+  language = encodeURIComponent(language.toLowerCase());
+
+  const url = `${CSE_BASE_URL}?cx=${CSE_ID}&key=${API_KEY}&hl=${language}` +
+      `&lr=lang_${language}&q=${query}&start=${startIndex}`;
+
+  // console.log('Google cse query: ' + url);
+
+  const fetchResponse = await fetch(url);
+  if (fetchResponse.status == 200) {
```
sebastianbenz: can you use `fetchResponse.ok` instead.

sebastianbenz: also, move the success case to the end and early exit if the request fails.

platform/lib/routers/search.js
```diff
@@ -0,0 +1,268 @@
...
+
+const express = require('express');
+const fs = require('fs');
+const path = require('path');
+const config = require('@lib/config.js');
+const project = require('@lib/utils/project.js');
+const googleSearch = require('@lib/utils/googleSearch.js');
+
+// google custom search does not support a page size > 10
+const PAGE_SIZE = 10;
```
sebastianbenz: move this into a constant in googleSearch.js

pages/content/amp-dev/search.html
tharders: I changed the logic, so that the form will display the first page and the amp list will display the rest. Now there is no duplicate request. Although the second page is loaded directly, even when it is not visible. But maybe the problem is, that the amp list is in the viewport when when the results from the first page are not rendered. But I think all in all it is much cleaner now.

pages/content/amp-dev/search.html
tharders: Everything that is in amp-list will be rendered with each page load.
But now I moved it into the form where it makes perfect sense.

pages/content/amp-dev/search.html
tharders: I could remove the event listener completely.
But since I used this tempQuery that was not used directly there would have been no need to use a debounced event.

platform/lib/routers/search.js
tharders: I would prefer to keep it this way as long as we have the components versions in the file system. I think it would be fine, to update the file component version file on deployment.
When there are new components we always need a deployment for the new components documentation anyway.

platform/lib/routers/search.js
tharders: As long as there is no network request we do not need to name it fetch.
But I renamed it buildAuto... because we now have a different structure.

platform/lib/routers/search.js
tharders: I am not quite sure which part you mean

platform/lib/routers/search.js
tharders: I wanted to avoid complicating the build process.
I thought since we just check for existence this is no real problem.
On my local machine it is pretty fast :-)
But of course such a sample list could more flexible and also be helpful in other places.
Since we have the `used_components` property for examples we can even have a list with examples for a component. Search could then simply pick the first one. Maybe we can fill some blanks with this logic. Although the build in components are not listed. Maybe we can simply search for them in the source code.
Let me know what you think about this.

pages/content/amp-dev/search.html
```diff
@@ -17,121 +17,116 @@
...
+        Please try again later.
+      </div>
+
     </form>
-    <div [hidden]="showResults != true" hidden>
-      <p [text]="search.result.totalResults + ' results found'" class="info"></p>
+
+    <div [hidden]="totalResults <= 10" hidden>
 
       <amp-list [height]="(100) * search.result.components.length + (80) * search.result.pages.length"
```
sebastianbenz: use: `is-layout-container` instead of calculating the height https://amp.dev/documentation/components/amp-list/#[is-layout-container]-(optional)

platform/lib/routers/search.js
sebastianbenz: I agree. We need to generate the file during build though (using the approach described above). But that's a different PR.

platform/lib/routers/search.js
```diff
@@ -45,46 +45,33 @@ const DESCRIPTION_META_TAG = 'twitter:description';
...
-function getAutosuggestComponents() {
-  let result = [];
-  const components = require(COMPONENT_VERSIONS_PATH);
-  for (const component in components) {
-    if (components.hasOwnProperty(component)) {
-      result.push(component);
+function buildAutosuggestComponentResult() {
+  let components = [];
+  const componentVersions = require(COMPONENT_VERSIONS_PATH);
+  for (const component in componentVersions) {
```
sebastianbenz: easier: `const components = Object.keys(componentVersions).concat(BUILD_IN_COMPONENTS, IMPORTANT_INCLUDED_ELEMENTS)`

platform/lib/routers/search.js
sebastianbenz: `PLAYGROUND_URL + '/?url='
        + encodeURIComponent(PREVIEW_HOST + page.exampleUrl)` where we construct the playground URL. I think this is used in a few places. Maybe move it into config.js.

platform/lib/routers/search.js
sebastianbenz: Just realized that we have the data already available in: `dist/examples/static/samples.json`.

sebastianbenz:
One thing I noticed, can you please replace the built in components list here: https://github.com/ampproject/amp.dev/blob/future/platform/lib/pipeline/componentReferenceImporter.js#L33 with the new one in AmpConstants. Thanks!

sebil:
@sebastianbenz the frontend of the search is now also available in this PR.

platform/lib/routers/search.js
tharders: Ok, using this data now.
I changed the build so that this file is available when the dev server starts.

platform/lib/routers/search.js
tharders: We have a samples.js utility now

sebastianbenz:
My feedback as discussed in our sync earlier:

* Cannot submit via enter
* Make search results cover whole page
* Don’t show autosuggest straight away => show only after first letter
* Component results take up too much space => reduce components to three, improve prioritization, condense layout, consider different visual treatment
* Search for “Actions Events” -> guide should be visible in first viewport
* Remove “Documentation: ” from component results
* Loadmore overlaps with loading indicator on slow network responses

frontend/templates/views/partials/search.j2
```diff
@@ -0,0 +1,217 @@
...
+{% do doc.amp_dependencies.add('amp-list', '0.1') %}
+{% do doc.amp_dependencies.add('amp-mustache', '0.2', 'template') %}
+
+<amp-state id="query">
+  <script type="application/json">""</script>
+</amp-state>
+
+<div id="searchLayer" class="ap-o-search" tabindex="-1">
+  <div class="ap-o-search-container">
+    <div class="ap-m-search-trigger ap-m-search-trigger-close" on="tap:searchLayer.toggleClass(class='active'),searchTriggerOpen.focus,AMP.setState({ query: '' })" role="button" tabindex="0">
```
sebastianbenz: nit: for readability it's better to pretty print actions

```
on="tap:searchLayer.toggleClass(class='active'),
               searchTriggerOpen.focus,
               AMP.setState({ query: '' }
```

platform/config/search-promoted-pages.json
```diff
@@ -0,0 +1,15 @@
...
+	"default": {
+		"components": [
+			"/content/amp-dev/documentation/components/reference/amp-img.md",
+			"/content/amp-dev/documentation/components/reference/amp-bind.md",
+			"/content/amp-dev/documentation/components/reference/amp-form.md"
+		],
+		"articles": [
+			"/content/amp-dev/documentation/guides-and-tutorials/start/converting/index.md",
+			"/content/amp-dev/documentation/guides-and-tutorials/learn/common_attributes.md",
+			"/content/amp-dev/documentation/guides-and-tutorials/develop/media_iframes_3p/index.md"
```
sebastianbenz: replace with https://amp.dev/documentation/guides-and-tutorials/learn/amp-actions-and-events/

sebil:
@googlebot I consent.

tharders:
> * Loadmore overlaps with loading indicator on slow network responses
https://github.com/ampproject/amphtml/issues/24034
But we were able to fix it with css

sebil:
@sebastianbenz please take a look at the latest changes.
Due to our design QA we've done a lot of fine-tuning.

sebil:
Thanks for the feedback @sebastianbenz , we will take care of it.

tharders:
> * Move "We found xxx results" below search results. There's no real value in the number (as it's skewed anyway) and it takes away important space.

Maybe we should remove the line completely?

> * `Important components` seems to use a different template than the normal search results. We should also have the sample and playground links here.

The json with the highlights is generated by a grow plugin.
The logic to add the example and playground link is in the search.js.
We could serve the highlights dynamically in this case and add the example link then.
But the json is loaded with every page load, so maybe it would be better to replicate the logic for the grow plugin.
I suggest to improve this in a later stage. 

> * We should filter the components list for duplicates (I know it's an index problem, but that's something we can easily fix on our side)

It would only be easy to filter duplicates (in this case different language versions) when they are on the same page. But it would be strange when the page is simply shorter than the others.
I would rather use the page-locale meta tag to filter only english results. For the other languages we cannot use this, since only the translated pages are in the index for those languages.

LongoOfficial:
> the component list still takes away too much space.

we could think about having the component cards instead of the list element:
image

of course, we need to include the links to playground + example. wdyt @sebastianbenz ?


> I still think we should change the height of the search results to 100%. [...] An alternative would be to show the underlying page on both sides.

Makes sense and could be a compromise. But not as strong as the initial approach, as it would be more noisy and distracting in the field of view.

sebastianbenz:
@tharders 

* yes, lets remove the result counting line
* I'd prefer fetching highlights dynamically, this way we can easily replace the results with the last query from a service worker. 
* +1 to filtering by language

@LongoOfficial

I don't think that we need a new UI pattern. I think it's just a matter of fine-tuning the current approach: 

* less margins between the two categories (maybe between each row as well). 
* no result count
* using the whole viewport height

Why would this be more more noisy and distracting in the field of view?

LongoOfficial:
search_result

@sebastianbenz it is less simple than the vertical crop (3 "boxes" instead of 2), but it is definitely a win for the results! plus, we also adjusted font size. wdyt?

sebastianbenz:
@LongoOfficial looks good - let's go for this! Thanks!

jaygray0919:
is a preview of the new search available?

sebastianbenz:
@jaygray0919 not yet. You need to checkout this PR and run it locally.

platform/lib/utils/googleSearch.js
```diff
@@ -36,21 +36,31 @@ credentials.get('GOOGLE_CSE_API_KEY').then((key) => {
...
+  let url = `${CSE_BASE_URL}?cx=${CSE_ID}&key=${API_KEY}&hl=${language}`
+      + `&q=${query}&start=${startIndex}`;
+  if (!noLanguageFilter) {
+    url += `&lr=lang_${language}`;
+  }
+  if (hiddenQuery) {
+    url += `&hq=${hiddenQuery}`;
+  }
+
+  console.log(`CSE url ${url}: `);
```
sebastianbenz: please remove

platform/lib/utils/googleSearch.js
```diff
@@ -36,21 +36,31 @@ credentials.get('GOOGLE_CSE_API_KEY').then((key) => {
...
 
-async function search(query, locale, page) {
+async function search(query, locale, page, options = {}) {
   if (!API_KEY) {
     throw Error('Custom search api key not initialized! Check log for errors on startup.');
   }
 
   const startIndex = (page - 1) * PAGE_SIZE + 1;
   query = encodeURIComponent(query);
+  const hiddenQuery = options && options.hiddenQuery ? encodeURIComponent(options.hiddenQuery) : '';
```
sebastianbenz: I don't think we need the null check here. options cannot be undefined except someone explicitly passes null. In that case we should fail.

platform/lib/utils/googleSearch.js
```diff
@@ -36,21 +36,31 @@ credentials.get('GOOGLE_CSE_API_KEY').then((key) => {
...
+  const noLanguageFilter = options && options.noLanguageFilter;
   let language = locale;
   if (language.length > 2) {
     language = language.substr(0, 2);
   }
   language = encodeURIComponent(language.toLowerCase());
 
-  const url = `${CSE_BASE_URL}?cx=${CSE_ID}&key=${API_KEY}&hl=${language}` +
-      `&lr=lang_${language}&q=${query}&start=${startIndex}`;
+  let url = `${CSE_BASE_URL}?cx=${CSE_ID}&key=${API_KEY}&hl=${language}`
```
sebastianbenz: Suggestion: use `URL` instead

```
const url = new URL(CSE_BASE_URL)
url.searchParams.set('cx', CSE_ID)
...
fetch(url). // fetch can handle URL objects
```

The benefit is that the code is easier to read and you don't need worry about `encodeURIComponent` as it's done automatically.

tharders: Good idea!

tharders:
@sebastianbenz I really hope we have it now :-)
The teaser components now have example + playground links.
We have cache headers.
And the layout is also optimized.
Please have a look.

platform/lib/routers/search.js
```diff
@@ -61,13 +72,26 @@ const AUTOSUGGEST_COMPONENT_RESULT = buildAutosuggestComponentResult();
...
 
 function handleAutosuggestRequest(request, response) {
+  setMaxAge(response, RESPONSE_MAX_AGE.autosuggest);
   response.json(AUTOSUGGEST_COMPONENT_RESULT);
 }
 
+function handleHighlightsRequest(request, response) {
+  const locale = request.query.locale ? request.query.locale : config.getDefaultLocale();
+  const data = require(`${HIGHLIGHTS_FOLDER_PATH}${locale}.json`);
+  for (const page of data['components']) {
```
sebastianbenz: nit: `s/data['components']/data.components`

platform/lib/routers/search.js
```diff
@@ -61,13 +72,26 @@ const AUTOSUGGEST_COMPONENT_RESULT = buildAutosuggestComponentResult();
...
+    addExampleAndPlaygroundLink(page, locale);
+    cleanupTexts(page);
+  }
+  setMaxAge(response, RESPONSE_MAX_AGE.highlights);
+  response.json(data);
```
sebastianbenz: Please add a lazy cache for the highlights response.

PR #2724, opened 2019-08-07
Revert New meta tags (locale, format) that can be used for filtering in Google CSE

matthiasrohmer:
Reverts ampproject/amp.dev#2721. It seems as if the [validator](https://travis-ci.org/ampproject/amp.dev/jobs/568902483#L870) doesn't like them 😅


matthiasrohmer:
Force merged to unblock another PR.

PR #2736, opened 2019-08-08
♻️ Move selector rewriting into amp-optimizer transformer

matthiasrohmer:
Fingers crossed: this moves the selector rewriting which formerly happened during build time into a custom amp-toolbox-optimizer transformer.

### DIFF
--- a/platform/lib/utils/cssTransformer.js
+++ b/platform/lib/utils/cssTransformer.js
@@ -1,0 +1,98 @@
+/**
+ * Copyright 2019 The AMP HTML Authors. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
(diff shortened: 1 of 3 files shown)


platform/lib/utils/cssTransformer.js
```diff
@@ -0,0 +1,101 @@
...
+rcs.selectorLibrary.setExclude(
+    new RegExp('^(?!' + SAFE_CLASS_NAMES.join('|') + ').*$')
+);
+
+/**
+ * A custom transformer for the @ampproject/toolbox-optimizer rewriting
+ * CSS selectors to save bytes in <style amp-custom>.
+ *
+ * It uses the advantage of the DOM used by the optimizer being compatible to
+ * the one used by posthtml
```
sebastianbenz: this is no longer true, no?

matthiasrohmer: Hah, thanks 😁

PR #2747, opened 2019-08-12
 only set Vary: AMP-Cache-Transform for html requests

sebastianbenz:


.gitignore
```diff
@@ -20,6 +20,7 @@ pages/podspec.yaml
 pages/css
 pages/icons
 pages/macros
+pages/.grow
```
matthiasrohmer: Finally, thanks 😅

PR #2750, opened 2019-08-12
🐛 Fix amp-access login

### LINKED ISSUE #2526: Login button does not work
https://amp.dev/documentation/examples/e-commerce/checkout_flow/preview/?format=websites

lluerich:
Since the ```/login.html``` in the ```/examples/static/```-folder was not reachable, I moved it to the same folder as the API itself.

@sebastianbenz I hope this is ok. If you come up with a better solution, let me know. 😃

Fixes #2526


examples/api/amp-access/amp-access.js
```diff
@@ -66,7 +66,7 @@ function handleAuthorization(request, response) {
 
 function handleLogin(request, response) {
   const returnUrl = request.query.return;
-  const filePath = utils.project.absolute('/examples/static/samples/files/login.html');
+  const filePath = utils.project.absolute('/examples/api/amp-access/login.html');
```
sebastianbenz: you can use relative paths instead `./login.html`.

examples/api/amp-access/amp-access.js
```diff
@@ -38,10 +38,10 @@ const POWER_USERS = {
 };
 const EXPIRATION_DATE = 24*60*60*1000; // 1 day in ms
```
sebastianbenz: please rename this file from amp-access.js to index.js.

examples/api/amp-access/amp-access.js
lluerich: Do you mean by using ```utils.project.relative(...``` or just the string ```./login.html```?

examples/api/amp-access/amp-access.js
lluerich: Thanks for pointing that out. PTAL.

examples/api/amp-access/amp-access.js
lluerich: ```./login.html``` only works, if you set up ```nunjucks.configure(__dirname)``` beforehand.
Was this what you were thinking of?
Either way, I will move the ```filePath``` into a constant.

sebastianbenz: ah ok, `path.join(__dirname, 'login.html')` would've been my next suggestion.

lluerich: works for me, and 1 less line of code 👍

Issue #2243, opened 2019-06-03
Create a shortcode for internal links


sebastianbenz:
Would be great to have a simple shortcode for internal links. Currently, I always have to copy paste the code from a different page. 

//cc @matthiasrohmer

matthiasrohmer:
So basically a wrapper around `{{ g.doc('content/amp-dev/...', locale=doc.locale) }}`? 😉 Did you have a specific syntax in mind?

I could imagine `[link example="amp-accordion"]` and `[link component="amp-web-push"]` for example. Guides would be a bit more verbose though as you'd still need to know their bucket and file type: `[link guide="develop/interactivity/get-familiar.md"]`.

sebastianbenz:
It's even worse: `{{g.doc('/content/amp-dev/documentation/guides-and-tutorials/learn/spec/amp-boilerplate.md', locale=doc.locale).url.path}}`. Or is `.url.path` not needed?

It'd be perfect if we could stay with the markdown link syntax and automatically rewrite relative links to the grow syntax.

PR #2754, opened 2019-08-13
Rewrite pod internal links in <a href to site links

tharders:
A new grow extension will rewrite pod internal links to site links.
Currently only links in <a href tags are handled.

This is the first step for issue #2243.

The second step would be to change all the ugly links in the content like
`{{g.doc('/content/amp-dev/documentation/guides-and-tutorials/learn/spec/amp-boilerplate.md', locale=doc.locale).url.path}}` to relative links like `../learn/spec/amp-boilerplate.md` or pod links like `/content/amp-dev/documentation/guides-and-tutorials/learn/spec/amp-boilerplate.md`
The first would be the best, since those links would be valid in markdown.
But the second could be done with a simple search and replace over all documents.
I would leave the links in the templates using g.doc for now.

It looks to me that the growReferenceChecker does not check regular markdown links, but only the g.doc syntax although a comment in the checker states otherwise.
When we change all the links, this has to be investigated and fixed if needed.

### DIFF
--- a/pages/extensions/internal_links/pod_internal_link_rewriter.py
+++ b/pages/extensions/internal_links/pod_internal_link_rewriter.py
@@ -1,0 +1,87 @@
+import os
+import re
+import traceback
+from grow.pods.pods import Pod
+from grow.documents.document import Document
+
+# find all <a href without a : in the link and not ending with /
+LINK_PATTERN = re.compile(r'<a\s+(?:[^>]+\s)?href\s*=\s*"((?:[^"#?](?!:))*?[^"/])((?:\?[^"]*)?(?:#[^"]*)?)"',
+                          re.IGNORECASE)
+
+
+class PodInternalLinkRewriter(object):
+
+  pod = None  # type: Pod
+
+  def __init__(self, doc):
+    """
+    :type doc: Document
+    """
+    self.doc = doc
+    self.pod = doc.pod
+
+  def rewrite_pod_internal_links(self, content):
+    """
+    :type content: str
+    """
+    try:
+      pos = 0
+      result = ''
+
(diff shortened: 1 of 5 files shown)


matthiasrohmer:
>It looks to me that the growReferenceChecker does not check regular markdown links, but only the g.doc syntax although a comment in the checker states otherwise.

Indeed 😅 That's my specialty: promising features in a comment the implementation then doesn't keep up to. Sorry.

Checking for `g.doc` was sufficient when I was implementing the checker as only those were blocking the build if corrupt.

tharders:
It is likely that there is also some other link conversion logic that builds those g.doc links (for example on import of the amphtml docs) that can be changed after we have this new extension.

pages/extensions/internal_links/pod_internal_link_rewriter.py
```diff
@@ -0,0 +1,87 @@
...
+    :type internal_link: str
+    """
+    try:
+      internal_path = internal_link
+
+      if not internal_path.startswith('/'):
+        internal_path = os.path.abspath(os.path.join(
+          PodInternalLinkRewriter.get_folder(self.doc.pod_path), internal_path))
+
+      target_doc = self.pod.get_doc(internal_path, self.doc.locale)
```
matthiasrohmer: If you've got time left, can you do me a favor? `get_doc` constructs a whole `Document` object on every call including parsing of a document's frontmatter and body as far as I know - maybe we could also speed up build times here by caching resolved URLs? Or would you prefer not to?

tharders: Done, please check.

PR #2757, opened 2019-08-13
♻️Migrate amp-analytics example backend to amp.dev

lluerich:
Migrated the amp-analytics example backend to amp.dev.
It updates the example [here](https://amp.dev/documentation/examples/components/amp-analytics/?format=websites).

### DIFF
--- a/examples/source/1.components/amp-analytics/embed.html
+++ b/examples/source/1.components/amp-analytics/embed.html
@@ -1,0 +1,62 @@
+<!doctype html>
+<html>
+  <head>
+    <meta charset="utf-8">
+    <style media="screen" type="text/css">
+      body {
+        background: white;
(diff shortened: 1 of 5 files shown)


examples/source/1.components/amp-analytics/embed.html
```diff
@@ -0,0 +1,62 @@
+<!doctype html>
+<html>
+  <head>
+    <meta charset="utf-8">
+    <style media="screen" type="text/css">
+      body {
+        background: white;
+        font-family: Roboto,Helvetica,Arial,sans-serif;
```
sebastianbenz: We're no longer using Robots on the site. Please replace it with our current fonts.

lluerich: Thanks, PTAL.

Issue #2756, opened 2019-08-13
Improve new overview pages


sebastianbenz:
# Desktop

1. I don't understand the color coding on the card. What does the white space mean?

<img width="1141" alt="Screen Shot 2019-08-13 at 15 31 29" src="https://user-images.githubusercontent.com/380472/62945762-ad0a4980-bddf-11e9-9281-b58e75e77459.png">

2. Too much space between yellow dot and title

<img width="1078" alt="Screen Shot 2019-08-13 at 15 29 16" src="https://user-images.githubusercontent.com/380472/62945770-aed40d00-bddf-11e9-8219-58510920a177.png">

3. Open in playground should open the playground in the currently active format 

<img width="860" alt="Screen Shot 2019-08-13 at 15 29 05" src="https://user-images.githubusercontent.com/380472/62945771-af6ca380-bddf-11e9-969c-ccd3dbf5c43a.png">

4. "None" example listed below
5. It's not clear that these are sample highlights. My first impression would be that these are all samples on the page. Change to: "Featured examples"
6. There's a lot of space below "Learn AMP by Example". 

<img width="1532" alt="Screen Shot 2019-08-13 at 15 28 33" src="https://user-images.githubusercontent.com/380472/62945774-af6ca380-bddf-11e9-90ca-9356407fe61c.png">

7. Same feedback as before. Mixing cards and simple sample listing looks super weird. We should get rid of the cards.

<img width="352" alt="Screen Shot 2019-08-13 at 15 28 15" src="https://user-images.githubusercontent.com/380472/62945775-af6ca380-bddf-11e9-9200-eae0792d4888.png">

8. I find the heading "Categories" a bit confusing here. Either "Example categories" or something like: "Filter by category". It's also not clear that these are filters for the samples below (same is true for components and guides pages).  

<img width="420" alt="Screen Shot 2019-08-13 at 15 27 56" src="https://user-images.githubusercontent.com/380472/62945776-b0053a00-bddf-11e9-9c66-ff8f01198f72.png">

9. The last item should be "show more" that jumps to the matching category below.

<img width="565" alt="Screen Shot 2019-08-13 at 15 27 44" src="https://user-images.githubusercontent.com/380472/62945779-b1366700-bddf-11e9-819a-35ee6668c317.png">

10. Typography looks. bad of Categories heading vs sub headings. //cc longo

[...]

PR #2776, opened 2019-08-16
🎨 Improve new overview pages

matthiasrohmer:
This addresses some of the comments in #2756.

- Non-existing linked documents don't show up as `None` anymore
- Example categories are now properly cased and divided
- Featured examples got a headline
- "Categories" has been rephrased to "Filter by category" on example and guides overview
- Playground teaser on filtered example overview opens the playground in matching format
- Format tags are shown in consistent order (websites, stories, ads, email) now. A wrong format key of `emails` so far causing a blank tag is additionally catched gracefully now
- Page title of filtered examples overview shares its format now

### DIFF
--- a/frontend/templates/views/partials/teaser-tags.j2
+++ b/frontend/templates/views/partials/teaser-tags.j2
@@ -1,0 +1,16 @@
+{% do doc.styles.addCssFile('/css/components/molecules/tag.css') %}
+
+<div class="ap-m-teaser-tags">
+  {% if 'websites' in formats %}
+  <div class="ap-m-tag ap-m-tag-websites"></div>
+  {% endif %}
+  {% if 'stories' in formats %}
+  <div class="ap-m-tag ap-m-tag-stories"></div>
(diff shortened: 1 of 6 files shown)


frontend/templates/views/partials/teaser-tags.j2
```diff
@@ -0,0 +1,16 @@
...
+  {% if 'websites' in formats %}
+  <div class="ap-m-tag ap-m-tag-websites"></div>
+  {% endif %}
+  {% if 'stories' in formats %}
+  <div class="ap-m-tag ap-m-tag-stories"></div>
+  {% endif %}
+  {% if 'ads' in formats %}
+  <div class="ap-m-tag ap-m-tag-ads"></div>
+  {% endif %}
+  {% if 'email' in formats or 'emails' in formats %}
```
sebastianbenz: This is the wrong place for this: `'email' in formats or 'emails' in formats`. We should normalize this when we load the frontmatter.

PR #2777, opened 2019-08-17
[Feature] Support multiple components/tags per extension

pbakaus:
**Important: Please merge #2770 before merging this one!**

This PR adds support for AMPHTML extensions that themselves have multiple custom elements (or tags) they expose to the user. The `amp-story` component is the most prominent example here.

# How does this change things?

Previously, our reference doc importer was only able to import one markdown file per extension, so the `amp-story` documentation needed to be self-contained in one massive file.

With this PR, the markdown files in the `amp-story` extension can be split into multiple files, which @newmuis and the gang have already done on master, it so seems: https://github.com/ampproject/amphtml/tree/master/extensions/amp-story

This turned out to be not an entirely trivial problem, since in the same folder, there are other markdown files that *don't* map to a tag, which we don't want to include. For that reason, **I'm parsing the extensions' protoascii file to search for tags that actually exist** (🤯).

# What does it look like?

Basically, like it should:

image

Enjoy.

/cc @CrystalOnScript


pbakaus:
@sebastianbenz I think you still got to approve this one if you think it looks good :) otherwise this should be ready to merge.

pbakaus:
weeeeeee! thanks :)

PR #2780, opened 2019-08-19
Fix links and reference checker

tharders:
The reference checker will now run for all documents (not only documentation)
It now works with normal links (not just the g.doc links).
It is fixed that an error is thrown when links had multiple matches.

Findings of the reference checker where corrected.


platform/lib/tools/growReferenceChecker.js
```diff
@@ -27,9 +27,9 @@ const POD_BASE_PATH = path.join(__dirname, '../../../pages/');
 // Which documents to check for broken references
 const PAGES_SRC = POD_BASE_PATH + 'content/amp-dev/documentation/**/*.{md,html}';
 // The location to search for documents in
-const PAGES_BASE_PATH = POD_BASE_PATH + 'content/amp-dev/documentation';
+const PAGES_BASE_PATH = POD_BASE_PATH + 'content/amp-dev';
 // The pattern used by Grow to make up references
```
sebastianbenz: can you update the comment to reflect the lastest regex

tharders:
@sebastianbenz Thinking about the comment I found that <a href links where missing.
I added support and then found problems with the relative link changing in the importer.
Both is fixed now. Please check again.

platform/lib/tools/growReferenceChecker.js
```diff
@@ -25,13 +25,29 @@ const fs = require('fs');
...
 // The location to search for documents in
 const PAGES_BASE_PATH = POD_BASE_PATH + 'content/amp-dev';
-// The pattern used by Grow to make up references
-const REFERENCE_PATTERN = /\[[^\]]+\]\(([^:\)\{?#]+)(?:[?#][^\)]*)?\)|g.doc\('(.*?)'/g;
-// Contains manual hints for double filenames etc.
+// The pattern to find links in markdown and html
+// It also matches source code blocks to skip these
+const REFERENCE_PATTERN = new RegExp(
+    // skip sourcecode block in markdown:
+    /^```[\s\S]*?```/.source
```
sebastianbenz: won't this fail for "```html"?

tharders: No, the "[\s\S]" will match any character until the ending ``` is found.

sebastianbenz: ah cool -  thanks

PR #2786, opened 2019-08-19
🎨 Improving the info-button of the level-toggle

### LINKED ISSUE #2775: Improving the courses toggle help button
Thanks for finishing up and pushing out #2636 . It looks great!

I wasn't able to preview this on my machine because of the bug ultimately fixed by @matthiasrohmer (and thanks for that!) So I'm just out the final experience now, in production. I like it! I raised one bug report, but otherwise I like it. My only remaining request is about the help button on the toggle.

Right now, you have to click the question mark to see the help text popup. Clicking the popup or outside the popup doesn't close it. You have to click the question mark again. Plus, once you click the question mark, it's active, and it gets a faint outline, in my Chrome, at least.

Screen Shot 2019-08-16 at 11 41 34 AM

Can we remove the outline, and can we make the popup vanish if it's clicked or if any element near it is clicked?

Would it be confusing to make this tooltip pop up on hover also? I think that's more natural for desktop users, though of course it won't work on mobile :)

dlemm:
Fixes #2775 
- Add the possibility to hover on the info-button in order to show the info-text. 
- Add the possibility to close the info-text by clicking somewhere on the page.


morsssss:
Thanks for this! The CSS looks promising to me :) I just can't build, once again...

```
ImportError: No module named document
[93983] Failed to execute script grow
```

I get that in the `future` branch as well as this one 😭

frontend/scss/components/molecules/level-toggle.scss
```diff
@@ -138,7 +138,31 @@ $borderSize: 5px;
...
+    &:focus {
+      outline: none;
+    }
+    &:hover {
+      .#{molecule('tip')} {
+        display: block;
+      }
+    }
+  }
+
```
tharders: The info button is really nice and works well with the mouse over.
Can we extract the info-button, so that we have a reusable component?
This would also reduce the css size somewhat.

PR #2795, opened 2019-08-20
Better vary filter

sebastianbenz:
Set vary filter based on path and mimetype.

### DIFF
--- a/platform/lib/routers/packager.js
+++ b/platform/lib/routers/packager.js
@@ -18,6 +18,7 @@
 
 const HttpProxy = require('http-proxy');
 const config = require('@lib/config');
+const mime = require('mime-types');
 
 const proxyOptions = {
   target: config.hosts.packager.base,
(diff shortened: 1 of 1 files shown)


platform/lib/routers/packager.js
```diff
@@ -52,11 +53,14 @@ const packager = (request, response, next) => {
     next();
     return;
   }
-  // We'll only set Vary: AMP-Cache-Transform for html or SXG requests
-  const acceptHeader = request.header('accept');
-  if (acceptHeader &&
-        (!acceptHeader.includes('text/html') &&
-         !acceptHeader.includes('application/signed-exchange'))) {
```
tharders: As far as I understand, the check for the accept header is mandatory here and should not be removed.
Or did I miss something here? I do not know much about this whole topic.

sebastianbenz: You're right. It's not mandatory, but helps our CDN to cache more efficiently. This change to please toolbox-linter ([context](https://github.com/ampproject/amp-toolbox/issues/456)). 

However, thinking more about this, the correct approach would be to check that both `text/html` and `application/signed-exchange` are present at the **same** time. Only if both are present we need to vary on the `amp-cache-transform` header (otherwise, `vary: accept` is already sufficient for the CDN to differentiate the responses).

sebastianbenz: Based on the discussing [here](https://github.com/ampproject/amp-toolbox/issues/456) and Devin's feedback we should always set the vary header as it's implemented in the PR.

PR #2815, opened 2019-08-23
Generate component versions at build time

sebastianbenz:
* update playground to no longer use the ABE backend (finally!)
* invoke generator before build
* update grow extensions
* update search integration


gulpfile.js/build.js
```diff
@@ -227,7 +243,14 @@ function buildPrepare(done) {
       lint.lintNode,
       // Build playground and boilerplate that early in the flow as they are
       // fairly quick to build and would be annoying to eventually fail downstream
-      gulp.parallel(buildPlayground, buildBoilerplate, buildSamples, importAll, zipTemplates),
+      gulp.parallel(
+          buildComponentVersions,
```
tharders: buildComponentVersions also has to run when the develop server is started

sebastianbenz: `gulp bootstrap` should be run after checkout which will add the file

PR #2841, opened 2019-08-27
analytics track search trigger

sebastianbenz:


sebastianbenz:
Why do you think that `get-courses` is no longer used?

PR #2843, opened 2019-08-27
Fix css limitation errors on some example pages

sebil:
The custom styles of some examples causing css limitation errors since the new search is there.
E.g. here: https://amp.dev/documentation/examples/e-commerce/product_page/?format=websites

This PR removes unused css and adds more classes to the cssTransformer to save more bytes.

Overall savings: 2.2 KB


lluerich:
@googlebot I consent.

Issue #2802, opened 2019-08-21
The prod sitemap_grow.xml contains the test pages


tharders:
The [grow sitemap](https://amp.dev/sitemap_grow.xml) contains the test pages like https://amp.dev/tests/examples/default_settings/ although they are not deployed on prod.

We use a filter to exclude those pages, but this does not exclude those pages from the sitemap. Pleas note that the build also uses a filter to generate only a specific locale, where we need to have a sitemap for "all" pages.

sebastianbenz:
This happens here: https://github.com/ampproject/amp.dev/blob/future/frontend/templates/layouts/sitemap.xml#L3

Do we have a generic way to iterate over all prod pages?

PR #2848, opened 2019-08-28
🔍 Do not list sample previews in sitemap

### LINKED ISSUE #2847: Preview pages are listed in sitemap
The sample preview pages are listed in the sitemap, but include the noindex tag:

Example: https://amp.dev/fr/documentation/examples/user-consent/story_user_consent/preview/

We shouldn't list these in the sitemap as well.

matthiasrohmer:
Fixes #2847 and #2802.

### DIFF
--- a/pages/content/tests/examples/default_settings.md
+++ b/pages/content/tests/examples/default_settings.md
@@ -2,6 +2,8 @@
 $title: Default settings
 $description: Example with default settings (preview=none, playground=true)
 $order: 1
+$sitemap:
+  enabled: False
 ---
 
 Start
--- a/pages/content/tests/examples/example_with_comments.md
+++ b/pages/content/tests/examples/example_with_comments.md
@@ -1,6 +1,8 @@
 ---
 $title: Example with HTML comments
 $order: 10
+$sitemap:
+  enabled: False
 ---
 
 Start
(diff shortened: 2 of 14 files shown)


frontend/templates/layouts/sitemap.xml
```diff
@@ -1,11 +1,13 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <urlset xmlns="http://www.sitemaps.org/schemas/sitemap/0.9">
   {% for doc in docs|sort(attribute='url') if not doc.sitemap.enabled is sameas false %}
+  {% if not '/documentation/examples/previews' in doc.pod_path %}
```
sebastianbenz: Can we not use the hard-coded path for this?

matthiasrohmer: That's just as we are doing it in structured-data.j2: https://github.com/ampproject/amp.dev/blob/5a7b0f89a95078499088006d1df80aedd7e57852/frontend/templates/views/partials/structured-data.j2#L29

Alternatively we could check for the template by using `doc.view == '/views/examples/preview.j2'`. What do you like better?

sebastianbenz: Yup, I don't like it there either :-). Can we instead set a flag in the front matter (e.g. in the samples builder)?

matthiasrohmer: Sure!

tharders: @matthiasrohmer This can then be used to hide the tests from the sitemap as well https://github.com/ampproject/amp.dev/issues/2802

PR #2860, opened 2019-08-30
🐛 Only try to access teaser if it exists

matthiasrohmer:
/cc @dlemm: you can only access frontmatter keys that exist. In comparison to other template languages jinja2 doesn't fail silently 🙂


matthiasrohmer:
You mean in 

https://github.com/ampproject/amp.dev/blob/a7eb1c5986c2d695573ac038c67be6aa879631b9/frontend/templates/views/partials/teaser.j2#L48

This is safe-guarded by 

https://github.com/ampproject/amp.dev/blob/a7eb1c5986c2d695573ac038c67be6aa879631b9/frontend/templates/views/partials/teaser.j2#L15

PR #2896, opened 2019-09-05
💬 Add success stories for kompas and asahi 

robinvanopstal:


sebastianbenz:
Can you please make sure that the images are optimally compressed. Thanks!

PR #2903, opened 2019-09-06
Don't parse the same sample twice

### LINKED ISSUE #2901: All sample styles are broken
For example: https://amp.dev/documentation/examples/interactivity-dynamic-content/favorite_button/

The `.favorite-button` selector is no longer present in the CSS.

sebastianbenz:
Fixes #2901

### DIFF
--- a/platform/lib/build/samplesBuilder.js
+++ b/platform/lib/build/samplesBuilder.js
@@ -80,6 +80,7 @@ class SamplesBuilder {
     this._cache = {};
     // Holds all relevant sample informations after samplew have been parsed
     this._sitemap = {};
+		this._parsedSamplesCache = new Map();
   }
 
   /**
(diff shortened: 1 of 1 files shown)


fstanis:
I don't think this is a good idea:

-   `abe.parseSample` only uses the path to determine the title. Actual content comes from the last parameter (`sample.contents.toString()`).
-   The sample may have changed on second processing if there are `@format` directives, which might have removed parts of it. Therefore, the "first pass" shouldn't be cached.

Ideally, there could be a `parseSample` variant that only extracts the metadata (this is why we need to run `parseSample` twice right now, to determine which samples (not) to transform. But I suspect building functionality that does it is not worth the effort.

The key question is why transform of a web -> web sample (which should be almost no-op, other than `@formats`) causes the styles to be lost. I'm investigating into that now.

sebastianbenz:
Closed in favor of @fstanis solution

PR #2927, opened 2019-09-09
Allow amp-form example endpoints to support application/x-www-form-urlencoded

fstanis:


examples/api/amp-form.js
```diff
@@ -21,6 +21,9 @@ const upload = multer();
 
 // eslint-disable-next-line new-cap
 const examples = express.Router();
+examples.use(express.urlencoded({
+  extended: false,
```
sebastianbenz: curios: is there a specific reason why you set `extended: false`?

fstanis: Not for a very good reason, but:

1. I saw that's how it was done [here](https://github.com/ampproject/amp.dev/blob/future/examples/source/1.components/amp-web-push/api.js) (but now I see that we also have it set to true in some places).

2. I figured if `extended` (nested objects) is not explicitly needed, it shouldn't be on.

Issue #2924, opened 2019-09-09
Add search to blog


sebastianbenz:
We should integrate the current search implementation into the blog as well. 

@matthiasrohmer do you know a good way to share the search markup as a snippet?

matthiasrohmer:
I'd go for the same approach as with header and footer: https://github.com/ampproject/amp.dev/tree/future/pages/content/shared. See #1800 for history 🙂

matthiasrohmer:
Now part of #3152.

PR #2934, opened 2019-09-10
Share search markup for search

sebastianbenz:
See #2924

### DIFF
--- a/pages/content/shared/header.html
+++ b/pages/content/shared/header.html
@@ -10,3 +10,4 @@ $sitemap:
 
 {% include 'views/partials/header.j2' %}
 {% include 'views/partials/burger-menu.j2' %}
+{% include 'views/partials/search.j2' %}


sebastianbenz:
@westonruter do you need anything else for the blog integration?

westonruter:
OK, so right now the blog is doing:

```css
#searchTriggerOpen,
.ap-o-burger-menu-label {
	display: none;
}
```

This styling could be removed with this deployed?

sebastianbenz:
@westonruter Is there a way to test this before deploying?

westonruter:
@sebastianbenz OK, so apparently this was deployed? So this can be tested now… and it breaks:

> image

The issue appears to be that the fragment being pulled in is using optimized AMP, which the AMP plugin does not allow as input, causing them to be stripped out by the sanitizer:

> image

It's currently not broken on https://blog.amp.dev, but it is broken on a https://test-amphtml-blog.pantheonsite.io/ presumably because the cached [header.html](https://amp.dev/shared/header.html) hasn't expired yet, but within a week it will be broken once the transient expires.

I'll preemptively add `.ap-o-search-container { display:none }` to blog.amp.dev to prevent the breakage from showing up.

westonruter:
For some reason the `[src]` attribute on `amp-list` is getting dropped when the https://amp.dev/shared/header/ fragment is imported, and this is causing what @mattludwig pointed out:

image

westonruter:
The reason for the AMP validation error is that the validator is not properly accounting for `data-amp-bind-src` as being an alias for `[src]`. I've filed a bug: https://github.com/ampproject/amphtml/issues/24661

westonruter:
For a quick fix, I've just stripped out the `div#searchLayer` until this is resolved.

westonruter:
@sebastianbenz What CSS specifically does this new search component require? It seems like `frontend/scss/components/organisms/search.scss`. Anything else?

matthiasrohmer:
@sebil, can you confirm adding `/css/components/organisms/search.css` to the blog's CSS is enough to successfully render search for the blog?

sebil:
@matthiasrohmer there is also `/components/molecules/search-trigger.css` which is included via the header markup. Besides that it should be enough, yes.

PR #2947, opened 2019-09-12
Generated search highlights are added to artifacts on travis

### LINKED ISSUE #2819: Search promoted JSON files are not generated during build
See temporary fix: https://github.com/ampproject/amp.dev/commit/8f272c26edd58b5255d7a0c197d0636e4e62db08

tharders:
This will fix #2819 

Please note that the files imported from amphtml/spec do not specify a description or teaser text.
Since we highlight one of those pages...
https://github.com/ampproject/amphtml/blob/master/spec/amp-actions-and-events.md / https://amp.dev/documentation/guides-and-tutorials/learn/amp-actions-and-events/?format=websites
...there would be no description for this entry, only a title. (Currently the description was added manually)

If this is a problem we should enhance the importer to auto generate a description.
We could use the first sentence  of the first paragraph (Should be the first block that starts with text). But of course this could result in some descriptions that are not fitting. Those descriptions would also be used to populate the meta tags for twitter (which is also used by the search).

Another option would be to use the first sentence only in the highlight list grow plugin.
Those pages would then still have no twitter description. (Search will use the extract that google search provides for the description)


tharders:
Thinking about it, we should at least extract the first sentence for the search highlights.
I will extend the grow extension with this PR.

tharders:
@sebastianbenz I have changed the grow extension to extract the first sentence of the first paragraph. So now the description is no longer empty.

pages/extensions/extract_highlights_info/page_info_collector.py
```diff
@@ -1,9 +1,17 @@
...
 import os
+import re
 import json
 from grow.pods.pods import Pod
 from grow.documents.document import Document
 
+FIRST_SENTENCE_PATTERN = re.compile(
+    # We also match HTML comments and markup [tag]...[/tag] blocks to skip them
+    r'<!--.*?-->|^\[(\w+)(?:[:\s][^\]]+)\].*?\[/\1\]|(^\w[^.\n]*(?:\n\w[^.\n]*)*(?:\.|$))',
+    re.MULTILINE | re.DOTALL)
```
sebastianbenz: This seems a bit fragile:

* What about other markdown features such as : `**Bold** _italic_ ...`? 
* What if a sentence ends with `!` or `?`? 
* What about other languages?

tharders: Very good points!

I changed the logic, so that we first create the html for the markup, remove tags and blocks we do not want and then search the first block.

The first sentence can now end with . or ! or ?

Also \w will match any unicode character, so that also arabic sentences are found.

tharders:
@sebastianbenz The description extraction logic is updated. Seems to be fine now.

Issue #2683, opened 2019-08-01
Hide translation warning on mobile


sebastianbenz:
It's not really actionable on mobile and takes away important screen space.

sebastianbenz:
<img width="439" alt="Screen Shot 2019-08-01 at 21 35 04" src="https://user-images.githubusercontent.com/380472/62321985-44989f80-b4a4-11e9-91be-516fc2d35d07.png">

tharders:
Fixed with #2957

PR #2957, opened 2019-09-13
🚸 Improve mobile UX

### LINKED ISSUE #2699: Homepage: hide playground on small devices
It's not very usable, so we should hide the whole section for devices `<=320px`

Screen Shot 2019-08-05 at 10 52 55

matthiasrohmer:
Fixes #2699 and #2683.


sebastianbenz:
Thanks!

PR #2958, opened 2019-09-13
🚸 Make Get Started links on home point to Docs/Get Started

matthiasrohmer:
Not a fix for a specific issue but something I felt to be better UX: the two Get Started links on the homepage point to the Guides & Tutorials overview page while we introduced the new Get Started page for the documentation which might be a better fit for those links.


sebastianbenz:
Makes sense

PR #2960, opened 2019-09-13
🚸 Warn if document doesn't match currently selected format

### LINKED ISSUE #2553: Filtering doesn't prevent pages from being shown for other formats
Examples:

*   https://amp.dev/documentation/guides-and-tutorials/develop/style_and_layout/custom_fonts/?format=email
*   https://amp.dev/documentation/guides-and-tutorials/learn/experimental/?format=ads

These pages are invisible in the menu on the left, which is good, but they can still be opened when another format is selected. This seems confusing, as it e.g. implies custom fonts are supported in emails, which is not the case.

matthiasrohmer:
Fixes #2553.

Bildschirmfoto 2019-09-13 um 12 13 24

Bildschirmfoto 2019-09-13 um 12 12 49

### DIFF
--- a/frontend/templates/views/partials/content.j2
+++ b/frontend/templates/views/partials/content.j2
@@ -10,8 +10,24 @@
   <h1>{{ doc.title }}</h1>
   {% endif %}
 
+  {% set formats = ['ads', 'email', 'stories', 'websites'] %}
+  {% if (doc.formats|sort or formats) != formats %}
+  {% do doc.styles.addCssFile('css/components/molecules/tip.css') %}
+  [% set availableFormats = {{ doc.formats|tojson }} %]
+  [% if not availableFormats.includes(format) %]
+  <div class="ap-m-tip ap-m-tip-important">
+    <div class="ap-m-tip-content">
+      {% do doc.icons.useIcon('icons/warning.svg') %}
+      <div class="ap-a-ico ap-m-tip-icon">
+        <svg><use xmlns:xlink="http://www.w3.org/1999/xlink" xlink:href="#warning"></use></svg>
+      </div>
+      <p>{{ _('Important: this document is not valid for your currently selected format') }} <b>[= format =]</b>!</p>
+    </div>
+  </div>
+  [% endif %]
+  {% endif %}
 
-  {# Add the level toggle button, if the document has a level: true in the frontmatter #}
+  {# Adds the level toggle button if leveled: true is set in the frontmatter #}
   {% if doc.leveled %}
   {% include '/views/partials/level-toggle.j2' %}
   {% endif %}


frontend/templates/views/partials/content.j2
```diff
@@ -8,10 +8,23 @@
 
   {% if doc.title and doc.view in ['/views/detail/docs-detail.j2', ] %}
   <h1>{{ doc.title }}</h1>
-  {% endif %}
 
+  {% do doc.styles.addCssFile('css/components/molecules/tip.css') %}
```
tharders: This css should only be included, if the tip is shown.
Should be possible to move this line into the if block.

pages/content/amp-dev/documentation/guides-and-tutorials/optimize-measure/amp-managing-user-state.md
```diff
@@ -490,7 +490,7 @@ We recommend validating the authenticity of query parameter values by using the
 </ul>
 <p>To do this on your non-AMP page, include the following JavaScript, which will remove all query parameters from the URL:</p>
 <pre>
-var href = location.href.replace(/\?[^#]+/, '');
+var href = location.href.replace(/\?[^{{'[% raw %]'}}#]{{'{% endraw %}'}}+/, '');
```
tharders: Those changes should not be part of this PR.

frontend/templates/views/partials/content.j2
matthiasrohmer: Hm, thought about this as well: wouldn't help. The `[% if %]` only get's evaluated server-side when the CSS has long been combined.

But you're right currently this PR would add the tip styling to *every* docs detail page.

pages/content/amp-dev/documentation/guides-and-tutorials/optimize-measure/amp-managing-user-state.md
matthiasrohmer: Thanks!

frontend/templates/views/partials/content.j2
tharders: Ok, I did not pay attention to square or curly brackets at first :-o

The only thing we could do is to only include the css if not all formats are supported.
But most of the guides are not for all formats, so that would not help much.

Another thing is, that the changes do not work for components since they use another template (component-detail.j2). I think that should be added.

I would not add the layer for examples, since we cannot specify multiple formats, but lots of examples are in general also correct for other formats (see the components section where every example is only visible for websites)

matthiasrohmer:
@tharders, please have another look. I safeguarded with

```jinja2
{% set formats = ['ads', 'email', 'stories', 'websites'] %}
  {% if (doc.formats|sort or formats) != formats %}
```

So at least documents allowing all formats do not get the tip styling included. 

It already worked for the component details page as they are all using the same `content.j2` partial.

Issue #2963, opened 2019-09-13
Version toggle is not shown for translations


tharders:
## Problem

Currently the importer adds the version information to all imported components.
* List of all versions
* File version
* A flag that shows the file is the latest version
* A special path mapping if the file is the latest version

That is fine as long as all documentation is imported.
But when we have translated documentation the version information is not updated!

As a result the version information is not available for the current translated files:
https://amp.dev/es/documentation/components/amp-carousel-v0.1/?format=websites

## Possible solution

It would be best, if the single files would not contain more than the file version.
In that case the translated files would not have to change when a new version is available.

A possible solution might be to put each component documentation into their own folder and set the version information in the `_blueprint.yaml`

The folder structure could be as follows:
![image](https://user-images.githubusercontent.com/6458408/64859517-bed34c80-d62a-11e9-843a-3ad312d41d18.png)

The `_blueprint.yaml` could look like this:
```yaml
$path: /documentation/components/{base}.html
$localization:
  path: /{locale}/documentation/components/{base}.html
$view: /views/detail/component-detail.j2
versions:
  - '0.2'
  - '0.1'
latest_version: '0.2'
```

In order to server the latest version with the old URL, we could create a nearly empty index.md file that would use a specific template to simply render the latest version.
Links from other files to that component should then point to the index.md when no specific version is chosen.

tharders:
With the proposed solution the latest version would be available under two links.
At least we should ensure that the canonical link of the latest version points to the index.

Also we must check how https://github.com/ampproject/amp.dev/pull/2463 fits into the picture.

Additionally we should think about the versioning in the search.
I suggest old versions should also have the version number in the title.
But this can also be addressed separately.

tharders:
@matthiasrohmer @sebastianbenz Please share your ideas on this topic.

tharders:
I just saw that the version attributes are actually inherited from the translated version!
So the problem why we do not see a version toggle seems to be that the version toggle is not rendered in the normal templates, but rather when importing the documents.
So we should change that.
Since all the version information is available in the document we can easily use this in the normal templates.

tharders:
Still the proposed directory structure would have a benefits over the current structure:
* With lots of versions and lots of translated documents the list of files gets currently quite long
* Currently a translator has to remove the version information that are in the original file in order to avoid problems when new versions are released

sebastianbenz:
Good catch @tharders! I ran also into this while looking into #2895. It gets more complicated:

[...]

PR #2974, opened 2019-09-16
Add support for versioned component documentation

### LINKED ISSUE #2895: Versioned component documentation - support for different versions per format
## Bug Report

Related to: #2770

Some component versions may not be available for all formats, e.g. `amp-carousel` 0.2 is not currently supported for emails.

I made ampproject/amphtml#24340 to correct this in the docs, so `extensions/amp-carousel/amp-carousel.md` (0.2 docs) now doesn't list email in formats, but `extensions/amp-carousel/0.1/amp-carousel.md` (0.1 docs) does. This follows the *New version, doc still in top level folder, old docs in version sub-folders* variation described in #2770.

### Current Behavior

Based on testing locally, this change makes amp-carousel docs hidden all together for emails.

image

### Expected behavior/code

When email is selected as format, the docs for amp-carousel should display 0.1 docs and hide the version selector. Alternatively, the docs should default to 0.1 version docs.

sebastianbenz:
Documentation supports now format specific versioning. For example,
amp-carousel v0.2 will only displayed for websites and ads. Email will
continue to only show the old v0.1 version.

This works as follows:

* Create a format-to-supported-versions mapping during component import.
This JSON file will be written to dist during `importAll`.
* Add available format specific versions to the SSR context
* SSR the version toggle based on the supported versions variable
* Rewrite non-versioned URLs to the latest supported version:

     amp-carousel?format=websites => amp-carousel?format=websites
     amp-carousel?format=email => amp-carousel-v0.1?format=email

Fixes #2895 and #2963


sebastianbenz:
I'm not proud of this and it feels very hacky. Suggestions to simplify this are welcome!

platform/lib/routers/growPages.js
```diff
@@ -168,6 +181,40 @@ function rewriteLinks(canonical, html, format, level) {
...
+  // get the supported version for this format
+  const versionsByFormat = formatComponentMapping[component];
+  if (!versionsByFormat) {
+    signale.warn(`No version mapping defined for ${component}. Run 'gulp importAll' to fix.`);
+    return;
+  }
+  // add format supported versions to template context
+  context.versions = versionsByFormat[context.format];
+  // rewrite path based on the format and available versions
+  const latestVersionByFormat = context.versions[context.versions.length - 1];
```
tharders: When a component doc is opened with an unsupported format `context.versions` is undefined which results in an unhandled exception here.

frontend/templates/views/partials/version-toggle.j2
```diff
@@ -2,9 +2,13 @@
 {# Since it doesn't use grow, global page context is not available. #}
 <h1>
 {{title}}
+[% if versions and versions.length > 1 %]
 <div class="ap--versions">
   {% for v in versions %}
-    <a{% if v !== version %} href="../{{title}}-v{{v}}.html"{% endif %}>{{v}}</a>
+  [% if versions and versions.includes('{{v}}') %]
```
tharders: I really don't like that we have the versions in the SSR, but do the rendering of the template on import.
The markdownDocument.insertVersionToggler replaces the title and adds the version toggle. So I see no reason why we should not simply remove the title and render the title (like other pages) in the template and add the version toggle there, using the generated json structure.

frontend/templates/views/partials/component-sidebar.j2
```diff
@@ -32,7 +32,8 @@
                   {% endif %}
                   {% set formats = component.formats or [] %}
                   {% set activeDocument = true if component.url == doc.url %}
-                  [% if format in {{ formats|tojson }} %]
+                  {# the latest document might not support a format, hence we have to check if any version supports the format #}
+                  [% if format in {{ component.supported_formats|list|tojson }} %]
```
tharders: I think this is the first time we really used the supported_formats property. But we could also use the generated json structure and get rid of the `supported_formats` property in the frontmatter. For nearly all components the values are the same as the normal `formats` anyway. We would not rely on the manually maintained `componentFormats.json` but only on the data we import.

platform/lib/pipeline/componentReferenceImporter.js
```diff
@@ -72,9 +74,13 @@ class ComponentReferenceImporter {
     // Keep track of all promises to complete function
     const savedDocuments = [];
 
+    const versionMapping = {};
     for (const extension of extensions) {
       const documents = await this._findExtensionDocs(extension);
       const versions = [...new Set(documents.map((doc) => doc.version).sort().reverse())];
+      const supportedVersions = {};
+      versionMapping[extension.name] = supportedVersions;
```
tharders: We should move this into the else below to only add the components where we have documentation for. Otherwise there are quite a few useless, empty components in the json.

platform/lib/routers/growPages.js
```diff
@@ -168,6 +181,40 @@ function rewriteLinks(canonical, html, format, level) {
...
+  if (version === versionsByFormat.current) {
+    // change amp-carousel-v$LATEST => amp-carousel
+    url.pathname = url.pathname.replace(`-v${version}`, '');
+  } else if (versionsByFormat.current !== latestVersionByFormat) {
```
tharders: Also check if the version is not set.
If a specific version was requested that is not the current version this code would add a second version part!

frontend/templates/views/partials/version-toggle.j2
```diff
@@ -1,10 +1,12 @@
...
 <h1>
-{{title}}
+{{doc.title}}
+[% if versions and versions.length > 1 %]
 <div class="ap--versions">
-  {% for v in versions %}
-    <a{% if v !== version %} href="../{{title}}-v{{v}}.html"{% endif %}>{{v}}</a>
+  {% for v in doc.versions %}
+  [% if versions and versions.includes('{{v}}') %]
+  <a{% if v != doc.version %} href="../{{doc.title}}-v{{v}}.html"{% endif %}>{{v}}</a>
```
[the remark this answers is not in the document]
tharders: Also doc.title is currently not always the component name which would result in wrong links
Seen with the translations of `amp-ad` now that the doc.title is explicitly defined.
The language switch also seems to generate a different link.

We could choose not to explicitly define the title in the document and inherit the title from the base language.
But the shown headline would be different for this component than before.

pages/content/amp-dev/documentation/components/reference/amp-ad-v0.1@tr.md
```diff
@@ -1,4 +1,5 @@
 ---
+$title: `amp-ad` / `amp-embed`
```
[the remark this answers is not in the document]
sebastianbenz: done

platform/lib/routers/growPages.js
```diff
@@ -178,6 +188,45 @@ const optimizer = AmpOptimizer.create({
...
+      if (!versions) {
+        signale.warn(`No version mapping defined for format ${format}. Try running 'gulp ` +
+        'importAll\' to fix.');
+        // set to the latest available version and tell the user on the page
+        versions = [versionsByFormat.current];
+      }
+      const latestVersionByFormat = versions[versions.length - 1];
+
+      // redirect to unversioned URL for latest format specific version
+      if (version >= latestVersionByFormat) {
```
tharders: String comparison is wrong here (2.0 vs 11.0)

platform/lib/pipeline/markdownDocument.js
```diff
@@ -15,15 +15,11 @@
...
-const fs = require('fs');
 const yaml = require('js-yaml');
 const {Signale} = require('signale');
 const utils = require('@lib/utils');
 
-// Prep version template
-const nunjucks = require('nunjucks');
-const VERSION_TOGGLE_TEMPLATE = nunjucks.compile(fs.readFileSync(
-    utils.project.absolute('frontend/templates/views/partials/version-toggle.j2'), 'utf8'));
+const TITLE_REGEX = /^#{1}\s(.+)/m;
```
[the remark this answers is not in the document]
tharders: TITLE_REGEX is no longer used and has to be removed.

pages/content/amp-dev/documentation/components/reference/amp-ad-v0.1@tr.md
```diff
@@ -10,7 +11,7 @@ teaser:
 
 Bir reklamın görüntüleneceği kapsayıcı. `amp-embed`, `amp-ad` etiketinin bir diğer adıdır ve bu etiketin tüm işlevlerini farklı bir etiket adıyla alır. Anlam açısından daha doğru olduğunda `amp-embed` etiketini kullanın. AMP dokümanları yalnızca HTTPS aracılığıyla sunulan reklamları/yerleştirmeleri destekler.
 
-# `amp-ad` / `amp-embed`
+
```
tharders: The title above and not this one should have been removed.

sebastianbenz: done

platform/lib/routers/growPages.js
sebastianbenz: done

platform/lib/pipeline/componentReferenceImporter.js
```diff
@@ -72,25 +74,42 @@ class ComponentReferenceImporter {
...
           savedDocuments.push(
               this._saveDocument(doc.tagName || extension.name, doc.document, doc.version));
+          doc.document.formats.forEach((format) => {
+            let versionsPerFormat = supportedVersions[format];
+            if (!versionsPerFormat) {
+              versionsPerFormat = [];
+              supportedVersions[format] = versionsPerFormat;
+            }
+            versionsPerFormat.push(doc.version);
+          });
```
pbakaus: I was a little confused by this block (line 100-107), I didn't test this but you can probably turn it into a one liner like so (Sets are nifty because their values are automatically unique, and the add method returns the set itself):

```
supportedVersions[format] = (supportedVersions[format] || new Set()).add(doc.version);
```

platform/lib/pipeline/componentReferenceImporter.js
```diff
@@ -72,25 +74,42 @@ class ComponentReferenceImporter {
...
         });
       }
     }
 
+    writeFileSync(FORMAT_COMPONENT_MAPPING, JSON.stringify(versionMapping, null, 2), 'utf-8');
```
pbakaus: nit: most other stuff in here is async, and you're returning promises anyway. No big deal though.

pages/content/amp-dev/documentation/components/reference/amp-ad-v0.1@tr.md
tharders: You got me wrong here.
Please have a look at another language translation of the same document.
The first h1 should have been removed. The second after the first paragraph should be kept.
And the title in the frontmatter should start with `$title: `

frontend/templates/views/partials/version-toggle.j2
```diff
@@ -1,10 +1,12 @@
...
 <h1>
-{{title}}
+{{doc.title}}
+[% if versions and versions.length > 1 %]
 <div class="ap--versions">
-  {% for v in versions %}
-    <a{% if v !== version %} href="../{{title}}-v{{v}}.html"{% endif %}>{{v}}</a>
+  {% for v in doc.versions %}
+  [% if versions and versions.includes('{{v}}') %]
+  <a{% if v != doc.version %} href="../{{doc.title}}-v{{v}}.html?format=[= format =]"{% endif %}>{{v}}</a>
```
tharders: Try to use doc.slug to avoid the problems with components where the title is not exactly the component name like in the translations of amp-ad.

frontend/templates/views/partials/component-sidebar.j2
sebastianbenz: You're right. Fixed.

frontend/templates/views/partials/version-toggle.j2
sebastianbenz: done

platform/lib/pipeline/componentReferenceImporter.js
sebastianbenz: done

sebastianbenz:
(still WIP)

platform/lib/pipeline/componentReferenceImporter.js
sebastianbenz: Unfortunately, doesn't work as JSON serialization requires an array.

pages/content/amp-dev/documentation/components/reference/amp-ad-v0.1@tr.md
sebastianbenz: done

frontend/templates/views/partials/version-toggle.j2
```diff
@@ -1,10 +1,12 @@
...
 <h1>
-{{title}}
+{{doc.title}}
+[% if versions and versions.length > 1 %]
 <div class="ap--versions">
-  {% for v in versions %}
-    <a{% if v !== version %} href="../{{title}}-v{{v}}.html"{% endif %}>{{v}}</a>
+  {% for v in doc.versions %}
+  [% if versions and versions.includes('{{v}}') %]
+  <a{% if v != doc.version %} href="../{{doc.slug}}-v{{v}}.html?format=[= format =]"{% endif %}>{{v}}</a>
```
tharders: slug is better, but when we simply add the component name to the frontmatter on import we could use this.
In that case I think we would no longer depend on the title for the link generation and could even have 'amp-ad / amp-embed' as a title again.

frontend/templates/views/partials/component-sidebar.j2
tharders: The `supported_formats` is still used here.
We should use the `format-component-mapping.json` data instead.

sebastianbenz:
@tharders `supported_formats` is still required as it's different to what's defined in each document. SupportedFormats contains all formats supported by a component (independent of the version). It's now generated based on the formats defined in the markdown files.

sebastianbenz:
PTAL

sebastianbenz:
format is document specific, e.g.: 

```
format(amp-carousel-v0.1) => websites, **email,** ads
format(amp-carousel-v0.2) => websites, ads.  
```

whereas supported formats is not:

```
supportedFormats(amp-carousel-v0.1) => websites, **email,** ads
supportedFormats(amp-carousel-v0.2) => websites, **email,** ads
```

I agree, naming is not good. `supportedFormats` is easier to use if it's tied to the component frontmatter which is why I left it there. The info is currently not included in format-mapping.json.

Another approach would be to completely change the way we generate the sidebar content by not using grow documents to generate the component list.

sebastianbenz:
> There was one more place where I made a comment to use 'DEFAULT_FORMAT'.

DEFAULT_FORMAT would've been the wrong choice there as this had to be websites. However, the format wasn't exported yet (which is now fixed).

sebastianbenz:
Thanks a lot for the review @tharders!

platform/lib/pipeline/componentReferenceImporter.js
pbakaus: `Array.from(supportedVersions[format])` :) no big deal though.

Issue #2964, opened 2019-09-13
Turn animation guide snippets into executable code samples 


CrystalOnScript:
## 📖 Missing or out-of-date documentation

[AMP Animation Guide](https://amp.dev/documentation/guides-and-tutorials/develop/animations/triggering_css_animations/?format=websites) has awesome animation examples that should be displayed as executable code samples. We should update it as such.

PR #2990, opened 2019-09-19
Examples support multiple code blocks

tharders:
To enable previews for examples like requested in #2964 the example tags now support multiple code blocks with text between.
Before this change it was already possible to add markdown between the example tags and the code block, but the result did not look nice.
Now the multiple code blocks with text between looks like the samples.


pages/content/amp-dev/documentation/guides-and-tutorials/contribute/contribute-documentation/formatting.md
```diff
@@ -474,6 +475,52 @@ This is what it looks like:
...
+    <br>
+    The second body part:
+    <br>
+    ```html
+      &lt;div [text]="'counter: ' + counter">click 'update' to start count&lt;/div>
+      &lt;button on="tap:AMP.setState({counter: counter + 1})">update&lt;/button>
+    &lt;/body>
+    ```
+  [/example]</pre>
+</div>
```
sebastianbenz: Can you also add a preview version. Can we avoid the `<br>`s in between?

tharders: Works without <br>
I used an image for preview, since we really need to have a block to make it clear what belongs together here

tharders:
@sebastianbenz I agree that a preview with multiple code blocks can be confusing when the editor does not explain it well. But looking at the [last animation example](https://amp.dev/documentation/guides-and-tutorials/develop/animations/triggering_css_animations/?format=websites) I think it would be better to use this approach than to add the explanations into the source code, or explain everything beforehand.
For this animation guide I think such example is fine inside the documentation and should not be extracted as a stand alone sample.

CrystalOnScript:
I'll play around with the animation guide and see if I can get things working with the current structure. I'll get back with a more informed opinion by EOD

CrystalOnScript:
I've opened a PR for what it might look like without this update [here](https://github.com/ampproject/amp.dev/pull/2999) - let me know what you think!

sebastianbenz:
@tharders I've discussed this with @CrystalOnScript. We've decided to not support this feature to keep inline samples simple. Thanks for looking into this!

PR #3024, opened 2019-09-25
Remove 'Email Provider integration' doc

fstanis:
It's outdated and I'm hoping to supersede it with the new viewer documentation and reference implementation docs.


sebastianbenz:
Is there content we should link to instead? Maybe setup a redirect.

fstanis:
There's no good substitute to link to right now, other than maybe the spec, but that sounds wrong IMO.

sebastianbenz:
Hm. I'm not happy about simply removing a page without replacement. Maybe instead remove the outdated bits and add a note that the guide is being rewritten?

fstanis:
Done. PTAL.

PR #3029, opened 2019-09-26
Only perform https redirect if scheme is set to https

tharders:
This enables local testing with other host name than 'localhost'
Useful for testing with other devices in a local network.
Still not all things will work without https (like search)

### DIFF
--- a/platform/lib/middleware/redirects.js
+++ b/platform/lib/middleware/redirects.js
@@ -92,6 +92,11 @@ module.exports = (req, res, next) => {
     return;
   }
 
+  // the following checks are only relevant for https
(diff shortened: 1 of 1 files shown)


matthiasrohmer:
Hm... why not change

https://github.com/ampproject/amp.dev/blob/2a420c0ea4b165f00c0e769177937ab17d2a2087/platform/lib/middleware/redirects.js#L85

to `if (config.isDevMode())` instead?

tharders:
@matthiasrohmer Good point, the localhost check was not nice. The good thing is that we can simply remove the check, since the next test for the 'www' prefix works on localhost.

PR #3046, opened 2019-09-30
Animated hamburger menu example 

CrystalOnScript:


examples/source/multimedia-animations/Animated_hamburger_menu.html
```diff
@@ -0,0 +1,95 @@
...
+<body>
+ 
+  <!-- ## Markup -->
+  <!--
+    The hamburger menu element contains three styled `span` elements inside a `div`. The hamburger `div` uses the `tap` AMP action to trigger the `toggleClass()` event, which targets an element by its `id`. In this example, it targets `hamburger`.
+
+    The `toggleClass()` event will add and remove any class it's given as an argument. This example passes `class='close'`.
+  -->
+  <div style="padding: 5px;">
+    <div id="hamburger" tabindex=1 role=button on="tap:hamburger.toggleClass(class='close')">
```
sebastianbenz: nit: use double quotes for all attributes.

examples/source/multimedia-animations/Animated_hamburger_menu.html
```diff
@@ -0,0 +1,95 @@
...
+</head>
+<body>
+ 
+  <!-- ## Markup -->
+  <!--
+    The hamburger menu element contains three styled `span` elements inside a `div`. The hamburger `div` uses the `tap` AMP action to trigger the `toggleClass()` event, which targets an element by its `id`. In this example, it targets `hamburger`.
+
+    The `toggleClass()` event will add and remove any class it's given as an argument. This example passes `class='close'`.
+  -->
+  <div style="padding: 5px;">
```
sebastianbenz: I'd remove this wrapper div and define the margin in the main CSS.

examples/source/multimedia-animations/Animated_hamburger_menu.html
```diff
@@ -0,0 +1,95 @@
...
+        #hamburger {
+            width: 60px;
+            height: 45px;
+            position: relative;
+            margin: 50px auto;
+            transform: rotate(0deg);
+            transition: .5s ease-in-out;
+            cursor: pointer;
+            outline: none;
+            }
```
sebastianbenz: nit: formatting

examples/source/multimedia-animations/Animated_hamburger_menu.html
```diff
@@ -0,0 +1,95 @@
...
+            opacity: 1;
+            left: 0;
+            transform: rotate(0deg);
+            transition: .5s ease-in-out;
+        }
+        #hamburger span:nth-child(1) {
+            top: 0px;
+            transform-origin: left center;
+        }
+
```
sebastianbenz: nit: consistently use empty lines as separator or not

examples/source/multimedia-animations/Animated_hamburger_menu.html
```diff
@@ -0,0 +1,95 @@
...
+  <title>Animated hamburger menu</title>
+  <script async src="https://cdn.ampproject.org/v0.js"></script>
+  <link rel="canonical" href="<% canonical %>">
+  <meta name="viewport" content="width=device-width,minimum-scale=1,initial-scale=1">
+  <style amp-boilerplate>body{-webkit-animation:-amp-start 8s steps(1,end) 0s 1 normal both;-moz-animation:-amp-start 8s steps(1,end) 0s 1 normal both;-ms-animation:-amp-start 8s steps(1,end) 0s 1 normal both;animation:-amp-start 8s steps(1,end) 0s 1 normal both}@-webkit-keyframes -amp-start{from{visibility:hidden}to{visibility:visible}}@-moz-keyframes -amp-start{from{visibility:hidden}to{visibility:visible}}@-ms-keyframes -amp-start{from{visibility:hidden}to{visibility:visible}}@-o-keyframes -amp-start{from{visibility:hidden}to{visibility:visible}}@keyframes -amp-start{from{visibility:hidden}to{visibility:visible}}</style><noscript><style amp-boilerplate>body{-webkit-animation:none;-moz-animation:none;-ms-animation:none;animation:none}</style></noscript>
+
+  <!-- Custom styles and CSS animations for the hamburger menu. -->
+    <style amp-custom>
+        #hamburger {
+            width: 60px;
```
sebastianbenz: use two spaces for indentation

CrystalOnScript:
@tharders - Thanks for the thorough feedback! I think this an interesting issue. If we approach the with the z-index fix it messes with the code preview:
Screen Shot 2019-10-30 at 4 54 45 PM

However, maybe your solution with the iframe would solve this?

I'm torn about implementing our own menu layer - I tried to take that approach at first, but I feel like we should communicate how to use the tools available in AMP and avoid having people build custom solutions.

CrystalOnScript:
Also just realized if we make the hamburger menu clickable for the user while the sidebar is open, and they close out of amp-sidebar without clicking the button it doesn't trigger the animation. However, we still leave the `on="sidebarClose:hamburger.toggleClass(class='close')"` on amp-sidebar and the user does click the hamburger menu it toggles the class on and then off again... So, I'm struggling a bit with one :/

tharders:
@CrystalOnScript I discussed this with @sebastianbenz and @sebil and we came to the conclusion that amp-sidebar does not work with a hamburger menu icon that becomes a close icon. So we think it would be best to self implement a menu layer with a little animation when opened and closed. 
If you need help with that let us know.

By the way I found out that you can remove a class with: `toggleClass(class='close', force=false)` and add a class with `toggleClass(class='close', force=true)`, so the problem with the amp-sidebar close event could have been solved this way.

examples/source/multimedia-animations/Animated_hamburger_menu.html
```diff
@@ -0,0 +1,127 @@
...
+        width: 0%;
+        opacity: 0;
+        transition: .1s;
+      }
+
+      #hamburger.close span:nth-child(3) {
+        transform: rotate(-45deg);
+      }
+      
+      .not-active {
```
tharders: The naming is not very accurate, since the class will never be removed.
Why not simply define the styles as default by using `#nav-menu` as selector.

examples/source/multimedia-animations/Animated_hamburger_menu.html
```diff
@@ -0,0 +1,127 @@
...
+        position: relative;
+        transform: translateX(-100%);
+        opacity: 0;
+        z-index: 499;
+        transition: transform .5s ease, opacity ease .2s;
+      }
+      .now-active {
```
tharders: When we define the 'not active' styles by using an id selector we also need to use the id here.
So the selector would be `#nav-menu.now-active`
(Or maybe simply use `active` instead of `now-active`)

examples/source/multimedia-animations/Animated_hamburger_menu.html
```diff
@@ -0,0 +1,127 @@
...
+ <div>
+  <div class="hamburger_wrapper">
+    <div id="hamburger" tabindex="1" role="button" on="tap:hamburger.toggleClass(class='close'),nav-menu.toggleClass(class='now-active')">
+      <span></span>
+      <span></span>
+      <span></span>
+    </div>
+  </div>
+  <div id="nav-menu" class="not-active">
+    <ul id="nav-list">
```
tharders: Is it really necessary to use an id here?
When it is just used for styling we could use `class="nav-list"` and adjust the css above accordingly.

examples/source/multimedia-animations/Animated_hamburger_menu.html
```diff
@@ -0,0 +1,127 @@
...
+  -->
+ <div>
+  <div class="hamburger_wrapper">
+    <div id="hamburger" tabindex="1" role="button" on="tap:hamburger.toggleClass(class='close'),nav-menu.toggleClass(class='now-active')">
+      <span></span>
+      <span></span>
+      <span></span>
+    </div>
+  </div>
+  <div id="nav-menu" class="not-active">
```
tharders: With the css changes above we could remove the `class` attribute here.

tharders:
I just saw that the z-index for the menu is higher than the z-index for the amp.dev menu.
Please use a z-index of 10 for both places where z-index is specified.
Also we need to move the nav-menu completely out of the screen. We can do this by using `-100vw` instead of `-100%` for the `translateX` property.

examples/source/multimedia-animations/Animated_hamburger_menu.html
```diff
@@ -0,0 +1,127 @@
...
+        padding: 5px;
+        z-index: 500;
+      }
+
+      #hamburger {
+        width: 60px;
+        height: 45px;
+        position: relative;
+        transform: rotate(0deg);
+        transition: .5s ease-in-out;
```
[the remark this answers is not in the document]
CrystalOnScript: nope - removed!

examples/source/multimedia-animations/Animated_hamburger_menu.html
CrystalOnScript: I originally tried it that way, but the css specificity doesn't allow a class to override an id styles - unless I'm doing it wrong, which is possible, heh

examples/source/multimedia-animations/Animated_hamburger_menu.html
CrystalOnScript: `active` appears to be a class already used by amp.dev and it applies the colors to the side menu, lol.

examples/source/multimedia-animations/Animated_hamburger_menu.html
```diff
@@ -0,0 +1,129 @@
...
+<!doctype html>
+<html ⚡>
+<head>
+  <meta charset="utf-8">
+  <title>Animated hamburger menu</title>
+  <script async src="https://cdn.ampproject.org/v0.js"></script>
+  <!-- 
+    ## Introduction 
+    Small interactive UI animations communicate state and purpose to users. A common case is a hamburger menu button that animates into an "X" when open. This animation communicates that the user can press the same button again to close the menu. 
+  -->
```
tharders: The ` <link rel="canonical"` below this comment will still be shown as a code block in the documentation.
That does not really make sense, does it?
Simply move the text at the top of the `CSS animations` comment block and separate them with a new line so that no code block will be shown between them.

examples/source/multimedia-animations/Animated_hamburger_menu.html
tharders: When you use the id for both states it will work.
So the basic and inactive styling goes with `#nav-menu`
and the active styling goes with `#nav-menu.now-active`

examples/source/multimedia-animations/Animated_hamburger_menu.html
tharders: Hehe, then lets keep `now-active`

PR #3051, opened 2019-10-01
Go links for AMP roadshow tutorials

morsssss:
I don't suppose there's a way for me to test these go links on my machine?

I get nervous committing things without being able to test them first :)


morsssss:
Thanks for mentioning that... added the slashes!

PR #3052, opened 2019-10-01
added go links for presentation

CrystalOnScript:
Go links for docs and stuff!


platform/config/go-links.yaml
```diff
@@ -5,22 +5,26 @@
 /ampcs: /events/amp-cs-2019/#Schedule
 /auto-sw: https://github.com/ampproject/amp-sw/
 /camp-code: https://github.com/ampproject/samples/tree/master/amp-camp
+/codesamples: https://amp.dev/documentation/guides-and-tutorials/contribute/contribute-documentation/formatting/?format=websites#preview-code-samples
```
sebastianbenz: suggestion: use `-` as delimiter similar to the other links `code-samples` 
also, links to amp.dev don't need to be absolute

PR #3056, opened 2019-10-02
Better go links for AMP roadshow tutorials

morsssss:
@sebastianbenz , thanks for showing me how to test these.

Turns out the stories link didn't work so well without `?format=stories`.

(Which inspires me to think that it would be nice if pages about stories like that one didn't require the query string...)


sebastianbenz:
Good point. We actually should set the format as a default if there's only one supported format. I've filed https://github.com/ampproject/amp.dev/issues/3057.

sebastianbenz:
Can you fix the merge conflict (and also add the trailing slash).

morsssss:
Ok, done

PR #3061, opened 2019-10-02
support query params

sebastianbenz:


ithinkihaveacat:
I don't completely understand exactly what is happening here (by default URLs containing `?` are rejected by the packager? why is this not the default?), but it seems in practice it will either work as expected or not, so LGTM.

/cc @twifkak

twifkak:
Change LGTM. The default is for two reasons:
 1. To reduce the chance that amppkg [signs personalized content](https://wicg.github.io/webpackage/draft-yasskin-http-origin-signed-responses.html#seccons-over-signing). This is in addition to [checking](https://github.com/ampproject/amppackager/blob/c9993b8ac4d17d1f05d3a1289956dadf3f9c370a/packager/signer/validation.go#L234) that it's publicly cacheable. "Contains query params" is a weak proxy indicator for "contains personalized content" but I wanted to play it safe.
 2. To reduce the [risk of content-sniffing](https://wicg.github.io/webpackage/draft-yasskin-http-origin-signed-responses.html#seccons-content-sniffing), by reducing the entropy of the fallback URL. This is in addition to [limiting the allowed codepoints in the URL](https://github.com/ampproject/amppackager/blob/c9993b8ac4d17d1f05d3a1289956dadf3f9c370a/packager/signer/validation.go#L126).

Both are defense-in-depth things, so it's fine to reduce given knowledge of your site.

sebastianbenz:
Thanks both for taking a look!

Issue #3031, opened 2019-09-27
Grow installation instructions in README seem wrong


DerekNonGeneric:
## 📖 Missing or out-of-date documentation

**Describe the content that is missing or should be up dated**
The Requirements section of the README says to use `pip` instead of its installer when installing Grow. Simply running `pip install grow` seems to only install the module to `~/.local/lib/python2.7/site-packages`. At least this is what I experienced on Debian 10. I think it might be necessary to run Grow's install script, which adds `bin/grow` to the user's home directory and then adds that to the user's `PATH`.

**Which part of the site would that content live at?**
The amp.dev README file.

**What's the type of content?**
Local development instructions.

**Optional: Proposed information architecture**
I don't know why the instructions specify to not use the installer. I was only able to install it on my system with this procedure:

1. `pip install grow`
1. `curl https://install.grow.io | bash`

I would probably include both steps in the README unless someone knows why not.

DerekNonGeneric:
@matthiasrohmer what should `which grow` be outputting?

DerekNonGeneric:
I have discovered that there is a step that is missing! It is explained [Installing to the User Site](https://packaging.python.org/tutorials/installing-packages/#id6) from the pip docs. Following those instructions, I was able to convert what we currently had to these three commands. The first command is for local user installation of Grow. The second command can be omitted if that directory is already in the user's `$PATH`, but the _user base binary directory_ **does need** to be added to `$PATH` on a fresh install of Debian 10. The third command avoids having to log in and out for the changes to take effect.

```sh
$ pip install --user grow
$ echo "export PATH=\"$(python -m site --user-base)/bin\":\$PATH" >> ~/.bashrc
$ source ~/.bashrc
```

DerekNonGeneric:
Now using the raw Python module, running `grow --version` produces a warning:

```
Warning: libyaml missing, using slower yaml parser.
```

I was able to remedy this warning by doing the following.

**Debian-based (tested)**
```bash
sudo apt install -y python-yaml
```
**macOS (untested)**
```bash
brew install libyaml
```
---
Simply installing the pip package did not resolve the warning on Debian.

```bash
pip install --user pyyaml
```
^ This might be slower parser mentioned in the warning.

matthiasrohmer:
[...]

PR #3064, opened 2019-10-03
Include OS-specific pip setup instructions

DerekNonGeneric:
Fixes: https://github.com/ampproject/amp.dev/issues/3031

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,22 +28,53 @@ We welcome contributions to amp.dev.
     $ nvm install --lts
     ```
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -28,22 +28,53 @@ We welcome contributions to amp.dev.
     $ nvm install --lts
     ```
 
+1. Install Python 2.7 and ensure pip is properly set up by adding the _pip user base binary directory_ to `$PATH`.
```
sebastianbenz: Can you mark this step as optional or mention that this is only required if python and pip are not yet installed.

DerekNonGeneric: Alright, done.

PR #3076, opened 2019-10-07
Fix sample doc pages not validating

sebastianbenz:
The samples define head tags are injected into the template
causing validation errors due to duplicated tags. I couldn't
come up with a clean handle this on a template level. Instead,
I've opted for writing a optimizer transformer that'll remove
duplicates.

### DIFF
--- a/platform/lib/utils/HeadDedupTransformer.js
+++ b/platform/lib/utils/HeadDedupTransformer.js
@@ -24,8 +24,8 @@ const TAGS_TO_DEDUP = {
 };
 
 /**
- * A custom transformer for the @ampproject/toolbox-optimizer rewriting
- * CSS selectors to save bytes in <style amp-custom>.
+ * A custom transformer for the @ampproject/toolbox-optimizer removing
+ * all duplicated viewport and link rel=canonical tags.
  */
 class HeadDedupTransformer {
   constructor(config) {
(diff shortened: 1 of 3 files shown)


platform/lib/utils/HeadDedupTransformer.js
```diff
@@ -0,0 +1,79 @@
...
+    name: 'viewport',
+  },
+  link: {
+    rel: 'canonical',
+  },
+};
+
+/**
+ * A custom transformer for the @ampproject/toolbox-optimizer rewriting
+ * CSS selectors to save bytes in <style amp-custom>.
```
tharders: This text does not fit.

platform/lib/utils/HeadDedupTransformer.js
```diff
@@ -0,0 +1,79 @@
+/**
```
tharders: The filename should start with a lowercase character

sebastianbenz: uppercase when exporting a class

sebastianbenz:
Some amp components require custom head elements, e.g. amp-access configures it's auth endpoints.

platform/lib/utils/HeadDedupTransformer.js
sebastianbenz: done

sebastianbenz:
PTAL

sebastianbenz:
👍

Issue #2499, opened 2019-07-15
Code snippet formatting feels claustrophobic 


CrystalOnScript:
### Current Behavior
Code snippets that require side scrolling don't have much room between the end of the snippet and the side of the snippet box. It makes it difficult to copy just the code snippet and not other elements on the page. 

Please also provide:
- Page link: [link](https://amp.dev/documentation/guides-and-tutorials/develop/interactivity/advanced-interactivity/?format=websites)
- [Screenshots](https://snag.gy/Za9he3.jpg)

matthiasrohmer:
@sebil, can you please investigate with Longo?

PR #3080, opened 2019-10-08
💄 Optimize spacing at the end of line in code-snippets

sebil:
I added multiple `&nbsp;` as spacer after the code-snippets to force a space between the last character in the longest line and the edge of the box.

I know this is not the prettiest solution, but after trying to fix this with different approaches like padding, margin, width, calculation etc., I'm almost sure that this is the only solution.
Additionally I made sure that the characters are not selectable.

If anyone has an idea to fix this in a different way, please let me know.

Closes: #2499

Screenshot 2019-10-08 at 14 30 40


matthiasrohmer:
As discussed offline, no it wouldn't because the last line isn't always the longest.

Issue #2991, opened 2019-09-19
Sample preview pages need an experiment toggle 


sebastianbenz:
The sample doc page shows the toggle, but the preview page needs it as well as it's on a different origin.

//cc @matthiasrohmer

PR #3092, opened 2019-10-09
✨ Implement the experiment toggle in standalone sample preview 

sebil:
With this PR the existing experiment toggle from the example pages coming to the standalone sample previews (View demo button).

Closes: #2991

Screenshot 2019-10-11 at 12 30 11

## Desktop
Screenshot 2019-10-11 at 12 30 26

## Mobile
Screenshot 2019-10-11 at 12 31 10

### DIFF
--- a/pages/content/shared/experiments.html
+++ b/pages/content/shared/experiments.html
@@ -12,5 +12,16 @@ $sitemap:
   <script async src="https://cdn.ampproject.org/v0.js"></script>
 </head>
 <body>
+  <div class="ap-m-experiments ap-m-experiments-iframe">
+    {% include '/views/partials/experiments/experiments-controls.j2' %}
+  </div>
+  <script>
+  const experiments = [
+    [%- for experiment in experiments -%]
+      "[= experiment =]"[= ', ' if not loop.last =]
+    [%- endfor -%]
+  ]
+  </script>
+  {% include '/views/partials/experiments/experiments-script.j2' %}
 </body>
 </html>
--- a/pages/content/shared/experiments.html
+++ b/pages/content/shared/experiments.html
@@ -10,6 +10,14 @@ $sitemap:
   <meta name="viewport" content="width=device-width,minimum-scale=1">
   <style amp-boilerplate>body{-webkit-animation:-amp-start 8s steps(1,end) 0s 1 normal both;-moz-animation:-amp-start 8s steps(1,end) 0s 1 normal both;-ms-animation:-amp-start 8s steps(1,end) 0s 1 normal both;animation:-amp-start 8s steps(1,end) 0s 1 normal both}@-webkit-keyframes -amp-start{from{visibility:hidden}to{visibility:visible}}@-moz-keyframes -amp-start{from{visibility:hidden}to{visibility:visible}}@-ms-keyframes -amp-start{from{visibility:hidden}to{visibility:visible}}@-o-keyframes -amp-start{from{visibility:hidden}to{visibility:visible}}@keyframes -amp-start{from{visibility:hidden}to{visibility:visible}}</style><noscript><style amp-boilerplate>body{-webkit-animation:none;-moz-animation:none;-ms-animation:none;animation:none}</style></noscript>
   <script async src="https://cdn.ampproject.org/v0.js"></script>
+  <style amp-custom>
+    {% do doc.styles.addCssFile('css/base.css', -3) %}
+    {% do doc.styles.addCssFile('css/components/atoms/anchor.css', -2) %}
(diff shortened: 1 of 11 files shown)


sebil:
Status blocked, because unfortunately, the experiment toggle couldn't work on a different domain (preview.amp.dev) right now.

We'll take care of it.

platform/lib/routers/example/experiments.js
```diff
@@ -0,0 +1,37 @@
...
+// eslint-disable-next-line new-cap
+const sharedPages = express.Router();
+
+sharedPages.get('/shared/experiments/', async (req, res, next) => {
+  try {
+    const template = await Templates.get('/shared/experiments.html');
+    const rendered = template.render({'experiments': req.query.experiments.split(',')});
+    res.send(rendered);
+  } catch (e) {
+    // page not found
```
matthiasrohmer: The comment seems off 😉Also pass `e` to `next` as if something goes wrong that is most probably a nunjucks/rendering error.

platform/lib/routers/example/experiments.js
```diff
@@ -0,0 +1,37 @@
...
+const {Templates, createRequestContext} = require('@lib/templates/index.js');
+const express = require('express');
+
+// eslint-disable-next-line new-cap
+const sharedPages = express.Router();
+
+sharedPages.get('/shared/experiments/', async (req, res, next) => {
+  try {
+    const template = await Templates.get('/shared/experiments.html');
+    const rendered = template.render({'experiments': req.query.experiments.split(',')});
```
matthiasrohmer: Is `req.query.experiments` still a string if not present in the URL? Could be worth guarding with `(req.query.experiments || '')` or similar.

pages/content/shared/experiments.html
```diff
@@ -0,0 +1,36 @@
...
+    body { display: flex; min-height: 100vh; align-items: center; }
+  </style>
+</head>
+<body>
+  <div class="ap-m-experiments ap-m-experiments-iframe">
+    {% include '/views/partials/experiments/experiments-controls.j2' %}
+    <b class="reload-hint" hidden>{{ _('Please reload the page.' ) }} ↻</b>
+  </div>
+  <script>
+  const experiments = [
```
matthiasrohmer: You could use `[= experiments|tojson =]` instead of building the array literally.

platform/lib/routers/example/experiments.js
```diff
@@ -0,0 +1,37 @@
...
+ * distributed under the License is distributed on an "AS-IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+'use strict';
+
+
+const {Templates, createRequestContext} = require('@lib/templates/index.js');
```
matthiasrohmer:
```suggestion
const {Templates} = require('@lib/templates/index.js');
```

That's an open linter issue. 🙂

platform/lib/routers/example/experiments.js
sebil: Good point, an error occurred when the URL was empty. Thanks!

pages/content/shared/experiments.html
sebil: Thanks for pointing this out, I changed it to
`const experiments = [= experiments|dump|safe =]`

Way better!

sebil:
Thanks for your feedback and review! I've done all the requested changes and updated the screenshots above.
Feel free to test and feedback again. Otherwise, feel free to approve 😊

platform/lib/routers/example/experiments.js
```diff
@@ -0,0 +1,36 @@
...
+const {Templates} = require('@lib/templates/index.js');
+const express = require('express');
+
+// eslint-disable-next-line new-cap
+const sharedPages = express.Router();
+
+sharedPages.get('/shared/experiments/', async (req, res, next) => {
+  try {
+    const template = await Templates.get('/shared/experiments.html');
+    const rendered = template.render({'experiments': (req.query.experiments || '').split(',')});
```
sebastianbenz: Can we read the query parameters in the iframe embed? This way we can avoid SSR'ing the experimental iframe embed.

sebil:
The query parameter is read out of the `window.location` now, so we don't need SSR anymore. Indeed the router has to be existent, because of the preview URL. Otherwise, the iframe embed redirects to the main domain.

Additionally, I solved the parent reload issue with a `postMessage`, so we get rid of the `reload-hint` for manually reloading the page. 👍

sebil:
I thought the page was invalid before, so it'd be maybe okay.
I reverted the last commit now, so it's the manual reload button again. 👍

PR #3125, opened 2019-10-21
Search test should not rely upon external data

### LINKED ISSUE #3118: Search autosuggest test randomly fails
As discussed offline: `autosuggest should return an ordered list with components` in search.test.js randomly fails and blocks builds and/or pushes due to the commit hooks.

The response should probably be mocked if it makes sense and doesn't make the test obsolete.

tharders:
Fix #3118


matthiasrohmer:
Thanks, @tharders! @sebastianbenz, as of this morning: this would make the search test reproducible finally. Or would you rather have it stay dependent on component-versions.json?

PR #3128, opened 2019-10-22
Format toggle links to current page if target format is supported

### LINKED ISSUE #3126: Format switcher behavior is confusing to users
### Current Behavior
When user's are on a document under format X and then switch it to format y, they are redirect to that documentation section landing page.

Example:
User is under documentation>components>amp-form#websites and they switch to #email. The user is taken to documentation>components#email. This is jarring and confusing users. They may think amp-form may not be usable in AMP emails. 

### Expected behavior/code
Now that we have the "this component/document is not valid for the selected format" header on individual pages, we should not redirect users to the "landing page". Instead we should only switch the format and display the header if needed.

tharders:
Fixes #3126

### DIFF
--- a/frontend/templates/views/partials/format-toggle.j2
+++ b/frontend/templates/views/partials/format-toggle.j2
@@ -5,17 +5,6 @@
 {% if not '/courses/' in doc.pod_path %}
 {% set formats = ['websites', 'stories', 'ads', 'email'] %}
 
-{# Define various landing pages for sections #}
-{% if 'examples' in doc.url.path %}
-{% set landing_page = g.doc('/content/amp-dev/documentation/examples/index.html') %}
-{% elif 'components' in doc.url.path %}
-{% set landing_page = g.doc('/content/amp-dev/documentation/components/index.html') %}
-{% elif 'courses' in doc.url.path %}
-{% set landing_page = g.doc('/content/amp-dev/documentation/courses/index.html') %}
-{% else %}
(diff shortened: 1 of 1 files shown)


frontend/templates/views/partials/format-toggle.j2
```diff
@@ -32,7 +32,12 @@
     {% for format in formats %}
     [% if format != '{{ format }}' %]
     {% do doc.icons.useIcon('icons/amp-' + format + '.svg') %}
-    <a class="ap-m-format-toggle-link ap-m-format-toggle-link-{{ format }}" href="{{ landing_page.url.path }}?format={{ format }}">
+    {% if format in doc.formats %}
+      {% set format_target_url = doc.url.path %}
+    {% else %}
+      {% set format_target_url = landing_page.url.path %}
```
sebastianbenz: We should always link to the same URL (even if the component is not supported). Otherwise the behavior is confusing.

sebastianbenz: The page will show the unsupported format notification which explains the issue to the user.

tharders: I thought we should not link explicitly to unsupported formats.
But since we use the canonical url there should be no problem with the search engines so I will change this.

tharders:
@sebastianbenz Yea, me too! Was a pleasure to remove the large if else block :-D

sebastianbenz:
(build failure is unrelated to this change)

PR #3133, opened 2019-10-22
Download validator.js, so that tests can run offline.

tharders:
With this changes the unit tests run offline.
Only when there is a not responsive connection the test execution seem to wait for some request.
But it is not the update of the validator.js, since I use a timeout there.

Not sure if the folder where the validator.js is stored is perfect...
Feedback welcome :-)


platform/lib/format-transform/index.test.js
```diff
@@ -13,6 +13,16 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+const {join} = require('path');
+const {BUILD} = require('@lib/utils/project').paths;
+
+// Use the downloaded validator.js:
+const VALIDATOR_TARGET_PATH = join(BUILD, 'testing/validator.js');
```
sebastianbenz: Can we add a check that'll test if the file exists and prints a warning if it doesn't?

PR #3170, opened 2019-10-30
Removed Busan from roadshows

morsssss:
We had to cancel Busan :(


sebastianbenz:
Maybe it's better to keep it on the list, but mark it as cancelled to avoid confusion.

morsssss:
I see your point. Except, in this case, only three people signed up for Busan. We've notified them personally!

Personally, given this circumstance, I'd prefer to leave it at that rather than have us advertise on our site that we had to cancel a stop.

sebastianbenz:
SGTM then

PR #3178, opened 2019-11-01
Fixing AMP Start 2 demo path names to images.

alankent:
I think the images got moved during a reorg of files so the mock API calls were not returning images correctly. In other cases the referenced file did not exist at all, so I put a filename in of a different file that seemed close. There were also some incorrect file types in extensions.

(This used to be a part of PR #2767 but separating into smaller independent PRs.)


alankent:
Sorry, forgot to create new branch for committing the next files. The new CSS file is meant for a separate PR. Reverted.

PR #3254, opened 2019-12-03
Add go links for contribution process and I2I

rcebulko:
To be used in PR template in `ampproject/amphtml`

Related: https://github.com/ampproject/amphtml/pull/25841

### DIFF
--- a/platform/config/go-links.yaml
+++ b/platform/config/go-links.yaml
@@ -5,9 +5,9 @@
 /ampcs: /events/amp-cs-2019/#Schedule
 /auto-sw: https://github.com/ampproject/amp-sw/
 /camp-code: https://github.com/ampproject/samples/tree/master/amp-camp
+/contribute/code: https://github.com/ampproject/amphtml/blob/master/contributing/contributing-code.md
 /contribute/code-samples: /documentation/guides-and-tutorials/contribute/contribute-documentation/formatting/?format=websites#preview-code-samples
 /contribute/docs: /documentation/guides-and-tutorials/contribute/contribute-documentation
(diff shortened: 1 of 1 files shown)


platform/config/go-links.yaml
```diff
@@ -7,6 +7,7 @@
 /camp-code: https://github.com/ampproject/samples/tree/master/amp-camp
 /contribute/code-samples: /documentation/guides-and-tutorials/contribute/contribute-documentation/formatting/?format=websites#preview-code-samples
 /contribute/docs: /documentation/guides-and-tutorials/contribute/contribute-documentation
+/contribute/process: https://github.com/ampproject/amphtml/blob/master/contributing/contributing-code.md
```
sebastianbenz: How about `contribute/code`? Process is quite generic.

rcebulko: Done!

rcebulko: As I notice that most of these links are within amp.dev documentation, it does make me wonder if the contribution process belongs somewhere in amp.dev and the `CONTRIBUTING.md` in various repos should point to it. Out of scope for this PR, but thought I'd mention it.

matthiasrohmer: That'd be great! A first step could also be to simply import /amphtml/contributing/contributing-code.md by adding it to https://github.com/ampproject/amp.dev/blob/future/platform/config/imports/spec.json in order to have it available for a broader audience. 🙂

rcebulko: I'll do this in a follow-up PR which can update these go links accordingly, since there will probably be some discussion about the specifics of the import.

PR #3267, opened 2019-12-06
Update amp-autocomplete documentation to include examples for "inline" and "query"

caroqliu:


examples/source/1.components/amp-autocomplete.html
```diff
@@ -321,5 +321,113 @@ <h2>Display default items</h2>
     </form>
   </div>
 
+  <h2>Inline autocomplete</h2>
+  <!-- 
+    ## Displaying suggestions inline
+
+    Suggestions can be triggered on a specified character token in an `amp-autocomplete` for multiple autosuggestions in a single input by using the `inline` attribute.
```
nainar: It's not clear that the trigger is specified on the input attribute itself. 

Also make it clear that if no trigger is present this will trigger after every space etc

caroqliu: revised, ptal :)

PR #3277, opened 2019-12-10
✨ Cache latest search query and results in service worker

sebil:
With this PR the latest search query and results will be cached via the service worker.

Currently, there may appear a styling glitch in the results list (sporadically), so this is why this still is a Draft PR.

Feel free to test. I look forward to your feedback :)

### DIFF
--- a/frontend/templates/views/partials/header.j2
+++ b/frontend/templates/views/partials/header.j2
@@ -87,7 +87,7 @@
     </nav>
 
     {% do doc.styles.addCssFile('/css/components/molecules/search-trigger.css') %}
-    <div id="searchTriggerOpen" class="ap-m-search-trigger" on="tap:searchLightbox" role="button"  tabindex="0">
+    <div id="searchTriggerOpen" class="ap-m-search-trigger" on="tap:searchLightbox,AMP.setState({})" role="button"  tabindex="0">
       {% do doc.icons.useIcon('/icons/search.svg') %}
       <div class="ap-a-ico ap-m-search-trigger-icon">
         <svg><use xmlns:xlink="http://www.w3.org/1999/xlink" xlink:href="#search"></use></svg>
--- a/frontend/templates/views/partials/header.j2
+++ b/frontend/templates/views/partials/header.j2
@@ -87,7 +87,7 @@
     </nav>
 
     {% do doc.styles.addCssFile('/css/components/molecules/search-trigger.css') %}
-    <div id="searchTriggerOpen" class="ap-m-search-trigger" on="tap:searchLightbox,AMP.setState({})" role="button"  tabindex="0">
+    <div id="searchTriggerOpen" class="ap-m-search-trigger" on="tap:searchLightbox" role="button"  tabindex="0">
       {% do doc.icons.useIcon('/icons/search.svg') %}
       <div class="ap-a-ico ap-m-search-trigger-icon">
         <svg><use xmlns:xlink="http://www.w3.org/1999/xlink" xlink:href="#search"></use></svg>
--- a/frontend/templates/views/partials/search.j2
+++ b/frontend/templates/views/partials/search.j2
@@ -76,63 +76,61 @@
                     reset-on-refresh
                     single-item
           >
-            {% raw %}
             <template type="amp-mustache">
               <div class="ap-o-search-result-list">
-                {{^prevUrl}}
-                {{#result.components.length}}
-                <h4 class="ap-o-search-result-category">{% endraw %}{{_('Components')}}{% raw %}</h4>
-                {{/result.components.length}}
-                {{/prevUrl}}
-                {{#result.components}}
+                {{'{{^prevUrl}}'}}
+                {{'{{#result.components.length}}'}}
+                <h4 class="ap-o-search-result-category">{{_('Components')}}</h4>
+                {{'{{/result.components.length}}'}}
+                {{'{{/prevUrl}}'}}
+                {{'{{#result.components}}'}}
                 <div class="ap-o-search-result-item ap-o-search-result-item-highlight">
-                  <a class="ap-o-search-result-link" href="{{url}}">
-                    <h4 class="ap-o-search-result-title">{{title}}</h4>
-                    <p class="ap-o-search-result-description">{{description}}</p>
+                  <a class="ap-o-search-result-link" href="{{'{{url}}'}}">
+                    <h4 class="ap-o-search-result-title">{{'{{title}}'}}</h4>
+                    <p class="ap-o-search-result-description">{{'{{description}}'}}</p>
                   </a>
-                  {{#exampleUrl}}
+                  {{'{{#exampleUrl}}'}}
                   <div class="ap-o-search-result-linklist">
-                    <a class="ap-o-search-result-linklist-item ap-o-search-result-linklist-item-example ap-m-lnk" href="{{exampleUrl}}">
+                    <a class="ap-o-search-result-linklist-item ap-o-search-result-linklist-item-example ap-m-lnk" href="{{'{{exampleUrl}}'}}">
                       <div class="ap-a-ico ap-m-lnk-icon">
                         <svg><use xmlns:xlink="http://www.w3.org/1999/xlink" xlink:href="#internal"></use></svg>
                       </div>
-                      <span class="ap-m-lnk-text">{% endraw %}{{_('Open example')}}{% raw %}</span>
+                      <span class="ap-m-lnk-text">{{_('Open example')}}</span>
                     </a>
-                    {{#playgroundUrl}}
-                    <a class="ap-o-search-result-linklist-item ap-o-search-result-linklist-item-playground ap-m-lnk" href="{{playgroundUrl}}" target="_blank">
+                    {{'{{#playgroundUrl}}'}}
+                    <a class="ap-o-search-result-linklist-item ap-o-search-result-linklist-item-playground ap-m-lnk" href="{{'{{playgroundUrl}}'}}" target="_blank">
                       <div class="ap-a-ico ap-m-lnk-icon">
                         <svg><use xmlns:xlink="http://www.w3.org/1999/xlink" xlink:href="#internal"></use></svg>
                       </div>
-                      <span class="ap-m-lnk-text">{% endraw %}{{_('Open in playground')}}{% raw %}</span>
+                      <span class="ap-m-lnk-text">{{_('Open in playground')}}</span>
                     </a>
-                    {{/playgroundUrl}}
+                    {{'{{/playgroundUrl}}'}}
                   </div>
-                  {{/exampleUrl}}
+                  {{'{{/exampleUrl}}'}}
                 </div>
-                {{/result.components}}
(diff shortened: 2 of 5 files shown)


matthiasrohmer:
Will have a detailed look tomorrow, but what do you think: are the styling glitches really related to this PR or its changes or have the glitches just become apparent by repeatedly testing this? 😉 Because then we would be good to merge nevertheless as it wouldn't make things worse.

frontend/templates/views/partials/header.j2
```diff
@@ -87,7 +87,7 @@
     </nav>
 
     {% do doc.styles.addCssFile('/css/components/molecules/search-trigger.css') %}
-    <div id="searchTriggerOpen" class="ap-m-search-trigger" on="tap:searchLightbox" role="button"  tabindex="0">
+    <div id="searchTriggerOpen" class="ap-m-search-trigger" on="tap:searchLightbox,AMP.setState({})" role="button"  tabindex="0">
```
matthiasrohmer: `AMP.setState({})` seems unnecessary as states will get deep merged and `bigObjWithThousandKeys + {}` is still `bigObjWithThousandKeys`.

frontend/templates/views/partials/search.j2
```diff
@@ -76,63 +76,61 @@
                     reset-on-refresh
                     single-item
           >
-            {% raw %}
```
matthiasrohmer: I recall you mentioned to me that `{% raw %}` doesn't work here anymore. Can you quickly recap why? Sorry I forgot 😉

pages/static/serviceworker.js
```diff
@@ -1,3 +1,4 @@
+importScripts('https://storage.googleapis.com/workbox-cdn/releases/4.3.1/workbox-sw.js');
```
matthiasrohmer: Can you please have another look if there's no way to reuse the Workbox version amp-sw uses internally? 🙂

pages/static/serviceworker.js
```diff
@@ -9,3 +10,67 @@ AMP_SW.init({
     assets: [],
   },
 });
+
+const CACHE_NAME = 'AMP-DEV-SEARCH-CACHE';
+const LATEST_QUERY_URL = '/search/latest-query';
+
+
+// Match search request url
+const matchDo = ({url}) => {
```
matthiasrohmer: Valid for all functions here: use real named functions, so this would become:

```js
function matchDo(url) {
}
```

Also why doing so try to find proper function names as `Do` and `Cq` are not really obvious for someone reading the code for the first time. 🙂 Maybe scope all names to something starting with `search`.

pages/static/serviceworker.js
sebastianbenz: ...and if that's not possible, I'd suggest implementing it without workbox as we don't really need it's features for what we try to accomplish.

pages/static/serviceworker.js
```diff
@@ -9,3 +10,67 @@ AMP_SW.init({
...
+
+  cache.put(LATEST_QUERY_URL, new Response(`"${searchQuery}"`));
+
+  let response = await cache.match(event.request);
+
+  // If search results are already in cache, skip
+  // Else fetch new search request and cache whole results
+  if (response) {
+    return response;
+  } else {
```
sebastianbenz: nit: `else` is no needed here => less indentation

pages/static/serviceworker.js
```diff
@@ -9,3 +10,67 @@ AMP_SW.init({
...
+const handlerCq = async () => {
+  try {
+    const cache = await caches.open(CACHE_NAME);
+    const response = await cache.match(LATEST_QUERY_URL);
+
+    // Update latest query
+    if (response) {
+      return response;
+    } else {
+      return new Response('null');
```
sebastianbenz: same as above, `else` not needed. Also: wouldn't an empty string be a better response?

frontend/templates/views/partials/header.j2
sebil: I implemented this to force a state update on opening the search layer. This is necessary for the prefilled value of the search input field, which should contain the latest search query.

frontend/templates/views/partials/search.j2
sebil: It still works with `{% raw %}` but it was very ugly and unreadable. I just wanted to clean it a little. 😊

pages/static/serviceworker.js
sebil: Workbox is not longer in use now, I've rewritten the `registerRoute()` functions. 👍

pages/static/serviceworker.js
sebil: Thanks for the feedback! Everything clean now :-)

pages/static/serviceworker.js
sebil: Thanks, learned something new 👍

pages/static/serviceworker.js
sebil: > Also: wouldn't an empty string be a better response?

An empty string wouldn't work, because then it'll fire a search request to `/search/do?q=&...` when opening the layer without a cached query, which causes an error.
With `null` no request will be fired.

sebil:
Thanks for your feedback, I implemented most of the requested changes and commented on them where it wasn't possible.

frontend/templates/views/partials/search.j2
```diff
@@ -76,63 +76,61 @@
...
             <template type="amp-mustache">
               <div class="ap-o-search-result-list">
-                {{^prevUrl}}
-                {{#result.components.length}}
-                <h4 class="ap-o-search-result-category">{% endraw %}{{_('Components')}}{% raw %}</h4>
-                {{/result.components.length}}
-                {{/prevUrl}}
-                {{#result.components}}
+                {{'{{^prevUrl}}'}}
```
sebastianbenz: I wanted to write: "can we avoid the double `{{` this looks tedious to maintain". However, looking at this again, I have to say that I prefer this over `raw` as well, as it's clear now which templates are rendered client-side.

Different tags would be even nicer though...but that would probably be a pain to refactor.

pages/static/serviceworker.js
```diff
@@ -9,3 +9,67 @@ AMP_SW.init({
...
+  try {
+    const cache = await caches.open(SEARCH_CACHE_NAME);
+    const response = await cache.match(SEARCH_LATEST_QUERY_PATH);
+
+    // Update latest query
+    if (response) {
+      return response;
+    }
+
+    return new Response('null');
```
sebastianbenz: two comments: 

* join the return statements into a single one
* isn't it better to return an empty string?
* please log the error to console

pages/static/serviceworker.js
```diff
@@ -9,3 +9,67 @@ AMP_SW.init({
...
+const SEARCH_LATEST_QUERY_PATH = '/search/latest-query';
+
+// Handle search request
+async function searchDoRequestHandler(url, request) {
+  // Extract query parameter from search request url
+  // Put new query as response to AMP-DEV-SEARCH-CACHE
+  const searchQuery = url.search.match(/q=([^&]+)/)[1];
+  const cache = await caches.open(SEARCH_CACHE_NAME);
+
+  cache.put(SEARCH_LATEST_QUERY_PATH, new Response(`"${searchQuery}"`));
```
sebastianbenz: Should we cache all search queries or wouldn't it be better to clear the cache before adding a new value to avoid stale results?

frontend/templates/views/partials/search.j2
sebil: I prefer it, too, because of two reasons:
1. The `raw` way had a reversed logic when it comes to the translation strings. So to print out `{{_('My localized string!')}}` it must be `{% endraw %}` right before the string and `{% raw %}` again right after the string, what makes it confusing when reading.
2. When you use a syntax highlighting theme for jinja templates, the whole `raw` section looks like commented out.

The screenshot show both problems:
Screenshot

I don't see any use of refactoring the tags, because they are the standard tags of jinja templating. After understanding that the `<amp-mustache>` variables have to be fully "printed" to the file and don't get rendered, it makes sense that we use ` {{'{{^prevUrl}}'}}` to print out a "string" in jinja.

pages/static/serviceworker.js
sebil: > isn't it better to return an empty string?

Sorry, I answered that question once before but resolved the conversation.
https://github.com/ampproject/amp.dev/pull/3277#discussion_r356788027

sebil: Fixed the other 2 things.

pages/static/serviceworker.js
sebil: Fixed that. Now the previous result cache will be deleted before a new cache is created.

sebil:
> Code looks good. Few nits.
> 
> When trying it locally, the previous results weren't displayed though:
> 
> Screen Shot 2019-12-13 at 10 52 39

This is the glitch I mentioned in the description. It happens only sporadically and I can't find the reason yet.

sebastianbenz:
Found another problem: the URL string needs to be URL decoded

Screen Shot 2019-12-13 at 13 44 31

sebastianbenz:
The glitch most likely results from some amp-state weirdness. The correct search request gets fired (and returned). It happens quite regularly to me though.

sebastianbenz:
Correction: seems like the amp-list height is not updated (the results are rendered on the page, but hidden). Maybe a race condition between `[src]` binding and `is-container-layout`.

sebil:
> Found another problem: the URL string needs to be URL decoded

My bad, only tested it with single words.
Fixed it now. Thanks!

sebil:
@sebastianbenz I've fixed the glitch by reducing amp-list and amp-mustache templates of the highlights and results-list to one. Feel free to test it again.

But before we going on here, we recognized that the user has no chance to get back the initial highlights list, once he has searched for something that has been cached.
To fix this, we want to clear the cache by clicking on the (X) button of the search layer. This should keep the user flow alive (browsing trough pages and continue using the search) but also gives availability to reset the search state intuitively.

What do you think?

btw: Happy new year  🎉

sebastianbenz:
Awesome! Works great now! 

One thing: the last query gets deleted when the search window gets opened - even if no new search is triggered. Instead I'd expect it to be only updated after a new search.

sebil:
> One thing: the last query gets deleted when the search window gets opened - even if no new search is triggered. Instead I'd expect it to be only updated after a new search.

The last query only gets deleted when you click the close button (X) in the upper right corner.
But if you switch the page, press ESC or click outside of the search layer, the last query and results will remain as before.

We think, that this is the most intuitive way to reset or quit a "search session" by the user to show the initial highlights page again. (Popular components etc.)

What exactly do you mean with "new search"?

sebil:
There was a little bug: when pressing the ESC key and open the search again, the input field was empty.
I fixed this now and optimized it by using the placeholder instead of the value attribute. It feels better when you want to start a new search and not have to delete the previous query first.

sebastianbenz:
I'd prefer to always leave the last search result. I'm not a fan of adding a second meaning to the close button which is imo not intuitive from a user perspective.

PR #3280, opened 2019-12-11
amp-list rendering from amp-state height fix.

samouri:
**summary**
The height of each row is actually only 22px.
See the "Rendering amp-list" [here](https://amp.dev/documentation/examples/components/amp-list/?format=websites#rendering-amp-state).

**screenshot**
Screen Shot 2019-12-11 at 1 36 36 PM

### DIFF
--- a/examples/source/1.components/amp-list.html
+++ b/examples/source/1.components/amp-list.html
@@ -145,7 +145,7 @@ The template can also be specified using an ID of an existing `template` element
   <!--
     ## Rendering `amp-state`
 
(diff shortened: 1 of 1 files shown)


CrystalOnScript:
Hi @samouri! Thanks for taking the initiative to fix this - however, the documentation is correct. While the height of the individual items is `22px`, the overall size of the `amp-list` component is calculated as `[height]="items.length * 24"`. 
Screen Shot 2019-12-11 at 11 57 21 AM

samouri:
@CrystalOnScript, I didn't just modify the documentation, I also modified the code to keep them in sync.

I think it is a bug that the actual item height and the height calculation are off by 2px, because that leads to an increasing margin each time one clicks `Add item.` For example, by the time that 40 items have been added, the "Add item" button becomes 80px too far down (2px per item).

See 
Screen Shot 2019-12-11 at 3 31 12 PM

CrystalOnScript:
@samouri - thanks for clarifying, I misunderstood!

PR #3290, opened 2019-12-15
✨ Clean up stale Google Cloud instance templates

matthiasrohmer:
This will help reduce costs for Google Cloud. Before creating a new instance template old ones (except the last three) will be deleted.

Side note: I configured Google Cloud Storage (which stores Travis artifacts) to auto-delete everything older than 7 days.

### DIFF
--- a/gulpfile.js/deploy.js
+++ b/gulpfile.js/deploy.js
@@ -158,6 +158,36 @@ function instanceTemplateCreate() {
 }
 
 /**
+ * Removes stale instances templates
+ */
+async function instanceTemplatesClean() {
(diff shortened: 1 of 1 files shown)


matthiasrohmer:
Can you specify which assets? 🙂

sebastianbenz:
The ones uploaded during the build.

>

matthiasrohmer:
Not sure if we're talking at cross purposes. 😁 The ones uploaded during build are stored in Google Cloud Storage and are now auto-deleted if older than 7 days - that's done by configuration without code changes. Are there other ones I missed?

Issue #3289, opened 2019-12-15
[Translate_ko] Beginners Course index page


jeongsd:
**Declare into which language you're translating the document.**
Korean

**Name which file you wish to translate or update.**
https://github.com/ampproject/amp.dev/blob/future/pages/content/amp-dev/documentation/courses/beginning-course/index.md

**Give a rough estimation of how much of the file you are translating.**
index.md

**Provide a general estimation of how much time it will take to complete the translation task.**
1 day

@CrystalOnScript

CrystalOnScript:
@jeongsd that would be great!

PR #3294, opened 2019-12-16
[Translate-ko] Add beginning-course index page

jeongsd:
https://github.com/ampproject/amp.dev/issues/3289


morsssss:
This is excellent! Thanks so much, @jeongsd !

PR #3304, opened 2019-12-18
🐛 Return absolute URLs for search results

matthiasrohmer:
... making them relative was only to have better DX while testing.

### DIFF
--- a/platform/lib/routers/search.js
+++ b/platform/lib/routers/search.js
@@ -114,9 +114,13 @@ function createPageObject(csePageItem) {
     url: csePageItem.link,
   };
 
-  const hostMatch = page.url.match(REMOVE_HOST_PATTERN);
-  if (hostMatch) {
-    page.url = page.url.substring(hostMatch[0].length);
+  // For better DX while testing the search make links relative
+  // to point to local documents
+  if (config.isDevMode() || config.isLocalMode()) {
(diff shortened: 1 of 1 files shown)


platform/lib/routers/search.js
```diff
@@ -114,9 +114,13 @@ function createPageObject(csePageItem) {
     url: csePageItem.link,
   };
 
-  const hostMatch = page.url.match(REMOVE_HOST_PATTERN);
-  if (hostMatch) {
-    page.url = page.url.substring(hostMatch[0].length);
+  // For better DX while testing the search make links relative
+  // to point to local documents
+  if (config.isDevMode() && config.isLocalMode()) {
```
sebastianbenz: should this be `||`?

matthiasrohmer: 🤦‍♂Thanks.

westonruter:
When will this be deployed into production?

sebastianbenz:
It's already live in prod.

>

westonruter:
@sebastianbenz OK, I can see it for the `/search/do` endpoint but not for `/search/highlights`. See https://amp.dev/search/highlights?locale=en

Relative URLs are still being served.

Also, for `/search/do` not all URLs are absolute yet. Namely, `nextUrl` and `exampleUrl` are missing the host.

Issue #2931, opened 2019-09-09
Compress courses images


sebastianbenz:
Some of them are up to 712kb.

DerekNonGeneric:
For your consideration:
- [pngcrush](https://www.npmjs.com/package/pngcrush)
- [rollup-plugin-imagemin](https://www.npmjs.com/package/rollup-plugin-imagemin)
- [imagemin-webpack-plugin](https://www.npmjs.com/package/imagemin-webpack-plugin)
- [Kraken.io](https://kraken.io/web-interface)

morsssss:
(shame)

@robocel had kindly offered to do this earlier this year. Rob, whaddya say?

morsssss:
This has been fixed via #3322 . We could probably compress some images more aggressively, but it's so much improved!

Issue #3244, opened 2019-11-27
Use srcset on for <amp-img>


kristoferbaxter:
# Is your feature request related to a problem? Please describe.
Images loaded on `amp.dev` do not use `srcset` to choose a smaller image for different screen sizes. Using srcset would reduce the number of bytes sent for image requests.

```html
<amp-img class="ap--media i-amphtml-layout-responsive i-amphtml-layout-size-defined" src="/static/img/case-band-image-2.jpg.webp" alt="A screenshot of a website using AMP." layout="responsive" width="519" height="931" i-amphtml-layout="responsive"><i-amphtml-sizer style="display:block;padding-top:179.3834%;"></i-amphtml-sizer>
  <amp-img fallback="" class="ap--media i-amphtml-layout-responsive i-amphtml-layout-size-defined" src="/static/img/case-band-image-2.jpg" alt="A screenshot of a website using AMP." layout="responsive" width="519" height="931" i-amphtml-layout="responsive">
    <i-amphtml-sizer style="display:block;padding-top:179.3834%;"></i-amphtml-sizer>
  </amp-img>
</amp-img>
```

Here using src set could reduce the size of these images significantly for smaller resolution devices.

# Describe the solution you'd like
Leverage srcset for the default `webp` image and the fallback `jpg`.

# Optional: Teachability, Documentation, Adoption, Migration Strategy
This relatively small change would have a significant impact on the number of bytes sent on the homepage of `amp.dev` and would provide material for a blog post.

DerekNonGeneric:
I would like to take care of this. Can someone please assign the issue to me?

CrystalOnScript:
Thanks @DerekNonGeneric - assigned!

sebastianbenz:
@DerekNonGeneric thanks for offering to look into this! 

Few comments about this:

* I'd like to avoid solving this in our build system. The problem is: our amp.dev build image is already 1GB big. Having images in multiple dimensions would further increase this.
* My preferred way to solve this would be:
   1. adding an [AMP Optimizer transformer](https://github.com/ampproject/amp-toolbox/tree/master/packages/optimizer/lib/transformers) that automatically generates secret tags if they don't exist yet.
   2. use either an existing Google Cloud image hosting solution or setup [Thumbor](https://github.com/thumbor/thumbor) on GCP.

These constraints probably make this non as straightforward to solve as it first seemed. I'm happy to support you if you're still willing to tackle this.

[...]

PR #3322, opened 2019-12-21
Compressing AMP course images

robocel:
Addresses #2931 

- Shrinks nearly every image in the course significantly
- Replaces smattering of WebP images with PNG images to be consistent with how all other course images are


robocel:
@morsssss - I'm honestly surprised we hadn't resolved this yet, but here are the image compressions. I'm happy to make any revisions that anyone has regarding this!

PS - I don't think I can make the Googlebot happy, because the other "author" it's picking up is an account not tied to any existing GitHub account.

DerekNonGeneric:
I don't think images should be manually optimized like this. I may be needing higher resolution images for #3244. Maybe I should take care of this with my PR for that issue as I'm going to need to add image manipulation/compression to the build system.

morsssss:
@robocel , true, unfortunately, the nature of my job is that really important things don't get done, because I need to do theoretically even more really important things first! Thanks for doing this!

That said, @DerekNonGeneric , if you get `srcset` going on amp.dev, yes, it would be excellent to provide each of these images in 2-3 different sizes instead.

morsssss:
Also, @robocel , sorry about the continued obstacle posed by the CLA. As we move over to OpenJS, this will get simpler!

kompassriwijaya:
[transaction-report.zip](https://github.com/ampproject/amp.dev/files/4014513/transaction-report.zip)

kompassriwijaya:
[crash.log](https://github.com/ampproject/amp.dev/files/4014516/crash.log)

matthiasrohmer:
@sebastianbenz, can you verify the CLA for this? I think https://github.com/ampproject/amp.dev/pull/3322#issuecomment-568141597 expresses consent and this would be good to have in 🙂

sebastianbenz:
Done

morsssss:
@matthiasrohmer , thanks for noticing this!!

PR #3327, opened 2019-12-23
🚸 Do not enforce websites as default filter format and add filter hint

### LINKED ISSUE #2588: Respect format in canonical links
via @tharders: our `<link rel="canonical">` elements currently don't share a format GET parameter and therefore always fallback to *websites*.

This is wrong for documents that shouldn't actually be available for *websites*.

matthiasrohmer:
Fixes #2588. A common way to come to amp.dev is through Google search. As our canonical links don't have a `?format` parameter we defaulted to `websites` which had weird UX effects: a search for `amp-story-grid-layer` would bring you to the correct page but with a hint saying:

>Important: this component does not support your currently selected format websites!

This PR enhances UX by using the first format the visited document is available in as default and additionally adds a more prominent hint to filter the page with `amp-user-notification`.

Bildschirmfoto 2019-12-23 um 15 21 24


frontend/templates/views/partials/format-filter-hint.j2
```diff
@@ -0,0 +1,29 @@
+{% if doc.formats %}
+{% do doc.styles.addCssFile('css/components/molecules/format-filter-hint.css') %}
+
+{% do doc.amp_dependencies.add('amp-user-notification', '0.1') %}
+<amp-user-notification class="ap-m-format-filter-hint" id="format-filter-hint" layout="nodisplay">
+  <div class="ap-m-format-filter-hint-content">
+    <div class="ap-m-format-filter-hint-note">Select your format for a more streamlined experience</div>
```
sebastianbenz: @CrystalOnScript PTAL at the wording

pages/content/amp-dev/documentation/guides-and-tutorials/start/test.md
```diff
@@ -0,0 +1,26 @@
+---
+$title: Format test
```
sebastianbenz: Is this intended to be committed?

platform/lib/templates/SupportedFormatsExtension.js
```diff
@@ -0,0 +1,58 @@
...
+  run(context, args) {
+    const formats = args.formats || [];
+
+    // Check if the user did not select a specific format
+    // and if so overwrite the default format with a more fitting one
+    if (context.ctx.requestedFormat) {
+      return;
+    }
+
+    context.ctx.format = formats[0];
```
sebastianbenz: What about the case when there are no formats given?

platform/lib/templates/SupportedFormatsExtension.js
```diff
@@ -0,0 +1,58 @@
...
+
+    return new nodes.CallExtension(this, 'run', args);
+  }
+
+  /**
+   * Tries to figure out if there is a fitting default
+   * format and overwrites the one set by the global context
+   */
+  run(context, args) {
+    const formats = args.formats || [];
```
sebastianbenz: I don't quite understand why we need an extension for this?

matthiasrohmer: We don't need one necessarily. An equivalent solution would be to put the logic into the template:

```
{% if doc.formats %}
[% if not requestedFormat %]
[% set supportedFormats = {{ doc.formats|tojson }} %]
[% set format = supportedFormats[0] || 'websites' %]
[% endif %]
{% endif %}
```

platform/lib/templates/SupportedFormatsExtension.js
matthiasrohmer: Templating handled an undefined format gracefully, but now explicitly falls back to websites again.

pages/content/amp-dev/documentation/guides-and-tutorials/start/test.md
matthiasrohmer: No, thanks!

frontend/templates/views/partials/format-filter-hint.j2
sebastianbenz: ping

Issue #3330, opened 2019-12-24
Issues with building the website locally


fstanis:
I wanted to get some high level discussion started around this, as I've heard feedback about it on both Slack and during the Contributor Summit. And I also struggled with this personally on many occasions.

I spent the last few hours trying the developer experience fresh (inside an empty Docker container, to make sure I get a clear slate). Currently, it's very difficult to get the project running locally.

This doesn't work:

```bash
npm install
npm run develop
# error: dist/static/files/component-versions.json not found
```

It seems bootstrapping is required (which isn't clear in the README or in the error message), but this, too, fails:

```bash
npm install
npm run bootstrap
npm run develop
# error: dist/static/samples/samples.json not found
```

The correct commands to run are this:
```bash
npm install
npm run bootstrap
npx gulp buildSamples
npm run develop
```

What also works is running `npm run develop` twice.

The main issues here are:
-   Seems there's a race condition in `gulp develop` that prevents `samples.json` from being generated in time?
-   Requiring a bootstrap means we require `AMP_DOC_TOKEN` which is not hard, but raises the bar to entry. Users unfamiliar with access tokens or bash will struggle with this (e.g. how to make that variable persist and the implications of storing it locally).
-   Most of the error messages are unforgiving and confusing. Examples:
    -   The issues with `samples.json` or `component-versions.json` missing provide no additional information about how these files should be created / obtained.
    -   Not having an `AMP_DOC_TOKEN` set gives a clear error and points the user to the README. However, if the token is invalid, the error is `TypeError: Cannot read property 'tag_name' of undefined`.
    -   Anything going south with `gulp` will display `Did you forget to signal async completion?` with the actual error somewhere above.
    -   Grow errors are especially vague, e.g. `Error loading /content/amp-dev/documentation/examples/documentation/Stories_in_AMP.html` followed by a very long python stack trace.

Other issues:

-   README.md is currently *very* verbose. There's instructions for all sorts of scenarios and things to do and this is overwhelming, especially for new contributors.
    -   For example, there's a very long section explaining how to install node and python on different platforms. Do we really need this? Can we link to an external guide about this? If not, can we at least move it to a separate file?
-    `git config user.name` is called from [here](https://github.com/ampproject/amp.dev/blob/3be347aab4294d5790e83f31202b5aa99db0244b/platform/lib/utils/git.js#L21). This command fails if the username is not set, resulting in a confusing error when `npm install` is ran.
[...]

Issue #3023, opened 2019-09-25
Component docs: show latest prod version by default


sebastianbenz:
For example, the new amp-sidebar v1.0 docs should not be shown by default while the new version is still experimental. 

However, the should be linked from the older doc versions.

//cc @nainar in case we'd prefer to show the experimental version by default.
//cc @tharders

nainar:
I agree with you - experimental docs can be found via github.

patrickkettner:
@sebastianbenz is this still valid? Unless I am misunderstanding, the bento changes earlier this year sorted this?

sebastianbenz:
Correct. Thanks for checking in.

>

PR #3344, opened 2020-01-06
✨ Add new component intro table

robinvanopstal:
This adds all the needed functionality to show the new automatically generated component intro tables by using the data provided by [@ampproject/amp-toolbox/validator-rules](https://github.com/ampproject/amp-toolbox/tree/master/packages/validator-rules).

### Before
Bildschirmfoto 2020-01-06 um 21 29 59

### After
Bildschirmfoto 2020-01-06 um 21 29 51

Additionally this refactors the importer to be around 20x faster than before and moves all document related logic to a separate class.

Open tasks: design review with Longo, checking why the GrowReferenceChecker has problems with the new files, removing the manually maintained tables from the documents over at ampproject/amphtml and add missing teaser where needed.


matthiasrohmer:
@googlebot I consent.

platform/lib/build/samplesBuilder.js
```diff
@@ -570,6 +584,23 @@ class SamplesBuilder {
       }
     });
 
+    // Store the sample by it's used component to show all samples for a specific
```
tharders: A method that is named get... should never modify the state.
I know the code snipped above did already modify the parsedSample which is also bad.
I think we should refactor this.
Maybe call some methods in _createDocumentation that do the changes and then call getTeaserData with the usedComponents.

platform/lib/pipeline/componentReferenceDocument.js
```diff
@@ -0,0 +1,116 @@
...
+const EXTENSION_TYPE_ELEMENT = 'element';
+const EXTENSION_TYPE_TEMPLATE = 'template';
+
+
+class ComponentReferenceDocument extends MarkdownDocument {
+  constructor(path, contents, extension) {
+    super(path, contents);
+
+    if (!this.teaser.text) {
+      throw new Error('Teaser missing in', extension.name);
```
tharders: Was it discussed that documents without a teaser text should be skipped completely?

platform/lib/pipeline/gitHubImporter.js
```diff
@@ -63,6 +63,21 @@ class GitHubImporter {
     return this.fetchContents_(filePath, repo, master);
   }
 
+  /**
+   * Fetches all files contained in a directory flattened down to
```
tharders: The comment suggests the file contens are returned rather than the file path.

frontend/templates/views/partials/component-intro.j2
```diff
@@ -0,0 +1,61 @@
...
+    <div class="ap-o-component-intro-box-scripts-item">
+      <code>{{ script }}</code>
+      <amp-iframe
+          width="24"
+          height="24"
+          layout="fixed"
+          sandbox="allow-scripts"
+          src="{{ podspec.base_urls.platform }}/shared/copy-script#{{ script|urlencode }}">
+          <div placeholder>
+            ..
```
sebastianbenz: a better way would be to use a disabled version of the copy icon as placeholder

frontend/templates/views/partials/component-intro.j2
```diff
@@ -0,0 +1,61 @@
+{% do doc.styles.addCssFile('css/components/molecules/code-snippet.css') %}
+{% do doc.amp_dependencies.add('amp-fx-collection', '0.1') %}
+
+<div class="ap-o-component-intro">
+  <h2 class="ap-o-component-intro-headline">Description</h2>
```
sebastianbenz: Not sure we need the header "Description" here

pages/content/shared/copy-script.html
```diff
@@ -0,0 +1,44 @@
...
+    html, body { background: none transparent; }
+    body { display: flex; min-height: 100vh; align-items: center; }
+    svg { fill: #005af0; }
+    input { opacity: 0 }
+  </style>
+</head>
+<body>
+  <svg style="position: absolute; width: 0; height: 0; overflow: hidden;" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
+    <defs>
+      {{ doc.icons.emit() }}
```
sebastianbenz: Are these all icons or only the copy one?

pages/content/shared/copy-script.html
```diff
@@ -0,0 +1,44 @@
...
+    </defs>
+  </svg>
+  <div class="ap-a-ico">
+    {% do doc.icons.useIcon('/icons/clipboard.svg') %}
+    <svg><use xmlns:xlink="http://www.w3.org/1999/xlink" xlink:href="#clipboard"></use></svg>
+  </div>
+  <input id="script-tag">
+  <script>
+    const $scriptTag = document.getElementById('script-tag');
```
sebastianbenz: don't use `$` as prefix

pages/content/shared/copy-script.html
```diff
@@ -0,0 +1,44 @@
...
+    const scriptTag = decodeURIComponent(window.location.hash.substr(1));
```
sebastianbenz: rename to `stringToCopy`

platform/lib/build/samplesBuilder.js
```diff
@@ -558,7 +572,7 @@ class SamplesBuilder {
    * @param  {Object} parsedSample
    * @return {Object}
    */
-  _getUsedComponents(parsedSample) {
+  _getUsedComponents(sample, parsedSample) {
```
sebastianbenz: please extract this into a separate file as this might be useful in other places as well.

platform/lib/pipeline/componentReferenceImporter.js
```diff
@@ -15,289 +15,220 @@
  */
 require('module-alias/register');
 
-// By default, reference docs are imported from the latest release tag.
-// If a doc is broken in a release, add it to this this list to fetch from master instead.
-//
-// DON'T FORGET TO REMOVE ONCE IT'S FIXED
-const DOCS_TO_FETCH_FROM_MASTER = ['amp-script', 'amp-carousel'];
```
sebastianbenz: Where did this go?

sebastianbenz:
Noticed a few more things:

* the copy button should use `cursor: pointer;` 
* vertical pre padding not consistent for scripts
Screen Shot 2020-01-08 at 13 48 45

* layouts are not aligned with heading

Screen Shot 2020-01-08 at 13 48 41
* Copy icon cut off on mobile 
Screen Shot 2020-01-08 at 13 52 00
* two column layout does not work on mobile
Screen Shot 2020-01-08 at 13 52 33
* layout page does not work

```

[TIMING] localhost:8080/documentation/guides-and-tutorials/learn/amp-html-layout/layouts_demonstrated/?format=websites 0.086s [200]
500: /documentation/guides-and-tutorials/learn/amp-html-layout/layouts_demonstrated.html - 'grow.documents.document.Document object' has no attribute 'teaser'
Traceback (most recent call last):
  File "/Users/sbenz/Library/Python/2.7/lib/python/site-packages/grow/server/main.py", line 28, in __call__
    return self.wsgi_app(environ, start_response)
  File "/Users/sbenz/Library/Python/2.7/lib/python/site-packages/grow/server/main.py", line 70, in wsgi_app
    response = self.dispatch_request(request)
  File "/Users/sbenz/Library/Python/2.7/lib/python/site-packages/grow/server/main.py", line 66, in dispatch_request
    return handlers.serve_pod(self.pod, request, matched)
  File "/Users/sbenz/Library/Python/2.7/lib/python/site-packages/grow/server/handlers.py", line 108, in serve_pod
    rendered_document = controller.render(jinja_env=jinja_env, request=request)
  File "/Users/sbenz/Library/Python/2.7/lib/python/site-packages/grow/rendering/render_controller.py", line 246, in render
    raise exception
BuildError: 'grow.documents.document.Document object' has no attribute 'teaser'
```
* The example list should always show the component specific example first (e.g. `amp-analytics`). Can we find a better name for the component specific samples? E.g.: `Component samples: amp-analytics`

pages/content/shared/copy-script.html
matthiasrohmer: Just curious: why? I'm used to prefixing variables containing DOM elements with $ to know that the stored object has DOM API methods.

sebastianbenz:
What's the status here?

LongoOfficial:
@sebastianbenz There was one issue on smaller screens where the layout draws too many edges on a small space. We would suggest to adjust the padding of the intro background on small screen sizes so it goes from edge to egde. This will solve the visual nesting problem without breaking the whole design system. See attached screenshot. wdyt?
localhost_8080_documentation_components_amp-carousel__format=websites(iPhone X) (1)

sebastianbenz:
@LongoOfficial can you share a full mock that includes normal text as well?

LongoOfficial:
@sebastianbenz 
localhost_8080_documentation_components_amp-carousel__format=websites(iPhone X) (1)

LongoOfficial:
oh! and for the Component Version UI in the title, we imagined something like this:

50157f2e-38e1-4f1a-b1c8-397272f3e41a
70f8735b-e02f-485a-94d1-eae8d72b9d64

sebastianbenz:
+1 to the version UI

re margins: can we half these on mobile? I still think they're too big (for normal text). This probably means that we have to move the main content onto the background, but would also help with our problem of having too many layers.

LongoOfficial:
let's discuss it separately, as this will affect more than the component intro. I'll join the status call on monday.

platform/lib/pipeline/componentReferenceDocument.js
matthiasrohmer: As before a teaser text is now automatically extracted.

matthiasrohmer:
@sebastianbenz, to quickly give you an update: there were some problems with `MarkdownDocument` implicitly altering `MarkdownDocument._contents` which are now resolved and make things a lot easier to grasp. While doing so I also introduced some more tests for all the rewriting stuff to be on the safe side. The new version toggle is also implemented.

A problem that still persists is that sometimes downloading files randomly fails with:

```
[6:51:27 PM] [ComponentReferenceImporter] › ✖  error     Failed to fetch extensions/amp-web-push/amp-web-push.md { Error: getaddrinfo ENOTFOUND api.github.com api.github.com:443
    at GetAddrInfoReqWrap.onlookup [as oncomplete] (dns.js:56:26)
  cause:
   { Error: getaddrinfo ENOTFOUND api.github.com api.github.com:443
       at GetAddrInfoReqWrap.onlookup [as oncomplete] (dns.js:56:26)
     errno: 'ENOTFOUND',
     code: 'ENOTFOUND',
     syscall: 'getaddrinfo',
     hostname: 'api.github.com',
     host: 'api.github.com',
     port: 443 },
  isOperational: true,
  errno: 'ENOTFOUND',
  code: 'ENOTFOUND',
  syscall: 'getaddrinfo',
  hostname: 'api.github.com',
  host: 'api.github.com',
  port: 443 }
```

This could be due to rate limiting which I'll check tomorrow. 

For all comments related to the `SamplesBuilder` I'll create a separate issue as it needs some refactoring anyway.

The plan is to create a PR towards the ampproject/amphtml repo tomorrow and finalise the last bits that are missing in here to then finally have this going 🙂

matthiasrohmer:
Final testing in progress, related PR issued: https://github.com/ampproject/amphtml/pull/26383.

matthiasrohmer:
Tests went well, comments in the related PR are taken care of today. I'm going to reimplement the Pull-from-master logic today and also stumbled over https://github.com/ampproject/amp.dev/issues/3023 which I want to double check.

pages/content/shared/copy-script.html
matthiasrohmer: This will only emit icons added somewhere else in the document by `{% do doc.icons.useIcon %}` - so yes only the clipboard icon.

matthiasrohmer:
@sebastianbenz, could you please review this as I've also contributed? I'll fix the failing teaser text test tomorrow but besides that, this should be finally good to go and would be great to have out the door also as part of #3330.

platform/lib/pipeline/componentReferenceImporter.js
matthiasrohmer: Moved to a configuration file to be aligned with the behaviour of SpecImporter: platform/config/imports/componentReference.json

frontend/templates/views/partials/component-intro.j2
```diff
@@ -0,0 +1,60 @@
+{% do doc.styles.addCssFile('css/components/molecules/code-snippet.css') %}
+{% do doc.amp_dependencies.add('amp-fx-collection', '0.1') %}
```
sebastianbenz: why is this needed here?

platform/lib/build/samplesBuilder.js
```diff
@@ -383,13 +386,28 @@ class SamplesBuilder {
...
+
+      await writeFileAsync(COMPONENT_SAMPLES_DEST, JSON.stringify(this._componentSamples), {
+        flag: 'w+',
       });
-      this._log.success('Wrote sample sitemap.');
+
+      this._log.success('Wrote sample sitemap and component samples file.');
     } catch (_) {
-      this._log.info('Samples sitemap already exists');
+      this._log.error('Writing samples builder meta files failed:', e);
```
sebastianbenz: I know this was the same before, but wouldn't it be better to fail the build here?

platform/lib/build/samplesBuilder.js
sebastianbenz: ping

platform/lib/build/samplesBuilder.js
sebastianbenz: ping

sebastianbenz:
What's the status of this?

Issue #3214, opened 2019-11-15
Enable experimental toggle on documentation pages that have experimental components 


CrystalOnScript:
## Feature Request
Experimental components do not execute in inline executable code samples on non-example documentation pages, such as components.

[amp-mega-menu](https://amp.dev/documentation/components/amp-mega-menu/?format=websites)

# Describe the solution you'd like
I'd like to see experimental inline executable code samples :D

cc: @tharders

PR #3354, opened 2020-01-09
✨ Enable experiment toggle on component pages

sebil:
With this PR we could turn on experimental features for inline example previews on component detail pages.

The look is exactly the same as on the example pages, but the toggle itself is embedded with an iframe, like on the standalone preview pages, so you have to manually reload the page after toggling (see screenshots below).

In case there are no example previews, I've adjusted the text a little bit to be more general.

components-experimental-toggle
components-experimental-toggle-reload

Closes: #3214


frontend/templates/views/partials/content.j2
```diff
@@ -36,6 +36,23 @@
   {% include '/views/partials/level-toggle.j2' %}
   {% endif %}
 
+  {# Render widget to toggle experiments if there are any#}
```
sebastianbenz: do we need to import amp-iframe here?

frontend/templates/views/partials/content.j2
```diff
@@ -36,6 +36,23 @@
...
+  {% if doc.experimental %}
+  <div class="ap-m-experiments">
+    <h4>Experimental</h4>
+    <p>{{ _('This page could contain example previews with experimental features. Enable the experiment via the button below. Some components require the AMP Dev Channel to be enabled as well.') }} Learn more about <a href="{{ g.doc('/content/amp-dev/documentation/guides-and-tutorials/learn/experimental.md', locale=doc.locale).url.path }}">experimental features</a>.</p>
```
sebastianbenz: Can we show this only when there are actual samples on the page?

frontend/templates/views/partials/content.j2
```diff
@@ -36,6 +36,23 @@
...
+    <amp-iframe layout="fixed-height"
+                height="66"
+                sandbox="allow-scripts allow-forms allow-same-origin allow-popups allow-popups-to-escape-sandbox allow-presentation allow-top-navigation"
+                allowfullscreen
+                allowpaymentrequest
+                frameborder="0"
+                src="{{ podspec.base_urls.preview }}/shared/experiments/?experiments={{doc.component}}&style=default">
```
sebastianbenz: Does this work for inline examples which are served from `amp.dev` vs `preview.amp.dev`?

frontend/templates/views/partials/content.j2
sebil: There was no error, but I imported it anyways now.

frontend/templates/views/partials/content.j2
sebil: I don't think so, because the samples are written down as shortcodes in the content files. So we don't have any indicator, that says us if any or how many `[example ...]` shortcodes are in the content. That's why I adjusted the text.

Do you have an idea?

frontend/templates/views/partials/content.j2
sebil: I tested it and everything worked perfectly with examples from preview.amp.dev or rather localhost:8084 (preview port).
Am I understanding you right, that there are also examples that are not served from the preview?

frontend/templates/views/partials/content.j2
sebastianbenz: The ref doc importer could add a flag if the page contains interactive samples. @matthiasrohmer wdyt?

frontend/templates/views/partials/content.j2
sebastianbenz: Yes, pages with inline preview.

matthiasrohmer:
@sebastianbenz, here's the latest mock from Longo's shop. WDYT?

Artboard

sebastianbenz:
> here's the latest mock from Longo's shop. WDYT?

I think that's the right direction, few things I don't like:

* the copy button gets a little bit lost, can we make the experiment indicator a similar icon
* the experiment dialog should be a button and not a toggle
* the dialog is very heavy, maybe add a bit margin

LongoOfficial:
experiment

@sebastianbenz 
wouldn't make the experiment indicator a similar icon, as it does not have a functionality.
I had the copy button much more prominent at first, but it was too distracting. It's a short cut. Imagine it on every code block, it should be as unobtrusive as possible

sebastianbenz:
@LongoOfficial OK - sgtm

PR #3374, opened 2020-01-13
Grammar changes to <amp-script> docs

morsssss:


morsssss:
We should also modify the bit about the height restriction to reflect https://github.com/ampproject/amphtml/issues/26047 .

morsssss:
I didn't dare just remove that note on the height restriction, because it does still seem to apply in some cases...

CrystalOnScript:
You're right, I moved too quickly on that one. Giving it another pass.

morsssss:
Looking at [the code](https://github.com/ampproject/amphtml/pull/26066/files#diff-2309faf5b32ca308fd2bb6e38e44d29e), my impression is now that there's no height limit. It's just that we don't allow mutations on containers that can change size... is that it?

CrystalOnScript:
From what I read, I think you are right @morsssss - I gave it an update. Thanks!

CrystalOnScript:
@kristoferbaxter Thanks for taking the time to update this! Much appreciated!

PR #3379, opened 2020-01-15
Add new Interactivity & Dynamic Content Restrict Duplicate Selection

elisameyer:
Adding a html file and json file showing how to restrict duplicate selection of items in amp-list. One of the amp-snippets from glitch project [here](https://amp-snippets.glitch.me/)

This snippet uses 'amp-selector' in combination with 'amp-bind' to check whether or not an item is currently in the list (populated by amp-state 'Fruit_Array_holder'). If fruit selected is already in the list nothing changes, if not selected fruit is added to end of list.

### DIFF
--- a/examples/source/interactivity-dynamic-content/Restrict_Duplicate_Selection.html
+++ b/examples/source/interactivity-dynamic-content/Restrict_Duplicate_Selection.html
@@ -1,0 +1,92 @@
+<!---
+author: elisameyer
+formats:
+  - websites
+tags:
+  - dynamic-content
+--->
+
+<!-- ## Introduction -->
+<!--
+This is a sample showing how to restrict duplicate selection of items in amp-list. This uses
+'amp-selector' in combination with 'amp-bind' to check whether or not an item is currently in
+the list (populated by amp-state 'Fruit_Array_holder'). If fruit selected is already in the list
+nothing changes, if not selected fruit is added to end of list.
+-->
+<!-- -->
+<!doctype html>
+<html ⚡>
+  <head>
+    <meta charset="utf-8">
+    <title>Restrict Duplicate Selection</title>
+    <link rel="canonical" href="<% canonical %>">
+    <meta name="viewport" content="width=device-width, initial-scale=1" />
+    <script async="" src="https://cdn.ampproject.org/v0.js"></script>
+     <!-- ## Setup -->
+    <!-- Import the amp-bind component to hold the selected fruits -->
+    <script async custom-element="amp-bind" src="https://cdn.ampproject.org/v0/amp-bind-0.1.js"></script>
+    <!-- Import the amp-list component to import the list of fruits to select from -->
+    <script async custom-element="amp-list" src="https://cdn.ampproject.org/v0/amp-list-0.1.js"></script>
+    <!-- Import the amp-selector component so user can choose fruits -->
+    <script async custom-element="amp-selector" src="https://cdn.ampproject.org/v0/amp-selector-0.1.js"></script>
+    <!-- The amp-mustache component is required if using amp-selector with amp-list -->
+    <script async custom-template="amp-mustache" src="https://cdn.ampproject.org/v0/amp-mustache-0.2.js"></script>
+    <style amp-boilerplate>body{-webkit-animation:-amp-start 8s steps(1,end) 0s 1 normal both;-moz-animation:-amp-start 8s steps(1,end) 0s 1 normal both;-ms-animation:-amp-start 8s steps(1,end) 0s 1 normal both;animation:-amp-start 8s steps(1,end) 0s 1 normal both}@-webkit-keyframes -amp-start{from{visibility:hidden}to{visibility:visible}}@-moz-keyframes -amp-start{from{visibility:hidden}to{visibility:visible}}@-ms-keyframes -amp-start{from{visibility:hidden}to{visibility:visible}}@-o-keyframes -amp-start{from{visibility:hidden}to{visibility:visible}}@keyframes -amp-start{from{visibility:hidden}to{visibility:visible}}</style><noscript><style amp-boilerplate>body{-webkit-animation:none;-moz-animation:none;-ms-animation:none;animation:none}</style></noscript>
+  </head>
+  <body>
+    <h2> Restrict Duplicate Selection Example </h2>
+    <!-- ## Implementation -->
+    <!--
+      We use the `<amp-state>` element, list initially empty but will update once user selects fruits, to hold the selected fruits. User selects fruits from '<amp-list>' element fruit_list.
+
+      Every time the user tries to select a fruit the amp-selector element will validate whether or not fruit is already in Fruit_Array_holder.
+      If fruit is not already in the list, amp-selector will change the state of the Fruit_Array_holder with new fruit at end of list. If fruit is in Fruit_Array_holder state stays the same.
+    -->
+    <main class="restrict-duplicate-selection-example">
+    <amp-state id="Fruit_Array_holder" hidden="" aria-hidden="true">
+      <script type="application/json">
+        []
+      </script>
+    </amp-state>
+    <p>
+      Pick your favorite fruites from this list <br> You can only add new fruits to the list, you can't add the same fruit
+      multiple times.
+    </p>
+    <div style="resize: horizontal; overflow: auto;">
+      <p [text]="Fruit_Array_holder">Select fruits!</p>
+    </div>
+    <div id="fruit_list">
+      <amp-list
+        src="/static/samples/json/fruits.json"
+        layout="fixed-height"
+        width="auto"
+        height="900"
+        items="."
+        single-item
+        id="fruit_data_list"
+      >
+        <template type="amp-mustache">
+          <ul>
+            <amp-selector
+              role="tablist"
+              on="select:AMP.setState({defState:{fruit_list: 0, show_blocks: 1, inputValue: ''}})"
+              tabindex="0"
+            >
+              {{#items}}
+              <li
+                option="{{name}}"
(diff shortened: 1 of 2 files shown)


examples/source/interactivity-dynamic-content/Restrict_Duplicate_Selection.html
```diff
@@ -0,0 +1,92 @@
...
+        height="900"
+        items="."
+        single-item
+        id="fruit_data_list"
+      >
+        <template type="amp-mustache">
+          <ul>
+            <amp-selector
+              role="tablist"
+              on="select:AMP.setState({defState:{fruit_list: 0, show_blocks: 1, inputValue: ''}})"
```
sebastianbenz: tip: you can use line breaks to make actions more readable

```
on="select:AMP.setState({
  defState: {
      fruit_list: 0, 
      show_blocks: 1, 
      inputValue: ''
  }
})"
```

examples/source/interactivity-dynamic-content/Restrict_Duplicate_Selection.html
```diff
@@ -0,0 +1,92 @@
...
+formats:
+  - websites
+tags:
+  - dynamic-content
+--->
+
+<!-- ## Introduction -->
+<!--
+This is a sample showing how to restrict duplicate selection of items in amp-list. This uses
+'amp-selector' in combination with 'amp-bind' to check whether or not an item is currently in
```
sebastianbenz: You can use ` to code highlight comments, e.g. `amp-list`:

```
This is a sample showing how to restrict duplicate selection of items in `amp-list`....
```

It'd be great if you could do this every time you mention an AMP component or a piece of code (e.g. variable names).

examples/source/interactivity-dynamic-content/Restrict_Duplicate_Selection.html
```diff
@@ -0,0 +1,92 @@
...
+    <h2> Restrict Duplicate Selection Example </h2>
+    <!-- ## Implementation -->
+    <!--
+      We use the `<amp-state>` element, list initially empty but will update once user selects fruits, to hold the selected fruits. User selects fruits from '<amp-list>' element fruit_list.
+
+      Every time the user tries to select a fruit the amp-selector element will validate whether or not fruit is already in Fruit_Array_holder.
+      If fruit is not already in the list, amp-selector will change the state of the Fruit_Array_holder with new fruit at end of list. If fruit is in Fruit_Array_holder state stays the same.
+    -->
+    <main class="restrict-duplicate-selection-example">
+    <amp-state id="Fruit_Array_holder" hidden="" aria-hidden="true">
```
sebastianbenz: nit: for ids please use lowercase letters + dashes: `fruit-array-holder`. I'd also suggest renaming the amp-state to simply: `fruits`. You also don't need to explicitly set `hidden="" aria-hidden="true"` as this gets added automatically.

examples/source/interactivity-dynamic-content/Restrict_Duplicate_Selection.html
```diff
@@ -0,0 +1,92 @@
...
+  <body>
+    <h2> Restrict Duplicate Selection Example </h2>
+    <!-- ## Implementation -->
+    <!--
+      We use the `<amp-state>` element, list initially empty but will update once user selects fruits, to hold the selected fruits. User selects fruits from '<amp-list>' element fruit_list.
+
+      Every time the user tries to select a fruit the amp-selector element will validate whether or not fruit is already in Fruit_Array_holder.
+      If fruit is not already in the list, amp-selector will change the state of the Fruit_Array_holder with new fruit at end of list. If fruit is in Fruit_Array_holder state stays the same.
+    -->
+    <main class="restrict-duplicate-selection-example">
```
sebastianbenz: tip: change this to a `div` without class. This will then get removed automatically in the generated sample page.

examples/source/interactivity-dynamic-content/Restrict_Duplicate_Selection.html
```diff
@@ -0,0 +1,92 @@
...
+      </script>
+    </amp-state>
+    <p>
+      Pick your favorite fruites from this list <br> You can only add new fruits to the list, you can't add the same fruit
+      multiple times.
+    </p>
+    <div style="resize: horizontal; overflow: auto;">
+      <p [text]="Fruit_Array_holder">Select fruits!</p>
+    </div>
+    <div id="fruit_list">
```
sebastianbenz: do we need this div or could we remove it to keep things simple?

examples/source/interactivity-dynamic-content/Restrict_Duplicate_Selection.html
```diff
@@ -0,0 +1,92 @@
...
+      <amp-list
+        src="/static/samples/json/fruits.json"
+        layout="fixed-height"
+        width="auto"
+        height="900"
```
sebastianbenz: use the actual height of `242px`

examples/source/interactivity-dynamic-content/Restrict_Duplicate_Selection.html
```diff
@@ -0,0 +1,92 @@
...
+      multiple times.
+    </p>
+    <div style="resize: horizontal; overflow: auto;">
+      <p [text]="Fruit_Array_holder">Select fruits!</p>
+    </div>
+    <div id="fruit_list">
+      <amp-list
+        src="/static/samples/json/fruits.json"
+        layout="fixed-height"
+        width="auto"
```
sebastianbenz: you don't need to explicitly define this

examples/source/interactivity-dynamic-content/Restrict_Duplicate_Selection.html
```diff
@@ -0,0 +1,92 @@
...
+        height="900"
+        items="."
+        single-item
+        id="fruit_data_list"
+      >
+        <template type="amp-mustache">
+          <ul>
+            <amp-selector
+              role="tablist"
```
sebastianbenz: Not sure this is the right `role` here, afaik this is used for tabbed UIs. (I might be wrong)

examples/source/interactivity-dynamic-content/Restrict_Duplicate_Selection.html
sebastianbenz: on a second pass: is this needed at all?

examples/source/interactivity-dynamic-content/Restrict_Duplicate_Selection.html
```diff
@@ -0,0 +1,92 @@
...
+              on="select:AMP.setState({defState:{fruit_list: 0, show_blocks: 1, inputValue: ''}})"
+              tabindex="0"
+            >
+              {{#items}}
+              <li
+                option="{{name}}"
+                on="tap:AMP.setState({Fruit_Array_holder: Fruit_Array_holder.includes('{{name}}') ? Fruit_Array_holder: Fruit_Array_holder.concat(['{{name}}'])})"
```
sebastianbenz: tip: move this into the amp-selector's event handler to reduce markup.

examples/source/interactivity-dynamic-content/Restrict_Duplicate_Selection.html
```diff
@@ -0,0 +1,92 @@
...
+    <amp-state id="Fruit_Array_holder" hidden="" aria-hidden="true">
+      <script type="application/json">
+        []
+      </script>
+    </amp-state>
+    <p>
+      Pick your favorite fruites from this list <br> You can only add new fruits to the list, you can't add the same fruit
+      multiple times.
+    </p>
+    <div style="resize: horizontal; overflow: auto;">
```
sebastianbenz: do we need the styles here or could we simplify this to a simple single div?

examples/source/interactivity-dynamic-content/Restrict_Duplicate_Selection.html
```diff
@@ -0,0 +1,92 @@
...
+    </div>
+    <div id="fruit_list">
+      <amp-list
+        src="/static/samples/json/fruits.json"
+        layout="fixed-height"
+        width="auto"
+        height="900"
+        items="."
+        single-item
+        id="fruit_data_list"
```
sebastianbenz: do we need the id?

examples/source/interactivity-dynamic-content/Restrict_Duplicate_Selection.html
elisameyer: just tested this and it's not needed - looks like it was saved from an earlier iteration of this snippet - good catch!

examples/source/interactivity-dynamic-content/Restrict_Duplicate_Selection.html
elisameyer: I need these styles here so you can see the whole list on a mobile screen (or else div is cut off). I could put them as a class in the header instead of here.

examples/source/interactivity-dynamic-content/Restrict_Duplicate_Selection.html
elisameyer: id not needed here

examples/source/interactivity-dynamic-content/Restrict_Duplicate_Selection.html
elisameyer: Removing it

philkrie:
Reviewing this change, looks like there is a duplicate "Restrict Duplicate Selections" option in the Playground. Clicking on the bottom duplicate results in odd behaviour, such as most playground templates no longer appearing. Looking at the code submitted, I'm not sure what the bug could be or if this merge is even responsible.

Screen Shot 2020-01-27 at 8 49 22 PM

sebastianbenz:
@elisameyer you're right, dashes are not supported, I'd suggest using camel case `selectedFruits` as you would in JS.

@philkrie re duplicate "Restrict Duplicate Selections": that's a problem on our side. The samples automatically get transformed into a websites and email sample and both are listed here.

PR #3410, opened 2020-01-23
Add WompMobile to providers

mattwomple:


pages/shared/data/tools.yaml
```diff
@@ -466,3 +466,14 @@ providers:
       fit: contain
     formats:
       - email
+  - name: WompMobile
+    url: https://www.wompmobile.com
```
sebastianbenz: Could you please link to a specific product/tool. Otherwise it's a bit confusing if you're listed under tools.

pages/shared/data/tools.yaml
```diff
@@ -466,3 +466,14 @@ providers:
...
+    type: Development
+    description: WompMobile is an Accelerator Platform to develop, validate, launch and manage enterprise-grade AMP experiences.
```
sebastianbenz: Can you rephrase in a way to focus on the tooling aspect?

pages/shared/data/tools.yaml
mattwomple: Updated URL to point to relevant tool: https://www.wompmobile.com/amp-experience-platform/

pages/shared/data/tools.yaml
mattwomple: Updated description to specifically mention use of tools: "Built-in tools monitor for errors, ensure validation, purge the AMP cache, and create Signed-Exchange and Server-Side-Rendered AMPs."

PR #3422, opened 2020-01-27
fix potential xss via unescaped url string

sebastianbenz:


playground/backend/api.js
```diff
@@ -52,7 +52,7 @@ api.get('/fetch', async (request, response) => {
     response.send(doc);
   } catch (error) {
     console.error('Could not fetch URL', error);
-    response.send(`Could not fetch URL ${url}`).status(400).end();
+    response.status(400).send('Could not fetch URL').end();
```
matthiasrohmer: `end` is automatically called by `send` 🙂

PR #3448, opened 2020-01-29
Better error message on wg import

sebastianbenz:


patrickkettner:
LGTM, but does look like the commit does more than the message implies

PR #3450, opened 2020-01-30
Added dynamic menu example to amp.dev

philkrie:
Adding html file and json to demonstrate usage of amp-sidebar, amp-list, and amp-mustache to create dynamic, nested menus.

CSS might be reducible, but could be placed in one line to simplify visually instead.

@elisameyer @sebastianbenz


examples/source/interactivity-dynamic-content/Dynamic_Menu_With_Sub_Menus.html
```diff
@@ -0,0 +1,167 @@
...
+        items="." id="nav-list">
+        <template type="amp-mustache" id="amp-template-id">
+          {{#items}}
+          <li>
+            <a class="rightarrow"
+              role="button" 
+              on="tap:{{title}}_Slide.toggle;"
+              tabindex="0">{{title}}</a>
+          </li>
+          <amp-sidebar id="{{title}}_Slide" layout="nodisplay" class="main-nav" side="right">
```
sebastianbenz: Wouldn't  amp-nested-menu work. better here?`https://amp.dev/documentation/components/amp-sidebar/#building-nested-menus

examples/source/interactivity-dynamic-content/Dynamic_Menu_With_Sub_Menus.html
```diff
@@ -0,0 +1,167 @@
+<!---
+author: philkrie
+formats:
+  - websites
+tags:
+  - dynamic-content
```
sebastianbenz: add: `preview: default` so that the sample can be previewed standalone

examples/source/interactivity-dynamic-content/Dynamic_Menu_With_Sub_Menus.html
```diff
@@ -0,0 +1,167 @@
...
+      content: "<";
+      float: left;
+      padding-left: 10px;
+      font-weight: bolder;
+      font-size: 25px;
+    }
+
+  </style>
+</head>
+<!-- ## Implementation -->
```
sebastianbenz: You can directly annotate specific code blocks. In this case I would first annotate the button and explain that it will trigger the menu.

examples/source/interactivity-dynamic-content/Dynamic_Menu_With_Sub_Menus.html
```diff
@@ -0,0 +1,167 @@
...
+<body>
+  
+  <header>
+    <h3>
+      Dynamic Sidebar
+    </h3>
+    <p>
+      This side bar is dynamically constructed from a JSON response, which contains a top level and a sub level
+    </p>
+    <button aria-label="open sidebar" on="tap:sideNav.toggle" tabindex="0">
```
sebastianbenz: I don't think the aria-label is needed here as it's the same as the button text.

examples/source/interactivity-dynamic-content/Dynamic_Menu_With_Sub_Menus.html
```diff
@@ -0,0 +1,167 @@
...
+      Toggle Sidebar
+    </button>
+  </header>
+
+  <amp-sidebar id="sideNav" layout="nodisplay" side="left">
```
sebastianbenz: add the main explanation part here.

examples/source/interactivity-dynamic-content/Dynamic_Menu_With_Sub_Menus.html
```diff
@@ -0,0 +1,167 @@
...
+    <div class="mobile-menu-header">
+      <span>Menu</span >
+      <i class="closebtn" on="tap:sideNav.close" tabindex="0" role="on">×</i>
```
sebastianbenz: suggestion: `s/x/X`

philkrie: to clarify you mean replace x with X?

examples/source/interactivity-dynamic-content/Dynamic_Menu_With_Sub_Menus.html
philkrie: I'm not sure... I actually tried it out here: https://glitch.com/edit/#!/amp-snippets?path=dynamic_menu_sub_menu/index.html

When you click into a submenu, the transition occurs but nothing appears, despite the li elements all being populated with the correct info when you look at the DOM. The inspector even highlights the boundaries of these elements when you select them in the elements tab.

I can't figure out why they aren't appearing: do you mind taking a look? Otherwise, this would be the preferred method, but I'm not sure if this is an amp-list + amp-nested-menu interaction bug.

philkrie:
Hey Sebastian, bumping this. As mentioned in replies, I couldn't get this example to work w/ amp-nested-menu: it doesn't appear to support dynamic menus.

examples/source/interactivity-dynamic-content/Dynamic_Menu_With_Sub_Menus.html
sebastianbenz: Looks like a bug to me. Can you please file a report.

sebastianbenz:
Thanks (and sorry for the long pause)! Looks like a bug. The sample looks good otherwise. We can either revert to the previous version or wait until the bug is fixed (the latter would be my preference).

philkrie:
+1, agreed that this should be the preferred pattern. Opened bug [here](https://github.com/ampproject/amphtml/issues/26874)

examples/source/interactivity-dynamic-content/Dynamic_Menu_With_Sub_Menus.html
cathyxz: Drive by comment: `<amp-nested-menu>` works when the entire `<amp-nested-menu>` is inside the `<amp-list>` template. See the documentation demo here: https://github.com/ampproject/amphtml/blob/master/extensions/amp-nested-menu/amp-nested-menu.md#dynamic-content-rendering

=) 

I think we can probably be more explicit about that in the documentation.

philkrie:
closing per @cathyxz 's comment in https://github.com/ampproject/amphtml/issues/26874

Issue #3282, opened 2019-12-12
Improve support page design


sebastianbenz:
We should address @DerekNonGeneric's feedback: yet to be mentioned is that this page is missing a few items:

Breadcrumbs
Title
Sidebar
Container
The reason for this is because it uses the very bland meta.j2 template.

https://github.com/ampproject/amp.dev/issues/2339#issuecomment-535322115

DerekNonGeneric:
@sebastianbenz, although a few of those were addressed with my PR, what I really wanted was for  [Support](https://amp.dev/support/) to be as visually appealing as [How to contribute](https://amp.dev/documentation/guides-and-tutorials/contribute/). That might be one of your goals too.

There are a couple of reasons why it instead looks more similar to [Get started](https://amp.dev/documentation/).
1. Both pages represent section overviews and I thought maintaining consistency was right.
2. The Support section didn't seem to have enough content at the time to warrant a sidebar.

Admittedly, this page leaves much to be desired. When I was working on this, I was thinking that it would a good idea to completely revamp the section overview templates so that all pages using them would benefit from the style improvements while maintaining a consistent and distinctive look.

P.S. To clarify, the template it is currently using isn't `meta.j2`, rather a new one derived from [Get started](https://amp.dev/documentation/).

matthiasrohmer:
I wonder if we should really invest in this page - and if yes what's the best way. Impulsively I'd shift over the missing information over to the [documentation's Get started page](https://amp.dev/documentation/). Also because the first few links in the list are identical to what is on Get started: docs, samples, tools, demos.

But I also don't really have an idea what needs a user has when clicking on this link to be honest. Personally I'd only go there as a last resort after having already read through all the other linked pages. The only interesting link then would be the one to StackOverflow and probably one to the Slack Team which isn't there yet.

What do you guys think?

DerekNonGeneric:
/to @matthiasrohmer

> I wonder if we should really invest in this page

Yes, we should. Removing the support page would be an accessibility concern. Aside from that, keep in mind that it's a section overview page. From an information architecture perspective, the FAQs depend on the existence of that page as they technically exist in the support section.

>  \- and if yes what's the best way.

[...]

PR #3466, opened 2020-02-01
♻️ Refactor breadcrumbs navigation

sebil:
While I was working on improving the supports pages (#3282) I saw that the breadcrumb navigation for the FAQ pages was broken/wrong linked. 
I couldn't resist refactoring the whole thing, because over time it became a little mess.

Introducing new breadcrumb navigation that:
- dynamically creates breadcrumbs without having to add them manually
- prevent double wording of last breadcrumb and page title
- shows items that have no page but exist as a dropdown trigger (eg. Guides & Tutorials > Develop)
- adds an anchor to the parent page but only if it's not the current page

---
breadcrumbs-1
---
breadcrumbs-2
---
Also in this PR, I removed the `support-overview.css` in the sake of the general `overview.css` to keep the code DRY.


DerekNonGeneric:
Nice! Does this support `doc.titles.breadcrumb`? Sometimes the titles are really long and the breadcrumb for that page ends up looking strange. Having a way to override it becomes necessary. This feature is documented in the Grow documentation I've linked below.

https://grow.io/reference/documents/#titles

sebil:
Hi Derek, thanks for your feedback.

No, I removed that feature, because the `doc.title` should never be used for the breadcrumbs as
1. this would cause double wording again
2. the breadcrumb item of the current doc didn't shows anyway

It now uses the `doc.collection.title` instead, which in most cases is the same as the navigation label.

Just in case of need, you could change that title in every `_blueprint.yaml` of each collection.

PR #3471, opened 2020-02-03
Removed past events from upcoming events

amanintech:
Removed wrongly mentioned events in upcoming events and moved them to past events in descending order


amanintech:
Welcome !

PR #3478, opened 2020-02-05
🔧 Fix filter order and quotes

DerekNonGeneric:
My mistake, the previous PR's quotes ended up getting repeated and the filter order was causing undesirable results. I've remedied this by changing the order of the filtering — this should be correct.

### DIFF
--- a/frontend/templates/views/partials/structured-data.j2
+++ b/frontend/templates/views/partials/structured-data.j2
@@ -103,7 +103,7 @@
 {% if description %}
 <meta name="twitter:description" content="{{ description }}">
 {% endif %}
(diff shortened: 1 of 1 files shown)


DerekNonGeneric:
Since we're on the topic, I wonder if the same treatment should be applied to `og:title`. I think this is the one Facebook uses. Not sure if it also has trouble with this.

```
<meta property="og:title" content="{{ title_prefix }}{{ title }}">
```

matthiasrohmer:
Bildschirmfoto 2020-02-05 um 14 59 54

Facebook seems to handle it just fine 🙂

Issue #3395, opened 2020-01-17
Wordpress Blog Posts from homepage are using originally uploaded images


kristoferbaxter:
### Current behavior
Blog posts on the homepage of `amp.dev` are using originally uploaded image content.

![Screenshot from 2020-01-17 13 48 10](https://user-images.githubusercontent.com/61764/72648531-1d3d6e00-3930-11ea-8671-e4540f2a570f.png)

This example image is 838KB and 685/385 pixels. For this surface, the size can be significantly smaller and as a result compress much better. 

Please also provide:
- Page link: https://amp.dev

### Expected behavior/code
Wordpress automatically creates smaller preview images for content uploaded that can be used.

https://blog.amp.dev/wp-json/wp/v2/posts?per_page=3&_embed

From the json feed on posts, look for `_links['wp:featuredmedia']` for the entity. : https://blog.amp.dev/wp-json/wp/v2/media/3064

This JSON feed will return the available media to use including a viable candidate `media_details.sizes.medium`. https://blog.amp.dev/wp-content/uploads/2019/12/Screen-Shot-2020-01-17-at-21.23.56-300x169.png

Using this image cuts the size of this resource from 838KB -> 76KB.

- What browser are you using? Chrome 79
- What O/S are you using? Linux

matthiasrohmer:
Fixed by #3486.

PR #3486, opened 2020-02-06
⚡️ SSR blog posts on home and use REST API over RSS feed

### LINKED ISSUE #2735: SSR latest blog posts on homepage
We should check more regularly for new blog posts than with every deploy. 

see #1883

robinvanopstal:
Use custom nunjucks filter to load a smaller version of blog post teaser images on homepage

Resolves #2735, #3395


platform/lib/templates/ImportBlogFilter.js
```diff
@@ -0,0 +1,57 @@
...
+const POST_COUNT = 6;
+const BLOG_PATH = `https://blog.amp.dev/wp-json/wp/v2/posts?per_page=${POST_COUNT}&_embed`;
+const DEFAULT_IMG = 'AMP_Blog_Square.jpg';
+const fetch = require('node-fetch');
+const moment = require('moment');
+
+async function importBlog(value, callback) {
+  let response = [];
+  try {
+    response = await fetch(BLOG_PATH).then(res => res.json());
```
sebastianbenz: don't mix `then` and `await`

sebastianbenz: you also need to check `response.ok` for the status code

platform/lib/routers/growPages.js
```diff
@@ -275,7 +275,14 @@ growPages.get(/^(.*\/)?([^\/\.]+|.+\.html|.*\/|$)$/, async (req, res, next) => {
 
   let renderedTemplate = null;
   try {
-    renderedTemplate = template.render(templateContext);
+    renderedTemplate = await new Promise((resolve, reject) => {
```
sebastianbenz: why do you need to wrap render into a promise?

robinvanopstal: Needed since we switched to asynchronous template rendering.

sebastianbenz: got it, you can also use Node's promisify helper for this as well. Why are we switching to async here?

matthiasrohmer: Async filters like the newly introduced `importBlog` only work when nunjucks is used in async mode which is implictly invoked by using `Template.render` with a callback.

sebastianbenz: thanks for the clarification

matthiasrohmer: Still, @robinvanopstal please refactor to `promisify` as Sebastian suggested 😉

platform/lib/templates/ImportBlogFilter.js
```diff
@@ -0,0 +1,62 @@
...
+ * limitations under the License.
+ */
+
+'use strict';
+
+const POST_COUNT = 6;
+const BLOG_PATH = `https://blog.amp.dev/wp-json/wp/v2/posts?per_page=${POST_COUNT}&_embed`;
+const DEFAULT_IMG = 'AMP_Blog_Square.jpg';
+const fetch = require('node-fetch');
+const moment = require('moment');
```
matthiasrohmer: As we're not using `moment` anywhere else and Wordpress delivers a well-formed date string it would be nice to get rid of this dependency.

platform/lib/templates/ImportBlogFilter.js
```diff
@@ -0,0 +1,62 @@
...
+    let imageUrl = '';
+    if (mediaDetails && !mediaDetails.file.endsWith(DEFAULT_IMG)) {
+      imageUrl = mediaDetails.sizes.medium.source_url;
+    }
+
+    posts.push({
+      title: post._embedded['wp:term'][0][0].name,
+      image: imageUrl,
+      headline: post.title.rendered,
+      date: moment(post.date).format('MMMM D, YYYY'),
```
matthiasrohmer:
```suggestion
      date: new Date(post.date).toLocaleString('en-us', {month: 'long', year: 'numeric', day: 'numeric'});
```

Not sure if this works for Node.js but in Chrome it prints:

```js
const date = new Date('2020-02-05T16:26:53');
date.toLocaleString('en-us', {month: 'long', year: 'numeric', day: 'numeric'});
"February 5, 2020"
```

PR #3493, opened 2020-02-06
Use fixed-height layout for calendar with desktop/mobile sizes

rcebulko:
The original implementation used `layout="intrinsic" height="0"` for the `amp-script` container element. A series of recent bug fixes fixed how `amp-script` works with `layout="intrinsic"`, and this broke the page by setting the container's height to 0.

This PR doubles up the `amp-script` tag and defines each as a `fixed-height` layout based on the grid sizes for desktop and mobile views. A CSS media query controls which one is displayed. Tested on desktop and emulated mobile; confirmed valid AMP with AMP Validator Chrome Extension.

### DIFF
--- a/frontend/scss/components/organisms/release-schedule.scss
+++ b/frontend/scss/components/organisms/release-schedule.scss
@@ -7,7 +7,12 @@ body.amp-mode-touch .touch-only {
 body.amp-mode-touch .no-touch-only {
   display: none;
 }
-#calendar {
+// The AMP optimizer doesn't support the media attribute, so we hide whichever
+// calendar is not is not displayed with a media query.
+#calendar-container-mobile {
+  display: none;
+}
+.calendar {
   display: grid;
   gap: 10px;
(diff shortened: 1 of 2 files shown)


frontend/scss/components/organisms/release-schedule.scss
```diff
@@ -7,7 +7,12 @@ body.amp-mode-touch .touch-only {
 body.amp-mode-touch .no-touch-only {
   display: none;
 }
-#calendar {
+// The AMP optimizer doesn't support the media attribute, so we hide whichever
```
danielrozenberg: That's unfortunate :(

rcebulko: > Not a huge fan of doubling the calendar, but I guess it's a legit workaround for now

My understanding is that's just how it's supposed to be done. I looked to see if we could have the `height` attribute be media-dynamic, but no dice.

Issue #3152, opened 2019-10-28
Better navigation tools for the blog


tobie:
## Feature Request

# Is your feature request related to a problem? Please describe.

It's super hard to find past content on the blog.

# Describe the solution you'd like

* Search tool
* category index in the sidebar
* category index pages with only the blog title and date plus eventually a short extract, **not** the whole blog post!
* archive index pages by month and year (again with just the title and categories, not the whole article as is currently the case).

# Describe alternatives you've considered

`site:blog.amp.dev` on search.

matthiasrohmer:
Hi @westonruter & @amedina could you and/or your team have a look at this? 🙂Let us know if we can help in any way.

westonruter:
For context here see https://github.com/ampproject/amp.dev/pull/2934#issuecomment-531443543. This is why the search markup is currently disabled on the blog.

It is blocked by https://github.com/ampproject/amphtml/issues/24661 which would be fixed by this PR: https://github.com/ampproject/amphtml/pull/24793

westonruter:
I'm finally getting to this.

@mathiasbynens I found a blocker, however. The search endpoint is returning relative URL results. For example https://amp.dev/search/highlights is returning URLs like `/documentation/guides-and-tutorials/start/converting/`. The same goes for the https://amp.dev/search/do

Can these endpoints be updated to return the URL with the host name? Otherwise they result in 404s. The alternative would be to hack the Mustache template to inject `https://amp.dev` before the template variables.

westonruter:
Another complication: The [shared header AMP fragment](https://amp.dev/shared/header/) includes the use of transformed AMP attributes and class names. These are causing validation errors in the AMP plugin which are causing its sanitizer remove them. This is problematic for class names since currently the plugin removes the _entire_ class name when it contains `i-amphtml-*`. Can this fragment be made non-optimized?

westonruter:
I can work around this by doing:

```php
$header = preg_replace(
	'/\bi-amphtml-([a-z-]+)(=".*?")?/s',
	'',
	$header
);
```

I can't, however, safely work around the issue with the relative URLs being returned in the search result endpoints.

sebastianbenz:
@matthiasrohmer can you please take a look

matthiasrohmer:
Will do. Can't get my head around how whenever we touched the optimizing process the shared fragments ended up optimized 😅

matthiasrohmer:
[...]

PR #3500, opened 2020-02-08
🐛 Use absolute URLs for search API endpoints

matthiasrohmer:
To add search to blog.amp.dev beside the endpoints returning absolute URLs the endpoints themselves need to be called by their absolute URLs. See https://github.com/ampproject/amp.dev/issues/3152#issuecomment-568150532 for context.

### DIFF
--- a/frontend/templates/views/partials/search.j2
+++ b/frontend/templates/views/partials/search.j2
@@ -11,7 +11,9 @@
 {% do doc.amp_dependencies.add('amp-lightbox', '0.1') %}
 {% do doc.amp_dependencies.add('amp-mustache', '0.2', 'template') %}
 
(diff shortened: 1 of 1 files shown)


matthiasrohmer:
Thanks for checking, @tharders! I was indeed assuming it to *just work* but it seems as if the regular `localhost` exceptions (i.e. simple HTTP allowed) don't apply when `src` is bound.

@sebastianbenz is that worth reporting?

@tharders please have another look. The dev environment should now use relative URLs again.

sebastianbenz:
@matthiasrohmer sounds like a bug to me, please report

matthiasrohmer:
Thanks! Filed ampproject/amphtml#26692.

PR #3502, opened 2020-02-09
zh-CN translation update

andyc15:
That's a lot of translation work need to be done.
Done about 1/2
Also I found out some mispelling like "affiliate link → affliate link"


andyc15:
@googlebot I signed it!

sebastianbenz:
@patrickkettner @CrystalOnScript PTAL

sebastianbenz:
Thanks a lot for the contribution and sorry for the slow response!

PR #3550, opened 2020-02-14
🐛 Correctly link to amp-script doc from example

matthiasrohmer:
Corrupt references in the examples are not picked up by `GrowReferenceChecker` as the pages are only built from the JSON files. We could think about checking the sample source files though. 🤔


examples/source/1.components/amp-script/index.html
```diff
@@ -1,5 +1,5 @@
 <!--
-  An [amp-script](content/amp.dev/documentation/components/reference/amp-script.md) hello world example.
+  An [amp-script](content/amp-dev/documentation/components/reference/amp-script.md) hello world example.
```
sebastianbenz: shouldn't this have a leading slash as well?

matthiasrohmer: Indeed! Fixed.

PR #3551, opened 2020-02-15
Update amp-video-import.html

iaroel:
Typo in script's src.


iaroel:
> @googlebot I signed it!

PR #3552, opened 2020-02-16
🚧 Experiment with packaged and shipped Grow

matthiasrohmer:
This needs a lot more testing, don't merge. Part of #3330.

Instead of using a pip-installed version of Grow this branch uses a packaged version and would additionally update the CI from Grow 0.7.4 to the latest 0.8.19. Furthermore it disables the confusing "Server started..." message.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,17 +13,12 @@ cache:
 install:
 - npm ci
 before_install:
-- |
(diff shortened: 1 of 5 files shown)


matthiasrohmer:
Issued https://github.com/grow/grow/issues/1057.

matthiasrohmer:
Will try this again as soon as Grow has released Python 3 support.

PR #3583, opened 2020-02-21
add upcoming roadshow dates and locations

patrickkettner:


patrickkettner:
image

morsssss:
Looks good - thanks for doing this so amazingly quickly!

My only question is - should Colombo be so far in the water? Would you want to stick it closer to India? I know we don't want to obscure Bengaluru, but what if it was simply a few pixels further north?

If that looks worse, please leave as is :)

patrickkettner:
hows this, @morsssss?

image

patrickkettner:
ping @morsssss

morsssss:
Cool! You may or may not want to move it over one pixel to the right so that we can see Bengaluru better. But LGTM either way.

patrickkettner:
done! would you be able to mark it as approved if it looks ok?

PR #3589, opened 2020-02-23
add custom form validation amp-script sample

sebastianbenz:
Context: https://stackoverflow.com/questions/60317056/how-to-implement-regex-in-amp


morsssss:
This LGTM. My only question is whether we want to encourage people to use `querySelector()` instead of `getElementById()`... because the former is only supported for very simple cases like this one.

PR #3598, opened 2020-02-24
Updated CSS limit to 75,000 bytes from 50,000

Cauchon:
https://github.com/ampproject/amphtml/pull/26475

https://amp.dev/documentation/guides-and-tutorials/develop/style_and_layout/


Cauchon:
@googlebot I signed it!

PR #3605, opened 2020-02-25
add AMP Conf banner to localization files

patrickkettner:


patrickkettner:
@sebastianbenz yes, for sure. we also need linting. It is a bit complex due to how many different ways yaml or other files can be injected. but at the very least ensuring the simple cases are added would be a good start

Issue #3592, opened 2020-02-24
Format user notification does not stick to header


sebastianbenz:
<img width="1280" alt="Screen Shot 2020-02-24 at 10 38 40" src="https://user-images.githubusercontent.com/380472/75142107-a34c9180-56f2-11ea-9239-25dd4dba1b82.png">

PR #3607, opened 2020-02-26
take a whack at fixing the new header filters

patrickkettner:
every part of this feels super gross, but it seems like a somewhat robust way to approach #3592.

I set `!important` on `position: sticky` , because [they did it first] :/(https://github.com/ampproject/amphtml/blob/5f75efe35b734a4fcf0884d373315de60cd42df9/extensions/amp-user-notification/0.1/amp-user-notification.css#L19).
I used a magic numbers for the height because the header doesn't have a sass variable available (that I can determine)

### DIFF
--- a/frontend/scss/components/molecules/format-filter-hint.scss
+++ b/frontend/scss/components/molecules/format-filter-hint.scss
@@ -13,10 +13,14 @@ $height: 60px;
   height: $height;
   color: color('white');
   background-color: color('ebony-clay');
+  position: sticky!important;
+  top: 50px;
+
   @include txt-2;
 
   @media (min-width: 768px) {
     display: block;
+    top: 85px;
   }
 
   &.amp-active + &-spacer {


frontend/scss/components/molecules/format-filter-hint.scss
```diff
@@ -13,10 +13,14 @@ $height: 60px;
   height: $height;
   color: color('white');
   background-color: color('ebony-clay');
+  position: sticky!important;
```
sebastianbenz: this is unfortunately not valid AMP

patrickkettner: Using important?

sebastianbenz: https://amp.dev/documentation/guides-and-tutorials/develop/style_and_layout/style_pages/

PR #3611, opened 2020-02-27
🐛 Prevent format filter banner from jumping

robinvanopstal:
Fixes: #3592

### DIFF
--- a/frontend/templates/views/partials/format-filter-hint.j2
+++ b/frontend/templates/views/partials/format-filter-hint.j2
@@ -5,6 +5,11 @@
 <amp-user-notification class="ap-m-format-filter-hint" id="format-filter-hint" layout="nodisplay">
 </amp-user-notification>
 
+{#
+  Amp user notification is used to handle the banners state.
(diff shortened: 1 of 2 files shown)


matthiasrohmer:
@sebastianbenz, this finally provides fairly decent UX for the banner but its actually a little hacky as the actual visible notification is not inside `amp-user-notification` anymore to be able to style it independently of `amp-user-notification`s `position: fixed`.

frontend/templates/views/partials/format-filter-hint.j2
```diff
@@ -3,14 +3,16 @@
 
 {% do doc.amp_dependencies.add('amp-user-notification', '0.1') %}
 <amp-user-notification class="ap-m-format-filter-hint" id="format-filter-hint" layout="nodisplay">
-  <div class="ap-m-format-filter-hint-content">
-    <div class="ap-m-format-filter-hint-note">{{ _('Select your format for a more streamlined experience') }}</div>
+</amp-user-notification>
```
sebastianbenz: Please add a comment that the next element needs to directly follow the user notification

Issue #2774, opened 2019-08-16
Table of contents in courses now appears twice


morsssss:
See, for example, [https://amp.dev/documentation/courses/beginning-course/our-first-amp-page/](https://amp.dev/documentation/courses/beginning-course/our-first-amp-page/).

The entries in the table of contents appear twice.

/cc @dlemm @matthiasrohmer 

Otherwise the new toggle looks really nice! :)

matthiasrohmer:
😕 Thanks for the report, Ben! Sadly I can't think of a good technical solution for this. One thing that would fix it partly though is to only duplicate sections that actually differ in content. This would also make maintenance a lot easier. Currently even course documents that are equal for beginners and experts have the same content wrapped by `[filter]` tags.

But even with this section headlines that are hidden by `[filter]` still show up as Grow has no clue of those tags and simply gathers all headlines from the document... I'll need to think about it a little more.

morsssss:
Welllll.... I remember that when we first looked at these courses on the site, I asked if we could simply hide the table of contents. It's somewhat useful, but no single webpage is that long... and the real downside is that the remaining width for code samples is small, so you have to scroll around to read the code samples.

In other worse, I think the usefulness of the table of the contents is outweighed by the fact that it makes the code samples hard to read.

Can we just suppress the TOC for now? Or permanently? 😃

tharders:
@dlemm  will disable the table of contents for the courses to have a quick fix this problem.
But we must prevent the underlying problem, since also the format filters are affected.

For example the paragraph "Disallowed property inside attribute in tag" on following page:
ttps://amp.dev/documentation/guides-and-tutorials/learn/validation-workflow/validation_errors/?format=websites#disallowed-property-inside-attribute-in-tag
is not available for email:
https://amp.dev/documentation/guides-and-tutorials/learn/validation-workflow/validation_errors/?format=email#disallowed-property-inside-attribute-in-tag

It feels like the most clean solution to fix the table of contents with filtering would be to add filter commands directly when the table of contents html code is generated.
We already patched the markdown TocTreeprocessor, so it would be possible to take this a step further.

PR #3622, opened 2020-03-02
🚸 Filter TOC based on format and level

### LINKED ISSUE #3545: Sections hidden for a certain AMP format aren't hidden in the table of content
### Current behavior

One example:

https://amp.dev/documentation/components/amp-form/?format=email#redirecting-after-a-submission

The "Redirecting after a submission" section is hidden when the selected format is "Email", but it's still displayed in the table of content on the right. Clicking on it from the table of content does nothing because the section that corresponds to the anchor is missing on the page.

### Expected behavior/code

Hidden section not displayed in the table of content.

matthiasrohmer:
That's ugly and did cost me more time than I had hoped. It's not possible to access `doc.format.toc` in any of the render hooks to get the generated TOC as this would reevaluate the document's content and make all transformations obsolete.

Therefore this passes a callable to the template in which you are free to call `doc.format.toc` as `doc.format.content` has already been rendered.

Resolves #3545, adds to #2774. @morsssss, this would allow the courses to have level-specific TOCs again. Do you want them back?

### DIFF
--- a/pages/extensions/amp_dev/markdown_extras/block_filter.py
+++ b/pages/extensions/amp_dev/markdown_extras/block_filter.py
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
 from __future__ import unicode_literals
 
+import json
 import re
 
 FILTER_TRIGGER = '[filter'
@@ -15,6 +16,7 @@ def trigger(original_body, content):
   if FILTER_TRIGGER in original_body:
     return _transform(content)
   return content
+
 
 def _transform(content):
     for match in FILTER_TAG_PATTERN.findall(content):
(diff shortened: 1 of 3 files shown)


pages/extensions/amp_dev/markdown_extras/block_filter.py
```diff
@@ -41,3 +43,35 @@ def _get_attributes(match):
     for match in matches:
         attributes[match[0]] = match[1]
     return attributes
+
+FILTERED_SECTION_PATTERN = re.compile(r'({% call filter\(.*?\) %})(.*?)({% endcall %})', re.MULTILINE | re.DOTALL)
```
sebastianbenz: Is this the same pattern used by the normal filter? Does it make sense to extract this?

matthiasrohmer: Not quite. That's the step in between: we have `[filter *]` in our Markdown which get's rewritten to jinja2 calls like the one in the pattern (`{% call filter * %}`) which then get evaluated to SSR statements. The right thing to do (which would als simplify all of this here) is to skip jinja2. Created https://github.com/ampproject/amp.dev/issues/3657 to track this.

I held back from doing this directly as I haven't thought about any implications this might have.

sebastianbenz: great - thanks!

patrickkettner:
[this](https://github.com/ampproject/amp.dev/pull/3622/files#diff-4349b7c171ced5334504141576da2cc9R23) seems to be breaking my local build when accessing components reference pages

image

patrickkettner:
cc @matthiasrohmer @sebastianbenz

sebastianbenz:
A zombie Python process did cause this for me. Try `kill Python` and run
gulp develop again/

>

patrickkettner:
That works for me, thanks
-- 
patrick

PR #3655, opened 2020-03-06
✨Add show more button to sample list

robinvanopstal:


frontend/templates/views/partials/component-intro.j2
```diff
@@ -43,13 +43,20 @@
...
+      <input type="checkbox" id="load-examples-checkbox"/>
       <div class="ap-o-component-intro-box-examples">
         {% for example in examples[doc.title] %}
         [% if format in {{ example.formats|tojson }} %]
         <a href="{{ example.url }}">{{ example.title }}</a>
         [% endif %]
         {% endfor %}
       </div>
+      {% if examples[doc.title]|length > 6 %}
+      <label class="ap-o-component-intro-load-btn" for="load-examples-checkbox">
```
sebastianbenz: Suggestion: use the `toggleClass` action instead as it makes this easier and the markup clearer.

frontend/scss/components/organisms/component-intro.scss
```diff
@@ -125,10 +125,20 @@
           padding-right: 12px;
         }
 
-        &:nth-child(1n + 7) {
+        &:nth-child(1n + 9) {
```
sebastianbenz: Can we reduce the number to 4 on mobile?

robinvanopstal: Sure! :)

PR #3664, opened 2020-03-09
🐛 Escape headline when filtering TOC

matthiasrohmer:
This fixes two problems: Markdown (or more correct: the resulting HTML) is stripped from the TOC but I've been searching for the headlines with them still containing Markdown syntax. I am assuming we only use code blocks in headlines and therefore simply replace backticks.

Second problem was that replacing failed for headlines that contained regular expression special chars. This is now correctly handled by using `re.escape`.


matthiasrohmer:
Tested with the `amp-bind` doc where *`src` (optional)* hasn't been filtered:

Issue #3670, opened 2020-03-10
Do not show examples headline in intro table if there are none


matthiasrohmer:
See `amp-bind` for e-mail for example.

PR #3676, opened 2020-03-11
🐛Hide examples headline in intro table if there are none

robinvanopstal:
Fixes: #3670


frontend/templates/views/partials/component-intro.j2
```diff
@@ -39,7 +39,7 @@
 
   {% set examples = g.json('/shared/data/componentSamples.json') %}
   {# Check if there are examples for the current component #}
-  {% if examples[doc.title] %}
```
matthiasrohmer: Keep those in addition to the new SSR statement.

Issue #3672, opened 2020-03-10
Documentation: toc not scrollable


sebastianbenz:
For example: https://amp.dev/documentation/components/amp-bind/?format=websites

I need to scroll the whole page in order to scroll the TOC on the right.

//cc @matthiasrohmer

PR #3695, opened 2020-03-16
🐛 Fix documentation toc scrolling

robinvanopstal:
Fixes: #3672


frontend/scss/components/organisms/toc.scss
```diff
@@ -52,6 +52,7 @@
     height: auto;
     top: 90px;
     right: auto;
+    padding-bottom: 90px;
```
sebastianbenz: nit: using a CSS var would've been nice hear to make it clear that both values are the same.

PR #3696, opened 2020-03-16
Add a `go.amp.dev` link for working group updates

rsimha:
This PR adds a short link go.amp.dev/wg-updates to redirect to AMP's [working group updates](https://github.com/search?o=desc&q=org%3Aampproject+label%3A%22Type%3A+Status+Update%22&s=created&type=Issues).


rsimha:
@sebastianbenz Do you think this is a misuse of the amp.dev short link system? If so, do you have any alternatives that will work for open source?

PR #3711, opened 2020-03-17
amp-sidebar example updated to include always visible desktop sidebar

krdwan:
Issue: https://github.com/ampproject/amphtml/issues/26196

Added a navigation sidebar to the amp-sidebar example that is always visible.


krdwan:
Reopened under a different branch.

PR #3712, opened 2020-03-17
📖 Update amp-sidebar example to include always visible desktop sidebar

krdwan:
Fixes ampproject/amphtml#26196

Added a navigation sidebar to the amp-sidebar example that is always visible.

@sebastianbenz

### DIFF
--- a/examples/source/1.components/amp-sidebar.html
+++ b/examples/source/1.components/amp-sidebar.html
@@ -31,7 +31,9 @@ author: juliantoledo
   <style amp-boilerplate>body{-webkit-animation:-amp-start 8s steps(1,end) 0s 1 normal both;-moz-animation:-amp-start 8s steps(1,end) 0s 1 normal both;-ms-animation:-amp-start 8s steps(1,end) 0s 1 normal both;animation:-amp-start 8s steps(1,end) 0s 1 normal both}@-webkit-keyframes -amp-start{from{visibility:hidden}to{visibility:visible}}@-moz-keyframes -amp-start{from{visibility:hidden}to{visibility:visible}}@-ms-keyframes -amp-start{from{visibility:hidden}to{visibility:visible}}@-o-keyframes -amp-start{from{visibility:hidden}to{visibility:visible}}@keyframes -amp-start{from{visibility:hidden}to{visibility:visible}}</style><noscript><style amp-boilerplate>body{-webkit-animation:none;-moz-animation:none;-ms-animation:none;animation:none}</style></noscript>
   <style amp-custom>
     :root {
+      --space-1: 0.5rem;
       --space-2: 1rem;   /* 16px */
+      --space-4: 2rem;
     }
     .sample-sidebar {
       width: 250px;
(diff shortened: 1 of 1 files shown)


examples/source/1.components/amp-sidebar.html
```diff
@@ -42,102 +44,178 @@
...
+          <div id="target-element-right">
+          </div>
+        </div>
+    </article>
+  </main>
+  <!-- ## Desktop Sidebar -->
+  <!--
+    The amp-sidebar can be used as a permanent navigation bar on the main page.
+    The toolbar target is defined as a left hand navigation element using css: flex.
+    (See the `aside` element under `main`.)
```
caroqliu: Instead of referring the reader elsewhere on the page, let's provide a code snippet of roughly what's going on. The important information is that the sidebar behavior is available as usual below a given width threshold and permanently affixed above it. This is a skeleton of what you could describe in the comments here:
```html
<main>
   <aside> ... target for the sidebar above the min-width threshold </aside>
</main>
<amp-sidebar ... > ... </amp-sidebar>
```

examples/source/1.components/amp-sidebar.html
```diff
@@ -42,102 +44,178 @@
...
+        flex-direction: row;
+      }
+      aside {
+        width: 300px;
+      }
+    }
   </style>
 </head>
 <body>
+  <!-- Header -->
```
sebastianbenz: Can you move these below into the Desktop Toolbar section? Otherwise they will be displayed at the top of the example page where it's a bit confusing:

Screen Shot 2020-03-19 at 10 35 38

examples/source/1.components/amp-sidebar.html
```diff
@@ -42,102 +44,178 @@
...
+    <h1>Header</h1>
+    <button
+      class="hamburger"
+      on='tap:sidebar-desktop.toggle'
+      aria-label="Click to open sidebar">
+      =
+    </button>
+  </header>
+  
+  <!-- Main Content -->
```
sebastianbenz: Nesting comments is not supported. Suggestion: remove

examples/source/1.components/amp-sidebar.html
sebastianbenz: See my suggestions above. Integrate header and main here (ideally via three consecutive code snippets).

examples/source/1.components/amp-sidebar.html
krdwan: Thanks!  I have removed the comments so these do not need to be annotated in the documentation.

examples/source/1.components/amp-sidebar.html
krdwan: I have removed, thanks!

examples/source/1.components/amp-sidebar.html
krdwan: Thanks for the feedback Caroline, Sebastian.  I have updated this section with three code snippets, bringing in the aside and CSS from previous sections.  This will make the flow of the page more logical and keeps the user from having to scroll around the page.  Thanks!

krdwan:
Thanks for the review @caroqliu @sebastianbenz !

Issue #3658, opened 2020-03-09
Improve filter block matching


matthiasrohmer:
Every once in a while a corrupt `[filter]` statement in one of the imported docs breaks the build. Cases that I remember when this happened:

- `[filter formats="websites]`: missing quote
- `[filter formats='websites"]`: non-matching quotes
- Unclosed `[filter]` block

Those should all be handled gracefully instead of breaking the build. Could be fixed with #3657.

patrickkettner:
closing as it is [outdated](https://github.com/ampproject/amp.dev/pull/3733#issuecomment-757328420
), and taken care of by grow v1

Issue #3657, opened 2020-03-09
Skip jinja2 macros to build filter SSR expressions


matthiasrohmer:
Our custom Markdown tag `[filter]` currently relies on one unnecessary transpiling step. The flow is as follows:

We have `[filter *]` in our Markdown which get's rewritten to jinja2 calls like `{% call filter * %}` which then get evaluated to SSR statements like [% if format.includes(... %]`.

Instead of relying on the template engine to build the SSR statements [our Markdown extension](https://github.com/ampproject/amp.dev/blob/future/pages/extensions/amp_dev/markdown_extras/block_filter.py) should directly emit them.

sebastianbenz:
Maybe I'm seeing things to simplistic, but couldn't we define custom [Nunjucks tags](https://mozilla.github.io/nunjucks/api.html#custom-tags) that match our filter syntax?

```
[% format "emails" %]
may the force be with you
[% endformat %].
```

This might require rewriting the existing docs, but that'd be fine imo.

matthiasrohmer:
Yes, that would work for the general functionality. Two things to consider:

1) We'd still need a way to parse the filtering in the Grow-layer to format-filter the TOC. Alternatively, we could also generate the TOC in our SSR-layer.
2) I only thought about solutions that work for all of our tags (`[tip]`, `[video]`, `[filter]`) - for the tags that involve templating the SSR approach feels a little wrong as it would really mix things up.

sebastianbenz:
1. TOC in SSR is probably going to be ugly. Probably not uglier than in Grow though.
2. Can you elaborate?

matthiasrohmer:
Sure! For `[filter]` the proposed approach works fine, since it doesn't create any new HTML but only works with the already existing structure.

I'll try to illustrate what I don't like about SSRing for example the `[tip]` shortcode. That's the flow currently:

*Data*: `/content/doc.md`

```md
---
$title: Document
---

# Headline
Lorem ipsum.

[tip type="note"]
That's a note.
[/tip]
```

This data file gets rendered by Grow in two phases. The first one is the markdown render step in which we also transform our custom markdown blocks to `{% call %}` blocks:

```jinja2
<h1>Headline</h1>
<p>Lorem ipsum.</p>

{% call tip(type='note') %}
That's a note.
{% endcall %}
```

The second one is jinja2 rendering all remaining jinja2 statements: `{% call %}`, '{% include %}`, ...

 ```html
<body>
[...]

PR #3733, opened 2020-03-19
🚧 Rework custom markdown tag handling

matthiasrohmer:
Works towards #3657 and #3658.

The basic problem of our custom markdown extensions (besides them consisting of multiple rendering layers) are that they are RegEx based at their core. This introduces pitfalls like the following:

```
[filter format="'websites"]
This is for websites.

[filter level="advanced"]
And this only for website experts.
[/filter]

[/filter]
```

The example above currently only works fine because it is translated to `{% call %}` tags which are evaluated by jinja2 which uses a parse-tree and gracefully handles nested tags. If we skip jinja2 and only rely on regular expressions this would fail as the inner closing tag would match the first opening tag. The only way I see to approach this is to use a parser for our custom tags while staying away from building our own - but I'm coming to that.

We initially switched to the RegEx based solution coming from [bbcode](https://github.com/dcwatson/bbcode). The problem was it is kinda slow, malformatted some of our documents and didn't have a great API. Therefore I reevaluated some others and came to [PySkCode](https://github.com/TamiaLab/PySkCode). It would serve all our needs and is simply extensible. Our `[tip]` tag, for example, would look like this:

https://github.com/ampproject/amp.dev/blob/925c266487380c97fa610b2fc28015b7d4c24309/pages/extensions/amp_dev/markdown_extras/block_tip.py#L12-L14

Although it's simple to extend it still comes with downsides: we mix HTML templating and logic. Before our tip template was simple jinja2:

https://github.com/ampproject/amp.dev/blob/80e0f4644fc2dc24fa2e73b82c166ec30a4b7bb1/frontend/templates/views/partials/tip.j2#L6-L15

But there are more downsides: PySkCode was actually built for Python 3. I could easily backport it for 2.7. (since Grow is still working on Python 3 support: https://github.com/grow/grow/issues/164) by using [3to2](https://pypi.org/project/3to2/) and doing some minor manual adjustments though now it's not really upgradeable anymore. Also, we are manually adding code to our repository which should actually be installed by a package manager because we are trying to get rid of the need for `pip` at the same time.

I'm somehow stuck deciding how to proceed that's why I'm opening this draft PR for discussion: I think writing a parser on our own isn't worth the effort which would, in my opinion, better be invested in switching to something that would us get rid of all the Python stuff all along. Using a third-party, existing parser comes with the downsides described above...

### DIFF
--- a/pages/content/amp-dev/documentation/guides-and-tutorials/start/test.md
+++ b/pages/content/amp-dev/documentation/guides-and-tutorials/start/test.md
@@ -1,0 +1,25 @@
+---
+$title: Test
+$order: 1
+description: 'Follow these best practices to create a satisfying snackable story'
+formats:
+  - websites
+  - stories
+author: matthiasrohmer
+---
+
+Stories on the web are an immersive, tappable, and easily shareable storytelling format. Stories is an opportunity for creators and publishers to structure content in a full-screen, visually rich, and engaging mobile first experience for users.  
+
+Stories are snackable, and readers like quick bites in their micro moments! Waiting for the subway or grabbing coffee opens up opportunities for condensed content consumption! Ensure each bite engages and delights by following these best practices for creating a satisfying snackable story.
+
+
+## At a glance
+
+[tip type=note]
+Write an awesome story!
+[/tip]
+
+*   Stories must be valid AMP. You can test yours in the [AMP validator](https://validator.ampproject.org/).
+*   [Metadata attributes](/content/amp-dev/documentation/components/reference/amp-story.md/#metadata-guidelines) ensure your story is defined well. Check that your story has them setup correctly.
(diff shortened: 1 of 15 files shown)


matthiasrohmer:
Ugh... the "Create Draft PR" button is easy to miss. Don't merge this.

sebastianbenz:
> better be invested in switching to something that would us get rid of all the Python stuff all along

Do you mean getting rid of grow?

matthiasrohmer:
Yes. Of course this is not to underestimate the effort needed to switch from Grow to any JS-based static site generator. But we've already built so many things for and around Grow for it to fit our needs that it's not really fun anymore 😅

sebastianbenz:
Would be good to understand what it would take to switch to Eleventy.

>

matthiasrohmer:
Outdated and things will be different with Grow 1 presumably. Hence closing.

PR #3757, opened 2020-03-24
📈 Add various analytics triggers

### LINKED ISSUE #3723: Add dedicated tracking for how users are navigating to and from documentation detail pages
TBD in detail. Generally: are they coming from teasers on the overview pages? Are they using the sidebar? Is the sidebar discoverable enough...?

matthiasrohmer:
Resolves #3723. This adds analytics triggers for the following events:

- Clicks on links inside the table of contents 
- Clicks on links inside the sidebar
- Toggling of the sidebar
- Toggling of the TOC
- Clicks on breadcrumb links
- Clicks on a new format inside the format toggle or the format notification
- Scroll depths (50% and 90% depth)

I've been wondering why `event_name` isn't sent to GTM with each request. I assume that's handled under the hood when the configuration is initially transformed by the GTM?

### DIFF
--- a/frontend/templates/views/partials/analytics.j2
+++ b/frontend/templates/views/partials/analytics.j2
@@ -1,3 +1,13 @@
+{% if '/documentation/examples/' in doc.pod_path %}
+{% set category = 'examples' %}
+{% elif '/documentation/components/' in doc.pod_path %}
+{% set category = 'components' %}
+{% elif '/documentation/courses/' in doc.pod_path %}
+{% set category = 'courses' %}
+{% elif '/documentation/guides-and-tutorials/' in doc.pod_path %}
+{% set category = 'guides-and-tutorials' %}
+{% endif %}
+
 {% set config = {
   'vars': {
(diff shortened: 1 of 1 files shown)


frontend/templates/views/partials/analytics.j2
```diff
@@ -1,38 +1,150 @@
...
+  'selector': '.ap-m-format-toggle-link',
+  'on': 'click',
+  'vars': {
+    'event_name': 'format-toggle',
+    'event_category': category,
+    'event_action': 'link'
+  }
+}}) %}
+
+{# Scroll #}
```
matthiasrohmer: We could move those out of the `{% if %}` as scroll depth is probably interesting for all our pages. @sebastianbenz?

sebastianbenz: I agree, we should have this for all

matthiasrohmer:
@sebastianbenz, enabled the scroll pings for all pages. Went with `other` for `event_category` in this case. LMK if you'd prefer another value.

PR #3795, opened 2020-04-03
Add Chinese Simplified Localization

hmsjy2017:
I have translated translations.md to Chinese Simplified.


CrystalOnScript:
Thank you for contributing! :)

PR #3798, opened 2020-04-04
Fixed a typo in the code example

alexandr-kazakov:


alexandr-kazakov:
@googlebot I signed it!

PR #3814, opened 2020-04-12
template, fix pages linking

elbakerino:
fix wrong links in travel templates landing page to result page

currently the links are without `.html`, and the result page is a 404: https://amp.dev/documentation/templates/preview/websites/travel/templates/travel-results.amp

This page with `.html` works: https://amp.dev/documentation/templates/preview/websites/travel/templates/travel-results.amp.html (but the APIs for that page are not there)


elbakerino:
@googlebot I signed it!

PR #3832, opened 2020-04-15
template, fix pages linking

elbakerino:
fix wrong links in travel templates result page to landing page, like #3814 - but now reverse.

checked all template files for other wrong links (missing `.html` after `.amp`), none found.


sebastianbenz:
Thanks!

PR #3847, opened 2020-04-21
add supported amp-soundcloud attributes

sheershoff:
Adds info on `data-color` and `data-secret-token` attributes


sheershoff:
@googlebot I signed it!

PR #3918, opened 2020-05-07
Add option to enable optimizer experiments

sebastianbenz:


sebastianbenz:
Nope - the configuration is just inconsistent and confusing 😎... (will be fixed when out of experimental)

PR #3933, opened 2020-05-11
Improvements to "import email" logic

fstanis:
This fixes a bug where `Content-Type` can contain more parameters (like encoding) and overall makes the content type parser more true to [RFC 2045](https://tools.ietf.org/html/rfc2045#page-12).


sebastianbenz:
Can you please run `npm run fix:node` to fix the lint errors.

PR #3955, opened 2020-05-14
Lorem ipsum in prod amp-stiky-ad

vegagame:
Lorem ipsum in prod is not a go idea


matthiasrohmer:
That's actually intended 😉We want dummy content on the page to have it scrollable to demonstrate the behaviour of `amp-sticky-ad`. 

Thanks though for watching out 🙂

PR #3978, opened 2020-05-18
Update image dimensions to benefit from new srcset generation

sebastianbenz:
This makes use of the new AMP Optimizer srcset generation feature.
Templates need to be updated to use the mobile dimensions for width and
height.


sebastianbenz:
First experiment. Not really feasible as this triggers the image optimization on every page load.

platform/lib/utils/imageOptimizer.js
```diff
@@ -0,0 +1,80 @@
...
+ * limitations under the License.
+ */
+
+const path = require('path');
+
+const Image = require('@11ty/eleventy-img');
+
+const DEFAULT_IMG_FORMAT = 'jpeg';
+
+const baseDir = path.join(__dirname, '../../..');
```
matthiasrohmer, line 23: I'd suggest using the project helper to build the paths:

```js
const { project } = require('@lib/utils');

const baseDir = project.absolute('/');
```

Or for this specific case there's also:

```js
const baseDir = project.paths.ROOT;
```

platform/lib/utils/imageOptimizer.js
```diff
@@ -0,0 +1,80 @@
...
+const imageBaseDir = path.join(baseDir, 'pages');
+const outputDir = path.join(baseDir, 'dist/static/img');
+
+const globalOpts = {
+  urlPath: '/static/img/',
```
matthiasrohmer, line 28: Maybe add `formats` to make it obvious it's a supported option?

platform/lib/utils/imageOptimizer.js
```diff
@@ -0,0 +1,80 @@
...
+};
+
+/**
+ * Resizes the image specified at the given src URL to the given width. Images
+ * specified via an absolute URL will be downloaded and cached locally.
+ *
+ * @param {string} src - the original img's src URL
+ * @param {number} width  - the target width
+ */
+const imageOptimizer = async (src, width) => {
```
matthiasrohmer, line 39: Why not `async function imageOptimizer(src, width)`?

platform/lib/utils/imageOptimizer.js
```diff
@@ -0,0 +1,80 @@
...
+      widths: [width],
+    });
+    if (!isAbsoluteUrl(src)) {
+      src = path.join(imageBaseDir, src);
+    }
+    // Resizes, compresses the image (and download if needed).
+    const stats = await Image(src, opts); // eslint-disable-line
+    return stats[format][0].url;
+  } catch (e) {
+    console.log(`Could not optimize image (${src}, ${width}w):`, e);
```
matthiasrohmer, line 53: Use the log helper for scoping and consistency:

```js
const log = require('@lib/utils/log')('Image Optimizer');
log.warn(`Could not optimize image (${src}, ${width}w):`, e);
```

sebastianbenz:
@matthiasrohmer PTAL - I've removed the image optimization parts and only left the template improvements

PR #3993, opened 2020-05-25
🎉  Thumbor

matthiasrohmer:
Open tasks:

- [ ] Check if thumbor is running locally, only then register middleware
- [ ] Remove images from amp.dev image's static dest to not duplicate them
- [ ] Marry with #3978
- [ ] Add to Travis job
- [ ] Initial setup for prod on GCloud

### DIFF
--- a/platform/lib/routers/thumbor.js
+++ b/platform/lib/routers/thumbor.js
@@ -1,5 +1,5 @@
 /**
- * Copyright 2018 The AMP HTML Authors. All Rights Reserved.
+ * Copyright 2020 The AMP HTML Authors. All Rights Reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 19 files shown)


platform/lib/routers/thumbor.js
```diff
@@ -0,0 +1,51 @@
...
+  return join('/static/', '/**/', `*.${extension}`);
+});
+
+thumborRouter.get(imagePaths, (request, response, next) => {
+  // Thumbor expects SECURITY_KEY as URL partial
+  request.url = join(SECURITY_KEY, request.url);
+
+  proxy.web(request, response, proxyOptions, (error) => {
+    log.info('Proxy error', error);
+    response.status(502).end();
```
sebastianbenz, line 47: should we fallback to the local file instead? My idea was to encode the requested width via query param which would make it easy to fallback to the local filesystem.

matthiasrohmer: Exactly how it's done now 😉

matthiasrohmer:
Merging to see how it all ends up on staging. PR for production to follow.

sebastianbenz:
Nice!

PR #3994, opened 2020-05-25
🐛 Fix roadmap quarter nav timeline animation

robinvanopstal:


frontend/scss/components/templates/roadmap.scss
```diff
@@ -16,6 +16,8 @@
 
 .#{template('roadmap')} {
   $root: &;
+  --filter-height: 210px;
```
matthiasrohmer, line 19: Make this a SCSS variable instead.

PR #4018, opened 2020-05-29
⚡️ Task for pruning stale resources

matthiasrohmer:
My feeling was that there is a lot of dead code here therefore I attempted to write a task to eliminate some of it. Currently lists 373 unused files but might include a lot of false-positives for now.


matthiasrohmer:
Will be superseded by the overall frontend refactoring.

PR #4047, opened 2020-06-09
update dependencies

sebastianbenz:


sebastianbenz:
This also updates AMP Optimizer which required changing the way how we configure the hero image preload

Issue #4045, opened 2020-06-08
Update Events area to reflect what's happening as of June 2020


morsssss:
Our Events offerings have grown stale. Let's update them!

The Events menu currently contains:
* AMP Conf 2019
* AMP Contributor Summit 2019
* AMP Roadshow

Here's what we'd like to do:

**AMP Conf 2019**: leave it, because we don't have our digital 2020 event planned yet, and because this is a nice page full of lots of links to talks

**AMP Contributor Summit 2019**: replace this with a link to the 2020 event

**AMP Roadshow**: already updated to reflect COVID-19 postponements. Leave as is

Also:
* We should make a banner announcing the Contributors Summit.
* We should make a new page announcing the online "events" we're doing in 2020
* We should reorder the menu links so that newer things are first

/cc @sebastianbenz , @CrystalOnScript , @alexduran4 , @patrickkettner

sebastianbenz:
LGTM - thanks Ben! What would be the title for the new events page?

morsssss:
I've got a few ideas. I think we'll know more as we flesh out its content.

Just to save time, do you know how to update the banner? That shouldn't take too long to do.

sebastianbenz:
You can update the banner here https://github.com/ampproject/amp.dev/blob/future/frontend/templates/views/partials/banner.j2

matthiasrohmer:
At least we had a banner 😁

morsssss:
Reopening so that we remember to do the other things in this issue!

PR #4053, opened 2020-06-09
Add banner for OpenJS Collaborator Summit

morsssss:
/cc @mrjoro 

The banner is somewhat long. It takes up two lines on mobile. I tried to compress this rather complex message... then again, I think the previous banner probably wrapped on mobile too :)


morsssss:
This fixes one of the issues in #4045 .

PR #4071, opened 2020-06-12
Shorter banner to avoid layout issues

morsssss:
Turns out that, on smaller devices, the previous banner wrapped to 3 lines, causing the text to overflow the banner area, which is fixed to 31px.

I tried making the banner height not fixed, but of course that doesn't play well with the fact that the hamburger menu icon is `position:fixed`.

To allow more text on smaller devices, we could do any of the following:
* fix all the header CSS so that the banner could change height with impunity
* make the font smaller on smaller devices. (0.75rem looked great.)
* just warn people not to make the banner too large 🙃

For now, I'm just shortening the message, even though we lose the date. But others may have a better idea!


morsssss:
(thanks to @kristoferbaxter for pointing this out)

morsssss:
Right! Yes, when I realized that the header was also sticky, that's when I figured I wouldn't try to mess with any of the intricately assembled header :)

Thanks for merging this!

PR #4093, opened 2020-06-17
added sender distribution guidelines for email

CrystalOnScript:
We're hoping to have this live before EOD Monday - does that align with the next amp.dev deploy @matthiasrohmer ?


sebastianbenz:
Yes, will go live before then.

sebastianbenz:
Deploying is running and will be live in 45min.

>

PR #4112, opened 2020-06-22
♻️  Queue requests to GitHub API

### LINKED ISSUE #4092: Trouble importing all of the documents
I will have to separate these commits out a little better. I had trouble importing all of the documents. Not all of them came in when running `npm run bootstrap` or even by simply running `npx gulp importAll`. I think the flaky GitHub downloads leads to winding up w/ documents missing — a bit of a hassle. :/ Anyways, I'll get back to this in a short while. Thank you for your patience.

_Originally posted by @DerekNonGeneric in https://github.com/ampproject/amp.dev/pull/3346#issuecomment-644847783_

matthiasrohmer:
Fixes #4092. But also increases import times 🙁

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,6 +104,8 @@ This command enables the import from GitHub to run flawlessly. The actual import
 $ npm run bootstrap
 ```
 
+**Tip**: Due to bad network conditions or GitHub's API rate-limiting there might be errors during import. Try running the above command with the `-- --queue-imports` flag to prevent them.
+
 You can then start developing in your local environment with the command below. The task will take care of building and copying all files, watching them for changes, and rebuilding them when needed. Beware that changes to the [Express](https://expressjs.com/) backend require the Gulp task to be restarted.
 
 ```sh
--- a/package-lock.json
+++ b/package-lock.json
@@ -17784,8 +17784,7 @@
     "p-finally": {
       "version": "1.0.0",
       "resolved": "https://registry.npmjs.org/p-finally/-/p-finally-1.0.0.tgz",
-      "integrity": "sha1-P7z7FbiZpEEjs0ttzBi3JDNqLK4=",
-      "dev": true
+      "integrity": "sha1-P7z7FbiZpEEjs0ttzBi3JDNqLK4="
     },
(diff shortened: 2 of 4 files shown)


matthiasrohmer:
Converting this back to a draft as neither Travis, @patrickkettner or @sebastianbenz are affected by #4092 and therefore I'd like to put queueing behind a flag and only advise users to use it if there were errors during the import.

matthiasrohmer:
@DerekNonGeneric, can you please check if this does the trick for you?

DerekNonGeneric:
There are a couple of reasons why they were not able to reproduce this bug.

1. The documents had incorrect paths when they were imported from `ampproject/amphtml` and were corrected later.

I wonder if the Grow extension that checks URLs can be run in the `ampproject/amphtml` repo too. That way bad URLs never get committed, which inevitably ends up needing to be dealt with on this end (by making a PR to `ampproject/amphtml` with the correction).

2. My reproduction steps were not carried out on a fresh clone of the repo.

This would be why they did not encounter the errors that result from the rate-limiting problem because all of the documents were already imported, hence no missing documents and no broken URLs.

> @DerekNonGeneric, can you please check if this does the trick for you?

Yes, I will check that out now. Thanks for this!

DerekNonGeneric:
> @DerekNonGeneric, can you please check if this does the trick for you?

I just gave it a spin and it does indeed work correctly now (although I had to run `npm run bootstrap` twice due to some rate-limiting failure the first time). I appreciate the effort towards making this as clean as possible. I wish we didn't have to use the GitHub API for this, the rate-limiting is the most annoying thing. Anyways, `npm run lint` does pass successfully now, so I will continue w/ the original Markdown linting PR now. There's not much left to do there.

To satisfy my own curiosity, I will look into finding an alternative to the GitHub API since I have a feeling that would be the only viable solution to this — will keep y'all posted if I find a better solution than what is proposed in this PR, but in the meantime, this PR seems to be better than what currently exists. 🤷‍♂️

matthiasrohmer:
Great, thanks for verifying! This is then ready for review.

PR #4125, opened 2020-06-26
Remove banner

morsssss:
Just removing the banner, now that the event is over!


jaygray0919:
> 
> 
> Shouldn't break anything 😉

Am planning to add this comment to a list of famous quotes. Date and location noted for proper citation using BibTeX.

morsssss:
@jaygray0919 😎 

I removed my earlier comment saying I had trouble building it and that then I couldn't see the change... because I succeeded in getting the build to work, and then I figured out why I couldn't see the change 🙄

PR #4133, opened 2020-06-30
✨Add input bar component for playground

robinvanopstal:


playground/src/import-url/import-url.js
```diff
@@ -89,13 +87,13 @@ class ImportURL extends FlyIn {
         this.importError(e);
       })
       .finally(() => {
-        this.urlBarSubmit.classList.remove('loading');
+        this.inputBar.toggleLoading();
```
matthiasrohmer, line 90: Hm... this might be safe as the submit button is guarded against double submits - but in a worst case scenario this could start loading again.

playground/src/input-bar/input-bar.scss
```diff
@@ -0,0 +1,57 @@
+@import 'base.scss';
```
matthiasrohmer, line 1: Noooo. That's the reason we've got the following in the playground CSS a dozen times:

https://github.com/ampproject/amp.dev/blob/38fd7ef3dba9ebc6b822b21308a0ec5f9282fa07/frontend/scss/base.scss#L7-L9

Already removed it for the other components in an upcoming PR.

PR #4135, opened 2020-06-30
Create amp-autocomplete-email sample

### LINKED ISSUE #4060: Update amp-autocomplete documentation for email
Including:
- supported subset of attributes (modifications in AMPHTML) - Done in [amphtml#28898](https://github.com/ampproject/amphtml/pull/28898)
- email-specific examples (modifications in amp.dev)

This issue is filed in this repo because the work of adding email-specific examples on amp.dev is not straightforward:
There is on attribute `filter` which is required on the web format and banned in the email format. That is, an example that validates for email exactly does not validate for web and vice versa.

Possible approaches:
- Modify one file that has mutually exclusive code samples via the format filter
- Document via two different files, much like the AMP actions page (preferred)

Open question:
- When documenting via two different files, what should the new file name be?

cc @CrystalOnScript

caroqliu:
Fixes #4060 

~Note: The validator errors with `Invalid URL protocol 'http:' for attribute 'src'` because the `<% hosts.platform %>` substitutes for `http:` rather than `https:` -- is it possible to address this to allow the sample to validate in the AMP4Email playground?~

### DIFF
--- a/examples/api/autosuggest.js
+++ b/examples/api/autosuggest.js
@@ -21,7 +21,6 @@ const upload = multer();
 
 // eslint-disable-next-line new-cap
 const examples = express.Router();
-const MAX_RESULT_SIZE = 4;
 const US_CAPITAL_CITIES = [
   'Montgomery, Alabama',
   'Juneau, Alaska',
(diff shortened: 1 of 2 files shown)


sebastianbenz:
> Note: The validator errors with Invalid URL protocol 'http:' for attribute 'src' because the <% hosts.platform %> substitutes for http: rather than https: -- is it possible to address this to allow the sample to validate in the AMP4Email playground?

Is this a problem outside of development mode? This should be replaced by `https://amp.dev` in prod.

caroqliu:
> Is this a problem outside of development mode? This should be replaced by `https://amp.dev` in prod.

Ah good catch, this should be fine then. :)

PR #4137, opened 2020-07-01
✨Improve UX for import-url- and state view in playground

robinvanopstal:


playground/src/app.js
```diff
@@ -80,6 +81,10 @@ events.subscribe(ErrorList.EVENT_ERROR_SELECTED, (error) =>
   editor.setCursorAndFocus(error.line, error.col)
 );
 
+// configure background layer to close fly ins
+const flyInBackground = document.getElementById('fly-in-background');
+FlyInBackground.createFlyInBackground(flyInBackground);
```
matthiasrohmer, line 86: I started to refactor statements like this to the following:

https://github.com/ampproject/amp.dev/blob/3871b0b59494633d2ef7b2b3e4ff796cbea7e114/playground/src/app.js#L69

https://github.com/ampproject/amp.dev/blob/3871b0b59494633d2ef7b2b3e4ff796cbea7e114/playground/src/error-list/error-list.js#L26-L35

This cleans up app.js while at the same time allowing us to do specific case switches. Also almost all of our components are singletons so it doesn't make much sense to keep the factory functions configurable.

playground/src/fly-in/base.js
```diff
@@ -41,20 +47,12 @@ class FlyIn {
...
-    this.target.classList.toggle('active');
-  }
-
-  hideFlyIn() {
-    this.target.classList.remove('active');
-  }
-
-  showFlyIn() {
-    this.target.classList.add('active');
+  toggleFlyIn(force) {
```
matthiasrohmer, line 54: Why not simply `toggle` to get rid of the duplicate wording? The same applies to `createFlyIn`.

playground/src/fly-in/fly-in-background.js
```diff
@@ -0,0 +1,36 @@
...
+
+export const EVENT_FLY_IN_CLOSE = 'event-fly-in-close';
+
+class FlyInBackground {
+  constructor(target) {
+    target.addEventListener('click', () => {
+      events.publish(EVENT_FLY_IN_CLOSE);
+    });
+
+    document.addEventListener('keydown', (e) => {
```
matthiasrohmer, line 30: Nice! 👌

PR #4138, opened 2020-07-01
✨  Create Validator based on Playground

matthiasrohmer:
Work in progress, for @robinvanopstal to keep track of the changes for other related Playground work.

- Refactored `EmbedModes` to `Modes` for all components to distinguish wether the Playground is running embedded, as validator or in default mode
- Refactored `URLImport` and `ErrorList` to `FlyIn*` and `Inline*` variants and put related logic in its own class
- Removed `base.scss` from all components except `app.critical.scss` to get rid of 15 times `* { box-sizing: border-box }`
- Started decoupling hard dependencies in app.js by using events and stop passing elements into the factory functions


matthiasrohmer:
This is now ready for review. Easily testable with `https://amp.dev/shared/header?optimize=true` or `https://amp.dev/shared/header?optimize=false`.

The following header update is still missing but I'd like to do this in an upcoming PR as this one is already fairly big.

Bildschirmfoto 2020-07-02 um 14 44 44

playground/src/button/button.js
```diff
@@ -15,6 +15,10 @@
 import buttonTemplate from './button.hbs';
 
 export function from(element, onClickHandler) {
+  if (!element) {
+    return;
```
sebastianbenz, line 19: Shouldn't this throw an exception?

playground/src/fly-in/fly-in-background.js
```diff
@@ -14,11 +14,18 @@
 
 import './fly-in-background.scss';
 import events from '../events/events.js';
+import modes from '../modes/index.js';
 
 export const EVENT_FLY_IN_CLOSE = 'event-fly-in-close';
 
+export function createFlyInBackground() {
```
sebastianbenz, line 21: This is a bit weird as it now always will create something

playground/src/modes/validator.critical.scss
```diff
@@ -0,0 +1,64 @@
...
+    @media (min-width: 786px) {
+      padding: 55px 5px 55px 55px;
+    }
+
+    @media (min-width: 1000px) {
+      padding: 55px;
+    }
+  }
+
+    h1 {
```
sebastianbenz, line 45: nit: formatting? Shouldn't prettier handle this?

playground/src/preview/preview.js
```diff
@@ -27,10 +28,10 @@ const PARAM_HEIGHT = 'height';
 const MOBILE_BREAK_POINT = 767;
 
 export const EVENT_AMP_BIND_READY = 'event-amp-bind-ready';
-
 export const EVENT_AMP_BIND_NEW_STATE = 'event-amp-bind-new-state';
 
 export function createPreview(container) {
+  if (!container) return;
```
sebastianbenz, line 34: Same as with the others. It's a bit weird to have a method called create which doesn't create something.

matthiasrohmer:
@sebastianbenz, yep both are valid points. I wanted to avoid guarding every other bit in app.js with `if (modes.*) { ... }` that's why I took the shortcut with the simple `return` for the button. But I plan to tidy this up again when everything is in place.

Already forwarded the visual feedback to Ferdi 🙂

playground/src/input-bar/input-bar.hbs
```diff
@@ -1,3 +1,7 @@
+{{#helpText}}
```
robinvanopstal, line 1: A simple if statement would increase readability here

playground/src/modes/validator.critical.scss
matthiasrohmer: We didn't configure prettier for SCSS yet but I formatted it anyway 😁

playground/src/fly-in/fly-in-background.js
matthiasrohmer: You probably reviewed a split-second too early, this is now also guarded by a mode-check.

playground/src/input-bar/input-bar.hbs
matthiasrohmer: Thanks!

matthiasrohmer:
@sebastianbenz does this also have your approval? I fixed the blue triangle to never cut the small text - Ferdi wasn't too eager about its angle.

I plan to revert https://github.com/ampproject/amp.dev/pull/4138#discussion_r448992956 and https://github.com/ampproject/amp.dev/pull/4138#discussion_r448988633 in an upcoming PR by moving the creation of the affected elements out of app.js to their related components.

PR #4140, opened 2020-07-02
✨ Highlight amp-state changes in playground state tree view

robinvanopstal:
Similar to the Chrome Dev tools, state changes are now highlighted

amp-state-view


playground/src/state-view/state-view.scss
```diff
@@ -1,4 +1,20 @@
+@import 'base.scss';
+
+@keyframes HIGHTLIGHT {
```
matthiasrohmer, line 3: -T

sebastianbenz:
OMG!!!!

Issue #4157, opened 2020-07-07
Playground error list: fly-in hides code 


sebastianbenz:
The current fly in implementation hides the code area as part of the generic fly in lightbox behavior. In the case of the error list, this should not happen as users need to be able to see the code when looking at errors.

//cc @matthiasrohmer @robinvanopstal

PR #4158, opened 2020-07-07
🐛 Only show fly-in background if lightbox class is added

robinvanopstal:
Fixes: #4157

### DIFF
--- a/playground/src/fly-in/fly-in-background.scss
+++ b/playground/src/fly-in/fly-in-background.scss
@@ -14,7 +14,7 @@
   transition: opacity .25s ease-in-out
 }
 
-.lightbox.active ~ #fly-in-background {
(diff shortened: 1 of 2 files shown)


playground/src/fly-in/fly-in-background.scss
```diff
@@ -14,7 +14,7 @@
   transition: opacity .25s ease-in-out
 }
 
-.active ~ #fly-in-background {
+.lightbox.active ~ #fly-in-background {
```
matthiasrohmer, line 17: Hm... we are using both: simple state classes (`.active`, `.open`, ...) and BEM modifiers (`.*--left`, `.*--hidden`). Shouldn't we commit and then stick to one of them?

PR #4189, opened 2020-07-16
🔒  Harden playground/validator fetch API

matthiasrohmer:
@sebastianbenz can you provide guidance on setting sensible limits? Currently requests to `*.amp.dev` would not be limited at all, all other hosts are allowed to be requested 10 times in 10 seconds before requests are blocked for 10 seconds.

Fetch requests are aborted as soon as file size is over 100KB - documents this size would probably render CodeMirror useless anyway.

All responses that don't have `content-type: text/html` set are also blocked.

### DIFF
--- a/playground/backend/api.js
+++ b/playground/backend/api.js
@@ -50,6 +50,12 @@ const RATE_LIMIT_TIME_FRAME = 10 * 1000;
 const HOST_RATE_LIMIT = 10;
 
 /**
+ * The maximum count of limits that is tracked.
+ * @type {Number}
+ */
+const MAX_LIMITS = 500;
+
+/**
  * Fetches a user-defined remote URL and returns the response,
  * verifies that the returned response is a proper HTML document
  *
@@ -59,6 +65,7 @@ const HOST_RATE_LIMIT = 10;
 async function fetchDocument(fetchUrl) {
   const response = await fetch(fetchUrl, {
     size: MAX_FILE_SIZE,
+    max: MAX_LIMITS,
     headers: {
       'Accept': 'text/html',
       'x-requested-by': 'playground',
--- a/playground/backend/api.js
+++ b/playground/backend/api.js
@@ -1,5 +1,5 @@
 /**
- * Copyright 2019 The AMPHTML Authors
+ * Copyright 2020 The AMPHTML Authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


playground/backend/api.js
```diff
@@ -56,7 +69,82 @@ async function doFetch(url) {
...
+  }
+
+  // Verify that this URL is currently allowed to be fetched
+  // and is not rate-limited
+  if (exceedsRateLimit(fetchUrl)) {
+    response
+      .status(429)
+      .send(
+        `${fetchUrl.host} has been requested too many times. ` +
+          `Please wait a few seconds and then try again.`
```
matthiasrohmer, line 135: @sebastianbenz, how transparent do we want to be towards the user? Rather just say something like "Request failed try again later"?

playground/backend/api.js
```diff
@@ -56,7 +69,82 @@ async function doFetch(url) {
       'Referer': 'https://amp.dev/playground',
```
sebastianbenz, line 69: Wondering whether it'd be better to use a mobile user agent.

playground/backend/api.js
sebastianbenz: this sounds good to me

playground/backend/api.js
```diff
@@ -16,36 +16,49 @@
...
+/**
+ * Fetches a user-defined remote URL and returns the response,
+ * verifies that the returned response is a proper HTML document
+ *
+ * @param  {String} fetchUrl
+ * @return {String|undefined}
+ */
+async function fetchDocument(fetchUrl) {
+  const response = await fetch(fetchUrl, {
+    size: MAX_FILE_SIZE,
```
sebastianbenz, line 61: Good one!

playground/backend/api.js
```diff
@@ -56,7 +69,82 @@ async function doFetch(url) {
...
+  const contentType = response.headers.get('content-type');
+  if (!contentType.includes('text/html')) {
+    throw new Error(`${fetchUrl} is no HTML document.`);
+  }
+
   return response.text();
 }
 
+const limits = new LRU({
+  maxAge: RATE_LIMIT_TIME_FRAME,
```
sebastianbenz, line 86: Should this have a maximum size?

matthiasrohmer: Have been pondering over this: one could flood the API with requests to random hosts to make a ongoing rate-limit fall out. But added a fairly high one with 1000 now.

playground/backend/api.js
matthiasrohmer: `Android`, `Nexus` and `Mobile` should be enough hints for most user-agent tests to count as mobile, no? Just wondering if we should update versions to prevent getting a "update your browser" page.

PR #4221, opened 2020-07-24
remove out dated translations

patrickkettner:
as we are actively translating the guides and tutorials catalog, and all existing translations are out of date, I removed the existing ones to reduce developer confusion.


sebastianbenz:
Did you check whether the guides area actually outdated?

patrickkettner:
no, this just removes them. I can check into ones that are newer than the newest commit for the canonical version

sebastianbenz:
Depending on how long it takes until we get the updated translations we should either keep all or only remove the outdated ones.

patrickkettner:
im not confident in my ability to judge outdated content. these are on track to be replaced by september, so I will close this and leave as is for the time being

PR #4222, opened 2020-07-24
remove request for help on banner shown when missing translations

patrickkettner:
__Pleaceholder issue to detemine correct path__

Since we are not crowdsourcing the upcoming translations, I want to do something about the banner.
In theory, the automated nature of gitlocalize means that there won't be pages that are untranslated, and therefore this message should never be shown. I am not sure if we want to keep this banner as a backup. 
If we _do_ keep it, I am not sure what the message would be - just an apology? a link to open an issue?


patrickkettner:
@sebastianbenz hows this wording?

PR #4231, opened 2020-07-28
✨ Playground: Adapt sharing dialog to new fly-in design

robinvanopstal:


playground/src/share-view/share-view.hbs
```diff
@@ -0,0 +1,8 @@
+<div class="share-view-intro">
+  <h4>Copy & paste unique URL</h4>
```
matthiasrohmer, line 2: Use `&amp;`

playground/src/share-view/share-view.hbs
```diff
@@ -0,0 +1,8 @@
...
+  <p>This unique URL contains all changes made up to that point. As you make further changes to the code, a new URL will be generated.</p>
+</div>
+<textarea id="share-url" class="share-view-url"></textarea>
```
matthiasrohmer, line 5: Does it make sense to set this to `disabled`?

playground/src/share-view/share-view.js
```diff
@@ -0,0 +1,75 @@
...
+import template from './share-view.hbs';
+
+import {Base64} from 'js-base64';
+import FlyIn from '../fly-in/fly-in.js';
+import * as Button from '../button/button.js';
+
+export function createShareView(editor) {
+  const target = document.getElementById('share-view');
+  const trigger = document.getElementById('share-toggle');
+  if (!target) return;
```
matthiasrohmer, line 25: Couldn't you exit early? If there's no toggle there's no need to initialize the view, is there?

playground/src/share-view/share-view.js
```diff
@@ -0,0 +1,75 @@
...
+  createShareUrl() {
+    const source = this.editor.getSource();
+    const shareUrl = new URL(window.location.href);
+    shareUrl.searchParams.delete('url');
+    shareUrl.hash = 'share=' + Base64.encode(source);
+
+    return shareUrl.toString();
+  }
+
+  onClick() {
```
matthiasrohmer, line 66: `onClick` is ambiguous - use `onClickCopy`

playground/src/share-view/share-view.js
```diff
@@ -0,0 +1,75 @@
...
+    this.textarea.select();
```
matthiasrohmer, line 67: Does this work cross-browser? For the copy-script-button, we additionally use `setSelectionRange` but I can't remember why.

```
scriptTag.select();
scriptTag.setSelectionRange(0, 99999);
```

playground/src/share-view/share-view.hbs
robinvanopstal: Nope, I think `disabled` would not let you copy it's content to the clipboard ..? I've added `readonly` instead.

playground/src/share-view/share-view.js
robinvanopstal: as discussed

PR #4248, opened 2020-07-30
✨ Add autocompletion for experiments fly-in

robinvanopstal:


playground/src/input-bar/input-bar.js
```diff
@@ -27,6 +34,82 @@ class Input {
...
+      this.updateAutocompleteOptions(options);
+    }
+  }
+
+  updateAutocompleteOptions(options) {
+    this.autocompleteOptions = options;
+    this.renderAutocompleteList(options);
+  }
+
+  renderAutocompleteList(options) {
```
matthiasrohmer, line 75: As this is triggered for every `keyup` it would be nice to optimize this: wouldn't it work to simply toggle visibility for filtered items?

playground/src/experiments/experiments.js
```diff
@@ -82,6 +107,10 @@ class Experiments extends FlyIn {
     }
   }
 
+  /**
+   * Reads activated experiments from local localStorage
```
matthiasrohmer, line 111: https://github.com/ampproject/amp.dev/pull/4248#discussion_r463094627

playground/src/input-bar/input-bar.js
matthiasrohmer: so ... no?

playground/src/experiments/experiments.scss
```diff
@@ -8,8 +8,8 @@
 
   .experiments-list {
     max-height: 70%;
-    padding: 0;
     overflow-y: scroll;
```
matthiasrohmer, line 11: Make this `overflow-y: auto` to prevent a scrollbar from showing when it is not needed.

Issue #4251, opened 2020-07-30
Spacing issues in examples


morsssss:
Working on [the `<amp-script>` examples page](https://amp.dev/documentation/examples/components/amp-script), I've hit a couple of snags:

### 1. Pretty-printing ###
I find it confusing that HTML attributes for `<script>` tags get automatically pretty-printed so that each is on a separate line - as that tends to mush into the JavaScript that people will need to focus on. @sebastianbenz showed me that I could defeat this behavior by adding `<script>` to `BEAUTIFY_OPTIONS.unformatted[]` in `DocumentParser.js'`:

```javascript
const BEAUTIFY_OPTIONS = {
  indent_size: 2,
  "wrap_attributes": "force",
  unformatted: ['noscript', 'style', 'head', 'script'],
  'indent-char': ' ',
  'no-preserve-newlines': '',
  'extra_liners': []
};
```

However, this is having no effect.

### 2. Indentation ###
The pretty-printing normally works beautifully for JavaScript in examples. However, I'm having spacing problems in cases where I need to wrap the whole example in a `<div>` - since AFAIK a single example can't have multiple parent tags, but needs to have a single parent tag, which then gets magically removed.

So, this code:
```html
    <div>
      <amp-script layout="fixed-height" height="36" script="time-script" class="sample">
        <div>The time is: <span id="time"></span></div>
      </amp-script>

      <script id="time-script" type="text/plain" target="amp-script">
        const fetchCurrentTime = async () => {
          const response = await fetch('<% hosts.platform %>/documentation/examples/api/time');
          const data = await response.json();
          const div = document.getElementById('time');
          div.textContent = data.time;
        }

        fetchCurrentTime();
      </script>
    </div>
```

gets formatted like this:

![image](https://user-images.githubusercontent.com/29419498/88965308-91aa7f80-d278-11ea-968e-a98d6cc25c09.png)

morsssss:
@matthiasrohmer , I wonder if this is something else we can look at after Pixi?

patrickkettner:
@morsssss can you explain 1 more? All of the `<script>`s are one lined on the linked page. Can you share a screenshot of an affected example

morsssss:
Yeah, you're right. It must have started working at some point, because even in the screenshot I included here, it works 🙄

Sorries!

It would be great to fix #2, though. Until we do, people will think that we don't know how to format JavaScript 😬

patrickkettner:
[...]

PR #4257, opened 2020-07-31
Overhaul of <amp-script> samples

morsssss:
Updated existing samples to match latest ways of educating folks about <amp-script> and to match changes in the way it works. Also made them each load in the playground independently.  And I added one new sample. Next I'll be adding some more...

Tried to keep <script> tags from being auto-beautified, without success.

We've now got 4 script hashes. I'm debating whether we should drop the script hashes and instead use `data-ampdevmode`. But I kind of feel like we should just keep adding script hashes to show how to do it.

### DIFF
--- a/examples/source/1.components/amp-script/index.html
+++ b/examples/source/1.components/amp-script/index.html
@@ -1,6 +1,18 @@
+<!---
+
+standaloneSnippets: true
+preview: true
+author: sebastianbenz, morss
+
+--->
+
 <!--
-  An [amp-script](/content/amp-dev/documentation/components/reference/amp-script.md) hello world example.
+  ## Introduction
+
+  The `amp-script` component allows you to run custom JavaScript.
+  Your code runs in a Web Worker, and certain restrictions apply.
 -->
+
 <!-- -->
 <!doctype html>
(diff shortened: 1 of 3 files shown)


examples/source/1.components/amp-script/index.html
```diff
@@ -10,158 +22,225 @@
...
-        if (/iPad|iPhone|iPod/.test(userAgent) && !window.MSStream) {
-          return "iOS";
+
+        fetchCurrentTime();
+      </script>
+    </div>
+    <!-- 
+        ## Showing live data
+
+        You can also use `setInterval()` or `setTimeout` to get fresh data.
```
sebastianbenz, line 157: I'd love to have an websockets sample as well

morsssss: That was on my list of samples to create. I'm hesitating because there's already an example on the documentaiton page. But we can totally do it. If so, perhaps that can be used for a live-blogging or stock price example.

sebastianbenz: Messaging would be a great example!

morsssss:
Note also #4251 .

PR #4261, opened 2020-07-31
Larger scrollbars for samples

morsssss:
I had trouble with scrolling the 4px scrollbars. How about 7px?

I never succeeded in seeing this change happen in my dev environment, but this CSS change did work on the production site. Please let me know if I'm doing this incorrectly....


matthiasrohmer:
@robinvanopstal, can you please review this? I think we are not using the mixin across the whole codebase yet, but I am not quite sure.

matthiasrohmer:
Replaced by https://github.com/ampproject/amp.dev/pull/4268.

PR #4262, opened 2020-07-31
fix rendering issue on RU version of amp-lightbox docs

### LINKED ISSUE #3659: Component page: amp-lightbox, does not work in translation
## What's the issue?

If you go to the page: https://amp.dev/documentation/components/amp-lightbox/?format=websites and switch the language in the site header, then the page will stop working. I tested it in Russian.

I have seen this bug many times, on different days, it is 100% there. I recorded a short video of what it looks like for me: http://bit.ly/2It6nbU To the browser console shows that the entire page is covered by amp-lightbox :smile: that's why you can't click anywhere on the page.

If you don't see this bug, try the following:

1. Go to the main page and switch to English in the site header.
2. In the site header, go to the components page. Next, in the sidebar on the left, select amp-lightbox.
3. Switch to Russian in the site header. Switch to Russian in the site header. In this case, I have 100% stops working.

## What browsers are affected?

Tested in different browsers from the computer, at different times. 

## Which AMP version is affected?

Version 2002251816300

patrickkettner:
fixes #3659

The page was not [valid AMP](https://validator.ampproject.org/#url=https%3A%2F%2Famp.dev%2Fru%2Fdocumentation%2Fcomponents%2Famp-lightbox%2F%3Fformat%3Dwebsites), the snippet intended to be a code sample was being rendered by the docs page.

### DIFF
--- a/pages/content/amp-dev/documentation/components/reference/amp-lightbox-v0.1@ru.md
+++ b/pages/content/amp-dev/documentation/components/reference/amp-lightbox-v0.1@ru.md
@@ -91,7 +91,7 @@ teaser:
         <td width="40%"><strong>scrollable (необязательно)</strong></td>
         <td>Наличие атрибута <code>scrollable</code> означает, что содержимое окна просмотра можно прокручивать, если оно выходит за рамки окна.
           <br><br>
-            <strong>Примечание.</strong> Атрибут <code>scrollable</code> нельзя использовать с компонентом <code><amp-lightbox></code> в объявлениях HTML с технологией AMP. Подробную информацию можно найти в <a href="#a4a">этом разделе</a>.</td>
+            <strong>Примечание.</strong> Атрибут `scrollable` нельзя использовать с компонентом `&lt;amp-lightbox&gt;` в объявлениях HTML с технологией AMP. Подробную информацию можно найти в <a href="#a4a">этом разделе</a>.</td>
           </tr>
           <tr>
             <td width="40%"><strong>scrollable (необязательно)</strong></td>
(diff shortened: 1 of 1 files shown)


pages/content/amp-dev/documentation/components/reference/amp-lightbox-v0.1@ru.md
```diff
@@ -91,7 +91,7 @@ teaser:
         <td width="40%"><strong>scrollable (необязательно)</strong></td>
         <td>Наличие атрибута <code>scrollable</code> означает, что содержимое окна просмотра можно прокручивать, если оно выходит за рамки окна.
           <br><br>
-            <strong>Примечание.</strong> Атрибут <code>scrollable</code> нельзя использовать с компонентом <code><amp-lightbox></code> в объявлениях HTML с технологией AMP. Подробную информацию можно найти в <a href="#a4a">этом разделе</a>.</td>
+            <strong>Примечание.</strong> Атрибут `scrollable` нельзя использовать с компонентом `&lt;amp-lightbox&gt;` в объявлениях HTML с технологией AMP. Подробную информацию можно найти в <a href="#a4a">этом разделе</a>.</td>
```
matthiasrohmer, line 94: Hm... this only works via one of our markdown extensions but wouldn't in other environments than Grow. We should therefore probably stick to the default markdown syntax which doesn't allow markdown in HTML elements. So this would become:

```suggestion
            <strong>Примечание.</strong> Атрибут <code>scrollable</code> нельзя использовать с компонентом <code>&lt;amp-lightbox&gt;</code> в объявлениях HTML с технологией AMP. Подробную информацию можно найти в <a href="#a4a">этом разделе</a>.</td>
```

PR #4263, opened 2020-07-31
Another new amp-script sample

morsssss:
This one shows it's possible to personalize content to the user's language.


examples/source/1.components/amp-script/index.html
```diff
@@ -242,5 +249,38 @@
...
+          'pr': 'Olá'
+        };
+  
+        const lang = navigator.language.slice(0, 2);
+        const translation = translationMap[lang];
+  
+        if (translation) {
+          let greeting = document.getElementById('translated-greeting');
+          greeting.innerHTML = translation;
+        }
```
sebastianbenz, line 281: nit: show a message if language not supported to avoid confusion

examples/source/1.components/amp-script/index.html
```diff
@@ -242,5 +249,38 @@
...
+    <!--
+        ## Personalization
+        Similarly, you can use the `navigator` object, or other means, to personalize content for your user. The following script detects the browser's language and displays a localized greeting.
+    -->
+    <div>
+      <amp-script layout="fixed-height"
+        height="40"
+        script="translation-script"
+        class="sample">
+        <h2 id="translated-greeting">Saluton</h2>
```
sebastianbenz, line 262: I'd leave this empty by default to avoid a visible flickr

examples/source/1.components/amp-script/index.html
```diff
@@ -242,5 +249,38 @@
         span.textContent = getOS();
       </script>
     </div>
+
+    <!--
+        ## Personalization
```
sebastianbenz, line 254: Under this category I'd expect other samples (cookies,...)

examples/source/1.components/amp-script/index.html
morsssss: I chose to make the default message in Esperanto. I suppose no one's going to pick up on that one 😉

sebastianbenz: See my comment below. The current implementation visible switches from Esperanto to a different language which is a flicker effect we should avoid.

morsssss: Yep. Just removed the default text from this sample and from another.

PR #4268, opened 2020-08-03
💄 Adapt mixin for larger scrollbars

robinvanopstal:
Replaces #4261.

Hey @morsssss, the `@mixin scrollbar` was only used in the playground so far. I now updated the styling and made use of it throughout the whole project. To create variations (as seen in the screenshots below) the `@mixin` needs to be overwritten. I think for e.g. the `copy-script` it makes sense to leave the scrollbar at `height: 2px`? 

Bildschirmfoto 2020-08-03 um 16 08 19

Bildschirmfoto 2020-08-03 um 15 22 44

Bildschirmfoto 2020-08-03 um 15 15 07

### DIFF
--- a/frontend/scss/_mixins.scss
+++ b/frontend/scss/_mixins.scss
@@ -180,16 +180,17 @@ box-shadow: 0 25px 50px 0 rgba(0,0,0,0.21);
 }
 
 @mixin scrollbar {
-  ::-webkit-scrollbar {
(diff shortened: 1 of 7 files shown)


robinvanopstal:
> @robinvanopstal, as discussed offline: bump the copy-script scrollbar to 4px then we're good to go 🙂

Done !

PR #4287, opened 2020-08-11
Fix Sample code JSON syntax

ya-luotao:


ya-luotao:
@googlebot I signed it!

PR #4300, opened 2020-08-14
Pixi linter api

tharders:
Add pixi api that uses AMP linter to return the most serious status for each check.
Extracted the request loading logic including request limit checks from the playground to reuse it. 
Currently the "amp" mode is used until the "pageexperience" mode is released.

### DIFF
--- a/pixi/backend/api.js
+++ b/pixi/backend/api.js
@@ -31,7 +31,7 @@ const execLint = async (url) => {
   const body = await res.text();
   const context = {
     $: cheerio.load(body),
-    headers: res.headers,
+    headers: {},
     raw: {
       headers: res.headers,
       body,
--- a/pixi/backend/api.js
+++ b/pixi/backend/api.js
@@ -27,25 +27,40 @@ const remoteFetch = new LimitedRemoteFetch({
 });
 
 const execLint = async (url) => {
-  const res = await remoteFetch.fetchResponse(url);
-  if (res.ok) {
-    const body = await res.text();
-    const context = {
-      $: cheerio.load(body),
+  const res = await remoteFetch.fetchHtmlResponse(url);
+  const body = await res.text();
+  const context = {
+    $: cheerio.load(body),
+    headers: res.headers,
+    raw: {
       headers: res.headers,
-      raw: {
-        headers: res.headers,
-        body,
-      },
-      url,
-      mode: LintMode.Amp,
-    };
+      body,
+    },
+    url,
+    mode: LintMode.Amp,
+  };
+  return lint(context);
+};
 
-    return lint(context);
-  }
-  return Promise.reject(
-    new Error(`Url returned an error: ${res.statusText}`)
-  );
+const mapLintResult = (result) => {
+  return Object.entries(result).reduce((mappedData, [key, checks]) => {
+    if (Array.isArray(checks)) {
+      mappedData[key] = checks
+        .map((item) => item.status)
+        .reduce((result, status) => {
+          if (!status) {
+            return result;
+          }
+          if (!result || StatusNumber[result] < StatusNumber[status]) {
+            return status;
+          }
+          return result;
+        });
+    } else {
+      mappedData[key] = checks.status;
+    }
+    return mappedData;
+  }, {});
 };
 
 // eslint-disable-next-line new-cap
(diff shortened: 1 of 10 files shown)


pixi/backend/api.js
```diff
@@ -15,17 +15,78 @@
...
+      body,
+    },
+    url,
+    mode: LintMode.Amp,
+  };
+  return lint(context);
+};
+
+const mapLintResult = (result) => {
+  return Object.entries(result).reduce((mappedData, [key, checks]) => {
```
sebastianbenz, line 46: We should find a better way to merge result messages directly in linter.

I think it'd prefer to directly pass through the linter results. This way the result data needs to be handled in client (where it needs to be read anyway).

pixi/backend/api.js
```diff
@@ -15,17 +15,78 @@
...
-    .send(JSON.stringify({'message': 'Hello World'}, null, 2));
+
+  const fetchUrl = request.query.url;
+  try {
+    const lintResult = await execLint(fetchUrl);
+    const result = {
+      status: 'ok',
+      data: mapLintResult(lintResult),
+    };
+    response.status(200).send(JSON.stringify(result, null, 2));
```
sebastianbenz, line 79: Use `response.json(result)` this will automatically set status 200 and the correct content type.

pixi/backend/api.js
```diff
@@ -15,17 +15,78 @@
...
+  } catch (e) {
+    log.error('Unable to lint', fetchUrl, e.stack);
+    const result = {status: 'error'};
+    if (e.errorId) {
+      // The messages for the special RemoteFetchError can be shown in the response
+      result.errorId = e.errorId;
+      result.message = e.message;
+    }
+    response.status(200).send(JSON.stringify(result, null, 2));
```
sebastianbenz, line 88: same as above

platform/lib/utils/limitedRemoteFetch.js
```diff
@@ -0,0 +1,156 @@
+/**
```
sebastianbenz, line 1: Suggestion: rename file to `rateLimitedFetch`

platform/lib/utils/limitedRemoteFetch.js
```diff
@@ -0,0 +1,156 @@
...
+  }
+
+  /**
+   * Fetches a user-defined remote URL and returns the response.
+   * Verifies that the returned response is a proper HTML document.
+   *
+   * @param  {String} urlString
+   * @return {Response}
+   */
+  async fetchHtmlResponse(urlString) {
```
sebastianbenz, line 59: Please rename to `fetchHtml` (the response is implied) as part of the extraction.

pixi/backend/api.js
```diff
@@ -15,17 +15,78 @@
...
+const LimitedRemoteFetch = require('@lib/utils/limitedRemoteFetch');
+
+const remoteFetch = new LimitedRemoteFetch({
+  requestHeaders: {
+    'Referer': 'https://amp.dev/page-experience/',
+  },
+});
+
+const execLint = async (url) => {
+  const res = await remoteFetch.fetchHtmlResponse(url);
```
sebastianbenz, line 30: Why not use fetchDocument here?

tharders: We pass the req.headers to the linter context.
Even if the checks we do might currently not need them I would still do this, because we have the headers at hand and the linter api supports them.

sebastianbenz: good point!

platform/lib/utils/limitedRemoteFetch.js
tharders: `fetchHtml` would sound like it returns html. Html is in the name to make it clear that only `text/html` is accepted.
Maybe when we turn it around it would not sound so much like it returns html `htmlFetch`?
Would we also want to rename `fetchDocument` since it does not return a response like `fetch` does?
Maybe `loadContent`?

pixi/backend/api.js
tharders: Unless we have a new Result format that makes more sense than the old and does not hide any information I would not implement any mapping in the linter.

The middleware is explicitly implemented for pixi, so for me it makes sense to return the best possible format that the frontend can use. Why should we transfer more data than needed, or increase the client JavaScript if we can easily handle it in the backend. The mapping will not eat much server time.

platform/lib/utils/limitedRemoteFetch.js
sebastianbenz: Hm, makes sense. Then lets use `fetchHtmlResponse` and `fetchHtmlDocument` for consistency.

pixi/backend/api.js
sebastianbenz: My concern was not server time but maintainability. If we change a check result in linter we might have to touch code in backend and client as well. I'd prefer to only update a single point on that case. Is data really a concern here? If yes, it makes sense to pre-filter.

tharders: Data is not really a concern.
The mapping is not implemented check specific, so we only would have to update the backend if the linter result api changes, or if we want more information in the frontend, but in that case I feel it would be fine to touch frontend and backend.
The main point why I did not like to pass the linter result structure directly is that the status for each check can either be a result, or a list of results. Handling both of these makes the frontend code unnecessarily complex (I fear the frontend code will get quite complex without additional hurdles that we can avoid easily).
 
But thinking about your concern about maintainability, I would at least change the format of the mapped result so that we return an object with a `state` property for each check that can then be extended easily.

sebastianbenz: In that case, let's avoid pre-mature optimization and keep things simple by passing the result through. We can always revisit this later in case it's needed.

PR #4303, opened 2020-08-14
Adding socket server to API

morsssss:
Ok - I got stuck with this. I've never attached a socket server to express.js before, and I don't know enough about our existing infrastructure to figure out quite how to do it.

Please see the comments in `socket-server.js`. Some sort of magic seems to expect every JS file in `examples/api` to add a `Router` to the `express` object and then export it.

Thanks for your assistance!


sebastianbenz:
You already had it worked out. `examples/api` is simply the wrong place for a non middleware based router. The solution is to move the socket server into a separate subdir in examples (`examples/socket-server`).

morsssss:
Argh.

So it was simply a matter of avoiding the magic that expects everything in `examples/api` to be express middleware.

I moved it, and it seems to work now. One last question: at what route will  the socket server live?

examples/source/1.components/amp-script/index.html
```diff
@@ -158,6 +173,36 @@
         fetchCurrentTime();
       </script>
     </div>
+
+    <!--
+      ## Using a WebSocket for a live blog
```
sebastianbenz, line 178: Suggestion: make the title more generic (e.g. Live updates via sockets). Due to the resizing problem, live blogs are tricky with amp-script and I'm not sure that we should recommend this.

examples/source/1.components/amp-script/index.html
```diff
@@ -158,6 +173,36 @@
...
+      </amp-script>
+  
+      <script id="live-blog-script" type="text/plain" target="amp-script">
+        const button = document.getElementById('live-blog-start');
+        const blogDiv = document.getElementById('live-blog-area');
+  
+        button.addEventListener("click", () => {
+          button.setAttribute('disabled', '');
+          button.textContent = 'Live blog begun';
+          const socket = new WebSocket('ws://localhost:8080/documentation/examples/api/socket/live-blog');
```
sebastianbenz, line 194: This is not going to work once deployed. Use `<% host.platform %>` for the host.

examples/source/1.components/amp-script/index.html
```diff
@@ -158,6 +173,36 @@
         fetchCurrentTime();
       </script>
     </div>
+
+    <!--
+      ## Using a WebSocket for a live blog
+      `amp-script` supports WebSockets. This example simulates a live blog.
+    -->
+    <div>
+      <amp-script layout="fixed-height" height="200" script="live-blog-script" class="sample" sandbox="allow-forms">
```
sebastianbenz, line 182: Does all the content fit into 200px?

sebastianbenz (changes requested):

examples/source/1.components/amp-script/index.html
morsssss: Changed title.

examples/source/1.components/amp-script/index.html
morsssss: argh! I got distracted by all my trouble getting the route to work. Thanks for noticing this.

examples/source/1.components/amp-script/index.html
morsssss: I used a scrolling container. I'd make it bigger, but actually this is enough space on the page to see that this works, and how it works.

examples/source/1.components/amp-script/index.html
morsssss: Actually, do I have access to a `host` property that doesn't include the scheme? Otherwise I get an unwanted "http://"

morsssss: I didn't spot anything like that in `config.js` or the html files.

Is it possible to do something like this? (Though it's probably better to make url parts available to other html files via backend changes)

`<% hosts.platform.replace(/.+\/\//, "ws://") %>`

sebastianbenz: This would be the right place to add a new variable `hosts.websocket`.

platform/lib/platform.js
```diff
@@ -64,6 +65,9 @@ class Platform {
           log.success(`server listening on ${PORT}!`);
           resolve();
         });
+
+        webSocketServer.startLiveBlog(this.httpServer);
```
sebastianbenz, line 69: This is our only socket server, so I wouldn't make the start name specific to the liveblog sample.

examples/source/1.components/amp-script/index.html
sebastianbenz: The updated title has the same problem.

examples/source/1.components/amp-script/index.html
sebastianbenz: Not happy about this approach (but this is my general live blog concern). I'd preferred a different sample that better suits amp-script. But it's fine.

platform/lib/platform.js
morsssss: Ah, I was planning for future socket servers. I can change this if you like. (As apparently you do)

examples/source/1.components/amp-script/index.html
morsssss: I think it will inspire people to see that this sort of thing is possible. But I understand your concern!

platform/lib/platform.js
morsssss: Done

examples/source/1.components/amp-script/index.html
morsssss: Ok, did this. It also gave me a chance to learn more about how amp.dev's configuration is built and how that propagates into samples....

sebastianbenz (approved):
Looks good! Thanks! Please merge once the build is passing.

PR #4321, opened 2020-08-19
Add page information to pixi linter api

tharders:


pixi/backend/api.js
```diff
@@ -27,11 +27,24 @@ const rateLimitedFetch = new RateLimitedFetch({
   },
 });
 
-const execLint = async (url) => {
+const COMPONENT_SRC_MATCHER = /\/v0\/([^.]+)-(\d+(?:\.\d+)*)\.m?js/;
```
sebastianbenz, line 30: I'd prefer using Cheerio here as it's easier to maintain.

sebastianbenz: nvm - ignore my comment

pixi/backend/api.js
```diff
@@ -40,7 +53,14 @@ const execLint = async (url) => {
     url,
     mode: LintMode.Amp,
   };
-  return lint(context);
+  const lintResults = await lint(context);
+  return {
+    https: res.url.startsWith('https://'),
```
sebastianbenz, line 58: nit: I'd do this check in the client for consistency reasons.

PR #4332, opened 2020-08-20
✨ Add linter check and tests

robinvanopstal:


pixi/src/checks/LinterCheck.js
```diff
@@ -0,0 +1,55 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+const API_ENDPOINT = API_ENDPOINT_LINTER;
+
+export default class LinterCheck {
+  async run(pageUrl) {
+    const apiUrl = `${API_ENDPOINT}${pageUrl}`;
```
tharders, line 19: pageUrl might contain parameters. it has to be URL escaped when you build the apiURL manually.
Or you do it the same way as the PageExperienceCheck

pixi/config.js
```diff
@@ -1,11 +1,13 @@
 module.exports = {
   development: {
+    API_ENDPOINT_LINTER: 'http://localhost:8080/page-experience/api/lint/?url=',
```
tharders, line 3: Can we use a Links without the server part here?
We could also use it the same way for the production linter setting then and it would work for the staging environment as well.

tharders (changes requested):
Please take a look at the comments

pixi/config.js
robinvanopstal: Sadly no, amp-script requires fetched URLs to be absolute.

tharders (approved):
Fine for now. We should think about different prod and stage linter URLs later.

pixi/src/checks/LinterCheck.js
```diff
@@ -0,0 +1,64 @@
+// Copyright 2020 The AMPHTML Authors
```
sebastianbenz, line 1: Rename file to `AmpLinterCheck`

pixi/src/checks/LinterCheck.js
```diff
@@ -0,0 +1,64 @@
...
+
+    try {
+      const apiResult = await this.fetchJson();
+      return this.createReportData(null, apiResult);
+    } catch (e) {
+      return this.createReportData(e);
+    }
+  }
+
+  createReportData(error, apiResult) {
```
sebastianbenz, line 33: I'd split this into two different methods: `createReport` and `createError`. This simplifies the parameters and you avoid the additional if check. We could do the same for all the checks.

pixi/src/checks/LinterCheck.test.js
```diff
@@ -0,0 +1,44 @@
...
+} from '../../mocks/linterCheck/apiResponse.js';
+
+import pixiConfig from '../../config.js';
+
+beforeEach(() => {
+  fetchMock.reset();
+});
+
+describe('Linter check', () => {
+  const apiEndpoint = pixiConfig['development'].API_ENDPOINT_LINTER;
```
sebastianbenz, line 20: s/pixiConfig['development']/pixiConfig.development

sebastianbenz (approved):
Looks good once nits are fixed.

pixi/src/checks/LinterCheck.js
robinvanopstal: Implemented for this one for now

PR #4333, opened 2020-08-20
✨ Implement Mobile Friendliness Check

matthiasrohmer:
The Search Console API has a little variety of responses. See /mocks/mobileFriendliness/apiResponse.js.

For now I decided that everything that is not a hard fail is handled as passed or do we want to be more granular here?


pixi/mocks/mobileFriendlinessCheck/apiResponse.js
```diff
@@ -0,0 +1,59 @@
...
+const unspecifiedStatus = {
+  testStatus: {
+    status: 'TEST_STATUS_UNSPECIFIED',
+  },
+};
+
+const internalError = {
+  testStatus: {
+    status: 'INTERNAL_ERROR',
+  },
```
caroqliu, line 43: These should still have `mobileFriendliness: 'MOBILE_FRIENDLY_TEST_RESULT_UNSPECIFIED'` iiuc? Talked with @nainar and we settled on this resolving to `Manual check` similar to Intrusive Interstitials.

pixi/mocks/mobileFriendlinessCheck/apiResponse.js
```diff
@@ -0,0 +1,59 @@
...
+  mobileFriendliness: 'MOBILE_FRIENDLY',
+};
+
+const mobileFriendlyWithIssues = {
+  testStatus: {
+    status: 'COMPLETE',
+  },
+  mobileFriendliness: 'MOBILE_FRIENDLY',
+  resourceIssues: [
```
caroqliu, line 13: Looked into this more. There are two types of issues, `mobileFriendlyIssues` and [`resourceIssues`](https://developers.google.com/webmaster-tools/search-console-api/reference/rest/v1/urlTestingTools.mobileFriendlyTest/run#resourceissue). The former should fail the whole test so does not fall into this `mobileFriendlyWithIssues` edge case, but we should still pipe the findings through to the recommendation so they can improve the mobile friendliness status. For the latter we can note the status and the recommendation should be specific to [blocked resources](https://support.google.com/webmasters/answer/6352293#blocked-resources). /cc @nainar @sebastianbenz as there may be different copy for these.

PR #4338, opened 2020-08-20
🐛 Try to get URL from amp-state if not filled

matthiasrohmer:
Workaround for https://github.com/ampproject/worker-dom/issues/912


sebastianbenz (approved):
Even better!

PR #4362, opened 2020-08-24
Allow HTTPS check to report `Analysis failed`

caroqliu:
It looks like the HTTPS check always hangs because the AMP linter does not return a `url` property (surfacing as `Cannot read property 'startsWith' of undefined`), and the UI does not update when the error surfaces. For now this PR (1) updates the error message and (2) allows it to fail if `apiResult.url` is `undefined` /cc @tharders


matthiasrohmer (approved):
Thanks, Caroline! 🙂

PR #4355, opened 2020-08-21
✨Pixi recommendations

robinvanopstal:

PR #4367, opened 2020-08-24
Pixi: Cleanup runtime errors and edge cases

caroqliu:
- One common error is cannot read property `result` of `undefined`, due to error cases for the Safe Browsing and HTTPS checks. In this case I return an empty `data` object `{}` to allow the "Analysis Failed" UI to surface.
- Recommendations are hidden during loading. It also looks like some of the flattened recommendations contain a few `undefined` entries so I disregarded those explicitly. We should dig more where they are coming from and consider disregarding any entries that do not have our expected properties (currently `id`, `title`, and `description`).

### DIFF
--- a/pixi/src/ui/PageExperience.js
+++ b/pixi/src/ui/PageExperience.js
@@ -145,7 +145,7 @@ export default class PageExperience {
     if (error) {
       console.error('Could not perform AMP Linter check', error);
     }
-    this.reportViews.httpsCheck.render(data.usesHttps);
+    this.reportViews.httpsCheck.render(data.result);
 
     return data.recommendations;
   }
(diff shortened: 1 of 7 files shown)


pixi/src/ui/PageExperience.js
```diff
@@ -145,7 +145,7 @@ export default class PageExperience {
     if (error) {
       console.error('Could not perform AMP Linter check', error);
     }
-    this.reportViews.httpsCheck.render(data.result);
+    this.reportViews.httpsCheck.render(data.usesHttps);
```
robinvanopstal, line 148: Hey @caroqliu, wouldn't it be more consistent to stick with `data.result` here? I think I've missed adapting the changes [here](https://github.com/ampproject/amp.dev/pull/4355/files#r476282034)

caroqliu: Sg, I'll remove this change entirely as t his is fixed in #4372

matthiasrohmer (approved):

PR #4376, opened 2020-08-25
limit the number of significant figures on pixi lab data results

patrickkettner:
before:
image

after:
image

### DIFF
--- a/pixi/src/ui/report/CoreWebVitalsReportView.js
+++ b/pixi/src/ui/report/CoreWebVitalsReportView.js
@@ -30,9 +30,9 @@ class SimpleScale {
   render(data, unit) {
     this.bar.style.transform = `scale3d(${data.score}, 1, 1)`;
 
-    this.value.textContent = `${data.numericValue / unit.conversion}${
-      unit.name
-    }`;
+    this.value.textContent = `${(data.numericValue / unit.conversion).toFixed(
+      1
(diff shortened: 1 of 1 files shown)


pixi/src/ui/report/CoreWebVitalsReportView.js
```diff
@@ -30,7 +30,7 @@ class SimpleScale {
   render(data, unit) {
     this.bar.style.transform = `scale3d(${data.score}, 1, 1)`;
 
-    this.value.textContent = `${data.numericValue / unit.conversion}${
+    this.value.textContent = `${(data.numericValue / unit.conversion).toFixed(3)}${
```
caroqliu, line 33: Suggestion: We could even do `1` or `2`. I believe PSI does `1`.

matthiasrohmer (approved):

matthiasrohmer:
@patrickkettner, beware of the lint errors: https://travis-ci.org/github/ampproject/amp.dev/builds/721013513#L915 😉

pixi/src/ui/report/CoreWebVitalsReportView.js
tharders: Do we really always want to show 3 decimal places?
Instead of "3s" we would then also have "3.000s"

PR #4384, opened 2020-08-26
Pixi content structure

tharders:
The whole content will be available under /pixi/index.json
And translated versions under /LANGUAGE/pixi/index.json

Only a few recommendations and status banners are available at this point.
The others have to be added later.
Also the global text keys are far from final at this stage.


pages/content/pixi/index.md
```diff
@@ -0,0 +1,15 @@
+---
+$sitemap:
+  enabled: false
+globalText:
```
sebastianbenz, line 4: Can we find a better name than `globalText`? It's not clear to me what this means.

pages/content/pixi/index.md
```diff
@@ -0,0 +1,15 @@
...
+  analyzeHeadline: Analyze your AMP page
```
sebastianbenz, line 5: Should we nest instead of using a prefix?

```
text: 
  analyze:
    headline: Analyze...
    fieldPlaceholder: ...
```

pages/content/pixi/index.md
```diff
@@ -0,0 +1,15 @@
...
+  analyzeFieldPlaceholder: Enter your URL
+  analyzeButton: Analyze
+tags:
```
sebastianbenz, line 8: Is tags the right name here? Wouldn't these be used for headlines as well?

pages/content/pixi/index.md
tharders: Done

matthiasrohmer (approved):

pages/content/pixi/index.md
tharders: As discussed offsite its now "staticText"

pages/content/pixi/index.md
tharders: Moved the tags into the "staticText" since here they are only the translation for the tag IDs.

PR #4386, opened 2020-08-27
New amp-script samples: multiple fetches, and communicating with an AMP component

morsssss:
Check it out!

I was starting to add a sample for protocol adapters, but then as I tested I realized they're still behind a flag.


examples/api/slow-response.js
```diff
@@ -62,6 +64,18 @@ async function slowJsonWithItems(request, response) {
   );
 }
 
+// Note that this function expects its delay measured in seconds.
+async function slowText(request, response) {
+  const delay = getDelay(request);
+  console.log('delay is' + delay);
```
sebastianbenz, line 70: please remove

examples/source/1.components/amp-script/index.html
```diff
@@ -158,6 +164,37 @@
...
+        ## Multiple fetches
+
+        In a container whose size can change, your code can make DOM changes until 5 seconds after the final `fetch()` completes. This example makes multiple calls to a slow API. It displays the result from each call when it returns.
+    -->
+    <div>
+      <amp-script layout="container" script="multi-fetch-script" class="sample">
+        <button id="multi-fetch">How slow is our API?</button>
+      </amp-script>
+      <script id="multi-fetch-script" type="text/plain" target="amp-script">    
+        const randomTime = () =>  Math.floor(Math.random() * 10) + 5;
```
sebastianbenz, line 177: Why make it random? How about making 10 requests where each requests takes 1s longer. This way we can visualize the cut off.

examples/source/1.components/amp-script/index.html
```diff
@@ -158,6 +164,37 @@
...
+          for (let i =0; i < 3; i++) {
+            fetch('<% hosts.platform %>/documentation/examples/api/slow-text?delay=' + randomTime())
+              .then(response => response.text())
+              .then(insertText);
+          }
+        }
+    
+        function insertText(text) {
+          const div = document.createElement('div');
+          div.innerHTML = text;
```
sebastianbenz, line 191: nit: use `textContent` instead of `innerHTML`

examples/source/1.components/amp-script/index.html
```diff
@@ -282,6 +318,43 @@ <h2 id="translated-greeting"></h2>
...
       </script>
     </div>
+    <!--
+        ## Interacting with AMP components
+        Your script can communicate with an AMP component using a state variable.
+        Bind an attribute in the component to an expression containing a state variable.
+        When your script modifies that state variable, the change will propagate to the component.
+    -->
+    <div id="carousel-sample">
+      <amp-carousel
```
sebastianbenz, line 328: Can you simplify this example? We don't need so much markup just to demonstrate the point. A single image with a binding src value would be sufficient.

examples/source/1.components/amp-script/index.html
```diff
@@ -282,6 +318,43 @@ <h2 id="translated-greeting"></h2>
...
+      <p>Slide <span [text]="slideIndex + 1">1</span> of 3</p>
+      
+      <amp-script layout="container" script="carousel-script" class="sample">
+        <button id="carousel-button">Surprise me</button>
+      </amp-script>
+
+      <script id="carousel-script" type="text/plain" target="amp-script">
+        const button = document.getElementById('carousel-button');
 
+        function gotoRandomSlide() {
```
sebastianbenz, line 351: This also adds too much complexity. Simply switching between two images is enonugh.

sebastianbenz (commented):
Great samples! Left a few suggestions how to simplify things.

samouri:
> I was starting to add a sample for protocol adapters, but then as I tested I realized they're still behind a flag.

Its been delayed for a bit due to a P0 Ads issue that I believe was resolved today. 
I think it will make it to prod before the weekend.

examples/api/slow-response.js
morsssss: D'oh!

examples/source/1.components/amp-script/index.html
morsssss: it could not be random. I think the randomness part makes it more fun to play with. But also, to me, it implies more clearly how that this works - that your code can make mutations until 5 seconds after the last fetch returns... not, say, after you make your last fetch, which is kind of what is implied to me by a linearly increasing set of delays.

Of course, I could keep on making changes for 4 seconds after the last fetch returns, but that seems to me to much.

examples/source/1.components/amp-script/index.html
morsssss: whoops. old habit.

examples/source/1.components/amp-script/index.html
morsssss: To you and me, `<amp-img>` is an AMP component. But I think that to most people it's just a replacement for ``. And what I wanted to show here was that you can use `<amp-script>` to supercharge an existing AMP component - to give it new features. That people should really try to do this instead of creating components from scratch in JS.

Of course, your example is super clear. In the new documentation, I do something similar - I just show it's possible to use amp-script to modify text in a `<p>`.

Maybe I should include your example too - showing an image changing - for a simple example showing how amp-script can interact with the DOM via data binding. Because I don't show that. And then I could include my example to show how it's possible to modify a component's behavior. Thinking about it more, I should even put the "Surprise me" button on top of the carousel, so we really see the component has gained a new feature.

morsssss: I added the simpler sample showing an image changing. I think this is an improvement - thanks!

matthiasrohmer (approved):

PR #4403, opened 2020-08-31
Add isAmp check to linter api

tharders:
The linter api now first checks if the page is a declared AMP page and only if that is true the linter checks are executed.

This PR also has the imported release schedule updated, which I did not really want to include here, but should do no harm.

### DIFF
--- a/pixi/backend/api.js
+++ b/pixi/backend/api.js
@@ -38,6 +38,11 @@ const findAmpComponents = ($) => {
   return versionMap;
 };
 
+const isAmp = ($) => {
+  const ampHtml = $('html[amp],html[⚡]');
+  return ampHtml.length > 0;
+};
+
 const execChecks = async (url) => {
   const res = await rateLimitedFetch.fetchHtmlResponse(url);
   const body = await res.text();
@@ -52,12 +57,21 @@ const execChecks = async (url) => {
     url,
     mode: LintMode.PageExperience,
   };
-  const lintResults = await lint(context);
-  return {
+  const result = {
     redirected: res.redirected,
     url: res.url,
+    isAmp: isAmp($),
+  };
+
+  if (!result.isAmp) {
+    return result;
+  }
+
+  const lintResults = await lint(context);
+  return {
     components: findAmpComponents($),
     data: lintResults,
+    ...result,
   };
 };
 
--- a/pages/content/pixi/status-banners/no-amp.md
+++ b/pages/content/pixi/status-banners/no-amp.md
@@ -1,0 +1,6 @@
+---
+$title: This URL is not an AMP page.
+type: error
+---
+Sorry, this tool is designed for AMP pages only.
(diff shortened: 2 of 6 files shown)


pages/content/amp-dev/documentation/guides-and-tutorials/learn/spec/release-schedule.md
```diff
@@ -35,7 +35,7 @@ A new release of AMP is pushed to all AMP pages every week on Tuesday. **Once a
 
 The AMP runtime and extensions are provided through a variety of different _release channels_. Each channel serves a purpose for developers and for the AMP HTML Project itself. See the [release cadence section](#release-cadence) for a more detailed breakdown of how and when code from the [`ampproject/amphtml`](https://github.com/ampproject/amphtml) repository makes it into release builds.
 
-To determine if a PR has been included in any of the following release channels, look for the GitHub labels _PR Use: In Canary_, _PR Use: In Production_, or _PR Use: In LTS_ (see the section on [determining if your change is in a release](#Determining-if-your-change-is-in-a-release) for more details).
+To determine if a PR has been included in any of the following release channels, look for the GitHub labels _PR Use: In Canary_, _PR Use: In Production_, or _PR Use: In LTS_ (see the section on [determining if your change is in a release](#determining-if-your-change-is-in-a-release) for more details).
```
caroqliu, line 38: Are changes to this file a side effect of `gulp bootstrap`? If so let's remove since it is not relevant to the larger purpose of the PR.

pixi/backend/api.js
```diff
@@ -38,6 +38,11 @@ const findAmpComponents = ($) => {
   return versionMap;
 };
 
+const isAmp = ($) => {
```
caroqliu, line 41: We also want to check if it imports `.*\/v0.js`

tharders: In the amp-toolbox the "isAmp" implementation just checks for the amp attribute.
What is the purpose of this additional check?
Should it be added with "and" or with "or"?

If it is "or" I think it is a very weak check to see if a page looks like it should be AMP since the pattern could match scripts that are not AMP easily. I think it would be better to check for `<style amp-` and `<script custom-element="amp-`

If it should be "and" I don't really see the benefit of adding it. When a user hosts the amp runtime on his server and the name no longer matches he should get an error that the page is not considered valid AMP and not that the page is no AMP at all (while the rest of the page might literally scream "I am an AMP page" ;-)

pages/content/amp-dev/documentation/guides-and-tutorials/learn/spec/release-schedule.md
tharders: Ok, it is removed.

pixi/backend/api.js
caroqliu: Works for me.

PR #4416, opened 2020-09-01
✨ Pixi - Add sharing

robinvanopstal:


frontend/templates/views/partials/pixi/share-dialog.j2
```diff
@@ -0,0 +1,33 @@
...
+<amp-lightbox id="share-dialog" layout="nodisplay">
+  <div class="ap-m-share-dialog"
+      role="button"
+      tabindex="0">
+    <div class="ap-m-share-dialog-container">
+      <h2>{{ _('Copy & paste URL') }}</h2>
+      <input class="ap-m-share-dialog-container-textarea"
+          type="text"
+          value="{{ doc.url }}"
+          [value]="'{{ doc.url }}' + pixi.shareUrl"
```
caroqliu, line 12: Can we have a state variable with the full value so we do not have to construct the string in the markup? Ideally this would be `[value]="pixi.shareUrl"`, where `pixi.shareUrl` also does the following:
- Call `encodeURIComponent` on the given `pageUrl`
- Append the `url` parameter (write now this generates `http://localhost:8081/page-experience/https://amp.dev` but we want `http://localhost:8081/page-experience/?url=https://amp.dev`

matthiasrohmer (approved):

PR #4421, opened 2020-09-02
Pixi status banner rules

tharders:
Basic pixi status banner rules

Changed AmpLinterCheck to not throw when the url could not be loaded in order to handle this case like any other check.

Also added unit test for recommendations rules.

The status banner rules are currently not used in the frontend.

### DIFF
--- a/pixi/src/checkAggregation/statusBanner.js
+++ b/pixi/src/checkAggregation/statusBanner.js
@@ -17,7 +17,7 @@ export default async function getStatusId(
   pageExperiencePromise,
   safeBrowsingPromise,
   linterPromise,
-  mobileFriendlinessPromise,
+  mobileFriendlinessPromise
 ) {
   try {
     const linter = await linterPromise;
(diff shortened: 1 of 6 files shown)


pixi/src/checkAggregation/statusBanner.js
```diff
@@ -0,0 +1,57 @@
...
+  linterPromise,
+  mobileFriendlinessPromise
+) {
+  try {
+    const linter = await linterPromise;
+    if (!linter.isLoaded) {
+      return 'invalid-url';
+    }
+    if (!linter.isAmp) {
+      return 'no-amp';
```
caroqliu, line 28: What do you think about putting these all in an enum? i.e. `Status.NO_AMP`

tharders: Might give a better overview what states we have.
But as long as we use it only once and the logic does not get too big I think it's not that urgent.

PR #4427, opened 2020-09-02
Implement pixi checks with page experience audit data

tharders:
The recommendation mapping is refactored to be more clear and also mapps the new checks.
Changed the "resourceIssues" result to "noResourceIssues" so that all basic checks return true if all is fine.


pixi/src/checkAggregation/recommendations.js
```diff
@@ -30,38 +73,24 @@ export default async function getRecommendationIds(
     mobileFriendlinessPromise,
   ]);
 
-  const result = [];
+  const result = [...fixedRecommendations];
```
caroqliu, line 76: I prefer intrusive interstitials recommendation to be last if possible, since a developer who runs their page several times (while acting on recommendations) will only find the manual step helpful the first time.

tharders: Good point.
I think we should add an order property to the recommendation markdown files and then sort them before rendering.
That would be better than having an order here.
We will do this in a separate PR.

PR #4429, opened 2020-09-02
add a version to pixi api response

### LINKED ISSUE #4428: [pixi] add version to API response
We should add a version to the API response, so that any consumer could be aware of breaking changes

patrickkettner:
fixes #4428

### DIFF
--- a/pixi/backend/api.js
+++ b/pixi/backend/api.js
@@ -19,6 +19,7 @@ const {lint, LintMode} = require('@ampproject/toolbox-linter');
 const cheerio = require('cheerio');
 const log = require('@lib/utils/log')('Pixi API');
 const RateLimitedFetch = require('@lib/utils/rateLimitedFetch');
+const API_VERSION = 0;
 
 const rateLimitedFetch = new RateLimitedFetch({
   requestHeaders: {
@@ -61,6 +62,7 @@ const execChecks = async (url) => {
     redirected: res.redirected,
     url: res.url,
     isAmp: isAmp($),
+    version: API_VERSION
   };
 
   if (!result.isAmp) {
@@ -72,6 +74,7 @@ const execChecks = async (url) => {
     components: findAmpComponents($),
     data: lintResults,
     ...result,
+    version: API_VERSION
   };
 };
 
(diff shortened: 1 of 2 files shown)


pixi/backend/api.js
```diff
@@ -19,6 +19,7 @@ const {lint, LintMode} = require('@ampproject/toolbox-linter');
 const cheerio = require('cheerio');
 const log = require('@lib/utils/log')('Pixi API');
 const RateLimitedFetch = require('@lib/utils/rateLimitedFetch');
+const API_VERSION = 0;
```
tharders, line 22: Do we simply want to use a simple interger or more complex version numbers?
I would prefer to start with a 1.
The API here is not likely to change in general, only in the linter results.
Would you update the version number when the linter results change?
That would require an update of the number here, when the linter changes.
Maybe we can be more clever and use two version numbers (or a combined one) where we have the overall API version and the linter version for the data structure, so that there is no need for manual updates when the linter changes.

pixi/backend/api.js
```diff
@@ -61,6 +62,7 @@ const execChecks = async (url) => {
     redirected: res.redirected,
     url: res.url,
     isAmp: isAmp($),
+    version: API_VERSION
```
tharders, line 65: You do not need that here, since the version is already added after the `await execChecks` below.

pixi/backend/api.js
```diff
@@ -72,6 +74,7 @@ const execChecks = async (url) => {
     components: findAmpComponents($),
     data: lintResults,
     ...result,
+    version: API_VERSION
```
tharders, line 77: You do not need that here, since the version is already added after the `await execChecks` below.

pixi/backend/api.js
```diff
@@ -88,6 +91,7 @@ api.get('/lint', async (request, response) => {
     const result = {
       status: 'ok',
       ...checkResult,
+      version: API_VERSION
     };
     response.json(result);
   } catch (e) {
```
tharders, line 97: We might want to add the version for the error case as well

tharders (changes requested):
Please have a look at the linter results and my comments in the code.

Arwisss (approved):

PR #4430, opened 2020-09-02
Pixi: Accessibility improvements

caroqliu:
- Adds aria labels to the input bar and site-level survey banner.
- Replaces `h4` with `h3` to follow sequential hierarchy of heading importance.
- LH score prior to these changes was 84 and 97 after:
image

# Note on the two remaining failures:

## Color contrast
The first is regarding color contrast of the white text over the gradient background. Lighthouse cannot detect the color contrast of text over image, therefore it assumes the text (white) is over the page's off-white `background-color`, resulting in a very low contrast of 1.04. 

While the automated score fails this, the actual manual check reveals that the score is about 5.64 (more than sufficient) in the deepest blue and 2.33 in the lightest. For accessibility purposes I changed the gradient ratio such that the lightest portion underneath white text will have at least a 3:1 ratio region to meet color contrast best practices, however this will not correct the automated findings.
image
image

Note also that `text-shadow: 0 0 1px #000` did not appear to solve the issue.

## Sequential headings
The header for `Follow us` in the site-level footer is `h5`, thus the test auto detects that `h4` is skipped. It would fix the audit if we replaced the tag with `<h4>` for this specific page or `<header>` to generically fix all pages, but I'm not sure if this will have adverse affects on the rest of the site. Filed #4431 to follow up.
image

### DIFF
--- a/frontend/scss/components/molecules/pixi-status-intro.scss
+++ b/frontend/scss/components/molecules/pixi-status-intro.scss
@@ -27,7 +27,8 @@
         margin-left: 30px;
       }
 
-      h4, p {
+      h3,
+      p {
         color: color('white');
       }
     }
--- a/frontend/scss/components/templates/pixi.scss
+++ b/frontend/scss/components/templates/pixi.scss
@@ -111,7 +111,7 @@
         background: linear-gradient(
           225deg,
           color('robins-egg') 10%,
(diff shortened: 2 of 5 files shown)


matthiasrohmer (approved):
Nice, thanks!

Issue #4358, opened 2020-08-24
Pixi: Finalize progress bars for Core Web Vitals


matthiasrohmer:
via @caroqliu. https://github.com/ampproject/amp.dev/pull/4354#issuecomment-678298594.

@nainar, @sebastianbenz: what do you think about visualizing the proportions?

sebastianbenz:
> visualizing the proportions

Can you explain?

matthiasrohmer:
Sure! I think what @caroqliu meant is that Page Speed Insights shows fast, average, slow as proportions like so:

![Bildschirmfoto 2020-08-24 um 10 06 59](https://user-images.githubusercontent.com/12857772/91019620-90f3d780-e5f1-11ea-8110-3445168d1027.png)

I adapted this behaviour for the Pixi scales:

![Bildschirmfoto 2020-08-24 um 10 13 10](https://user-images.githubusercontent.com/12857772/91020199-74a46a80-e5f2-11ea-97de-929c0ad68fef.png)

 In Ferdi's layout each speed category took up equal parts though:

![Bildschirmfoto 2020-08-24 um 10 05 27](https://user-images.githubusercontent.com/12857772/91019765-cf899200-e5f1-11ea-8c5c-e2dd026483b0.png)

The question is in which direction we want to go. In my opinion the scales are really hard to understand. Adding tooltips like PSI does will already help:

![Bildschirmfoto 2020-08-24 um 10 14 48](https://user-images.githubusercontent.com/12857772/91020440-b8976f80-e5f2-11ea-9fde-6356a1cb139a.png)

But additionally I think we should stick to proportions as only then a good page has a big green chunk.

sebastianbenz:
I like the scales and percentages from PSI, having only a visual indicator as in our mocks is less clear. However, marking the actual value on the same scale is a bit confusing. 

One note: what I don't like about PSI is that it uses three colors (orange, green, yellow) to encode a binary result (fail / pass). What does orange mean?

matthiasrohmer:
>However, marking the actual value on the same scale is a bit confusing.

Right! That's what I am stumbling over.

Is the PSI result really binary? Actually it puts each vital in a category of "slow", "average" and "fast" - these categories are also what the three colors map to.

caroqliu:
One option could be to remove the separation markers and thresholds — i.e. 2.5s, 4s for LCP — that are plotted so there is not as much clutter. We still want to make this information available, possibly in the `?` icon next to the test title. 

[...]

PR #4434, opened 2020-09-03
✨ Pixi - Implement new primary checks layout

### LINKED ISSUE #4417: Pixi: graph values rendering incorrectly in edge cases
small screen values overlapping
image

long value being clipped 
image

robinvanopstal:
Fixes #4417, #4358


frontend/templates/views/partials/pixi/primary-metric.j2
```diff
@@ -1,36 +1,42 @@
...
-  <div class="ap-m-pixi-primary-metric-row">
-    <div>
-      <strong>{{ _('Opportunity to improve') }}:</strong>
-      <span class="ap-m-pixi-primary-metric-improvement">{{ _('Calculating') }}</span>
-      {% include 'views/partials/pixi/analyzing-dots.j2' %}
-    </div>
+  <div class="ap-m-pixi-primary-metric-body">
+    <div class="ap-m-pixi-primary-metric-body-result">
+      <div>
+        <strong>{{ _('Your average score') }}:</strong>
```
caroqliu, line 23:
```suggestion
        <strong>{{ _('Your score') }}:</strong>
```

frontend/icons/questionmark-outline.svg
```diff
@@ -0,0 +1 @@
...3.35 403.35S289.013 915.35 512 915.35c222.988 0 403.35-180.363 403.35-403.35S734.987 108.65 512 108.65zm-7.471 587.428c30.832 0 55.785 25.029 55.785 55.823s-24.953 55.747-55.747 55.747c-30.832 0-55.823-24.953-55.823-55.747 0-30.832 24.991-55.785 55.785-55.785zm8.988-456.973c47.025 0 85.062 12.59 113.77 37.923 28.898 25.56 43.422 59.54 43.422 101.293 0 22.754-4.627 43.915-13.842 63.483-9.101 19.417-22.943 37.506-41.412 54.382l-37.051 33.942-5.689 5.31-3.793 3.982-1.934 2.238-3.034 3.944a36.313 36.313 0 00-1.669 2.313l-3.793 5.689-7.395 12.022c-9.86 16.421-15.094 48.427-15.094 95.49v8.912l-26.319-.19-39.63.152-1.062-19.645c-2.541-55.292 1.024-92.722 11.035-112.86 9.48-21.161 21.047-37.051 41.071-56.126l27.494-26.546c13.501-12.704 23.057-25.105 28.708-36.975 5.499-11.566 8.267-24.726 8.267-39.554 0-20.365-6.637-36.254-20.1-48.465-13.69-12.325-31.78-18.545-54.723-18.545-28.632 0-49.527 8.115-63.559 24.271-11.946 13.652-19.531 33.752-22.45 60.374l-.417 4.399c-2.162 20.327-19.417 35.647-39.592 35.382-11.035.152-21.692-4.324-29.277-12.363s-11.415-18.924-10.618-28.329a24.804 24.804 0 01.038-4.437c5.385-44.977 19.872-85.403 43.688-108.422 29.922-28.784 71.713-43.043 124.957-43.043z"/></svg>
```
matthiasrohmer, line 1: The end tag needs to be `</symbol>`

frontend/scss/components/atoms/pixi-analyzing-dots.scss
```diff
@@ -10,13 +10,12 @@
 }
 
 .#{atom('pixi-analyzing-dots')} {
-  display: inline;
+  display: none;
   letter-spacing: -2px;
   transition: opacity 0.2s ease;
```
matthiasrohmer, line 15: Remove the transition if they are no longer fading in.

caroqliu (commented):
This is a great start. Two overall notes:
- The portion for the field data that is filled in should always be green, right?
Screen Shot 2020-09-03 at 9 23 15 AM
- The lab data should not have percentiles. We still want an indicator over the green/orange/red regions for this one since it is a single data point on that scale. How are you calculating % passed for lab data?
Screen Shot 2020-09-03 at 9 23 58 AM

matthiasrohmer (changes requested):

pixi/src/checks/PageExperienceCheck.js
```diff
@@ -89,7 +112,7 @@ export default class PageExperienceCheck {
             fid: {
               id: 'fid',
               unit: UNIT_MS,
-              data: this.createLabData(labData['interactive']),
+              data: this.createLabData(labData['total-blocking-time']),
```
caroqliu, line 115: Note also that the lab data copy should be identified as `Interactivity | TBT` not `FID`

pixi/src/ui/report/CoreWebVitalsReportView.js
```diff
@@ -116,13 +63,15 @@ class CoreWebVitalView {
...
 
-    this.scale.render(data, unit);
+    this.scale.render(data);
 
     const responseCategory = data.category.toLowerCase();
     const displayCategory = CATEGORIES[responseCategory];
     this.container.classList.add(responseCategory);
     this.category.textContent = displayCategory;
 
+    const average = (data.numericValue / unit.conversion).toFixed(2);
```
caroqliu, line 73: 1 decimal place should be enough :)

pixi/src/ui/report/CoreWebVitalsReportView.js
```diff
@@ -20,73 +20,21 @@ const CATEGORIES = {
...
-        }%`;
-        break;
-      }
+  render(data) {
+    const percentile = Math.round(data.score * 100);
 
-      distributionOffset += distribution.proportion * 100;
+    this.indicator.style.width = `${percentile}%`;
+    this.label.textContent = `${percentile}% passed`;
+    if (percentile < 20) {
```
tharders, line 35: For some screen sizes 20% is not enough even with the english "passed" text.
I think we should increase this a little bit to have some space left for other translations that are longer or implement a logic that is more dynamic (if you can think of any we should maybe leave it for now and do that in an other PR)

pixi/src/ui/report/CoreWebVitalsReportView.js
robinvanopstal: Okay! Even for CLS ? :)

pixi/src/checks/PageExperienceCheck.js
```diff
@@ -35,21 +40,33 @@ export default class PageExperienceCheck {
     }
   }
 
+  createFieldData(fieldData, id) {
+    const metric = fieldData[id];
+    const score = metric.percentile / METRICS_MAX[id];
```
tharders, line 45: Looks like this logic is not correct.
The better the page the smaller the value, but that does not match the logic here

tharders (commented):
Nice, just a few comments

matthiasrohmer:
@caroqliu, I think Robin missed the meeting where we decided about the scales. From what I understood I think the scales are almost like we wanted them to be:

For field data the copy that currently says "75th percentile" should actually contain the absolute numbers from where a page's performance is considered good. So for LCP this would be 75% (or TBD) of the scale's maximum which is defined here:

https://github.com/ampproject/amp.dev/blob/78defa9e2247b872b3f39916af0619b7ea3737d6/pixi/src/checks/PageExperienceCheck.js#L20

My understanding was that we drop the weighted scale with three areas (green, orange, red) as they require a lot more context/knowledge from the user. That means we would only need to decide what is the maximum that we want to show for each metric and from what point on a metric is good enough.

Then for lab data it's basically the same. The 75th percentile limit would instead also show an absolute value.

The discussion was moving very quickly, so it could be that I got things wrong if that's the case feel free to correct me 🙂

pixi/src/ui/report/CoreWebVitalsReportView.js
caroqliu: Good point, no, we do want 2 for CLS. However for FID we want an integer ms result. Let's handle this in another PR.

pixi/src/checks/PageExperienceCheck.js
```diff
@@ -35,21 +52,33 @@ export default class PageExperienceCheck {
     }
   }
 
-  createLabData(metric) {
-    const labData = {
+  createFieldData(metric, id) {
+    console.log(METRICS_SCALES[id].slow);
```
caroqliu, line 56: please remove

pixi/src/ui/report/CoreWebVitalsReportView.js
```diff
@@ -20,73 +20,52 @@ const CATEGORIES = {
...
-    let distributionOffset = 0;
-    for (const distribution of data.distributions) {
-      if (data.percentile < distribution.max) {
-        this.value.style.width = `${
-          (data.percentile / distribution.max) * 100
-        }%`;
-        break;
-      }
+  render(data) {
+    const percentile = Math.round(data.score * 100);
```
caroqliu, line 62: We still need to keep the distribution from the "good" proportion and set that to the green bar width. We should not be using the score to plot the visual, only providing it in the "Your score" text portion.

frontend/icons/questionmark-outline.svg
```diff
@@ -0,0 +1,2 @@
...363-403.35 403.35S289.013 915.35 512 915.35c222.988 0 403.35-180.363 403.35-403.35S734.987 108.65 512 108.65zm-7.471 587.428c30.832 0 55.785 25.029 55.785 55.823s-24.953 55.747-55.747 55.747c-30.832 0-55.823-24.953-55.823-55.747 0-30.832 24.991-55.785 55.785-55.785zm8.988-456.973c47.025 0 85.062 12.59 113.77 37.923 28.898 25.56 43.422 59.54 43.422 101.293 0 22.754-4.627 43.915-13.842 63.483-9.101 19.417-22.943 37.506-41.412 54.382l-37.051 33.942-5.689 5.31-3.793 3.982-1.934 2.238-3.034 3.944a36.313 36.313 0 00-1.669 2.313l-3.793 5.689-7.395 12.022c-9.86 16.421-15.094 48.427-15.094 95.49v8.912l-26.319-.19-39.63.152-1.062-19.645c-2.541-55.292 1.024-92.722 11.035-112.86 9.48-21.161 21.047-37.051 41.071-56.126l27.494-26.546c13.501-12.704 23.057-25.105 28.708-36.975 5.499-11.566 8.267-24.726 8.267-39.554 0-20.365-6.637-36.254-20.1-48.465-13.69-12.325-31.78-18.545-54.723-18.545-28.632 0-49.527 8.115-63.559 24.271-11.946 13.652-19.531 33.752-22.45 60.374l-.417 4.399c-2.162 20.327-19.417 35.647-39.592 35.382-11.035.152-21.692-4.324-29.277-12.363s-11.415-18.924-10.618-28.329a24.804 24.804 0 01.038-4.437c5.385-44.977 19.872-85.403 43.688-108.422 29.922-28.784 71.713-43.043 124.957-43.043z"/>
```
caroqliu, line 1: I think the ending `/` needs to be removed

caroqliu (commented):
Almost there, thanks!

matthiasrohmer (approved):

robinvanopstal:
> re:scales for lab data
> 
> green	yellow	red
> CLS	<0.1	>=0.1, < .25	>=.25
> LCP	<2.5s	>=2.5s, < 4s	>= 4s
> FID	<100ms	>=100ms, < 300ms	>= 300ms
> I assume they would be laid out with 33.3% distribution to each value, but I leave it to ferdi to make the call on the visual (which would impact the extreme negative end of the scale)

Hi Patrick, thanks for providing the data! In some cases, they don't seem to match with the API result though ..? :) 
Bildschirmfoto 2020-09-04 um 14 58 19

PR #4440, opened 2020-09-04
✨ Pixi - Opportunity to improve

robinvanopstal:


matthiasrohmer (approved):

matthiasrohmer:
@tharders, @caroqliu: this intentionally differs from the design doc. I consider extracting the logic to run the PSI test for the cache into a separate function premature optimization at the moment.

As long as we rely on calculating the difference of runs of the same API to get the opportunity to improve I think it's fine to keep the logic at a central place. Let's keep the idea of extracting it for when we find a better metric.

PR #4444, opened 2020-09-07
✨ Pixi - Add tooltips

robinvanopstal:


matthiasrohmer (changes requested):
Think about consistently using `|slug` to create id-safe strings instead of `|lower` and `|replace`.

I think you've extracted class toggling to an `amp-action-macro` that you didn't push maybe...? Because testing fails with

>[Action] Target "tooltip" not found for action [tap:tooltip, AMP.setState({pixi: { tooltip: pixiInfoTexts['lcp'] }})]

frontend/templates/views/partials/pixi/tooltip.j2
```diff
@@ -1,6 +1,25 @@
 {% do doc.styles.addCssFile('css/components/molecules/pixi-tooltip.css') %}
 
-<div class="ap-m-tooltip" role="button">
-  {% do doc.icons.useIcon('icons/questionmark-outline.svg') %}
-  <svg><use xmlns:xlink="http://www.w3.org/1999/xlink" xlink:href="#questionmark-outline"></use></svg>
+<div id="tooltip-{{ tooltip.id|replace(".", "-") }}"
+    class="ap-m-tooltip">
+
+  <div class="ap-m-tooltip-toggle"
```
matthiasrohmer, line 6: Please make this a real `<button>` instead of setting `role`

frontend/templates/views/partials/pixi/tooltip.j2
```diff
@@ -1,6 +1,25 @@
...
+      on="tap:tooltip-{{ tooltip.id|replace(".", "-") }}.toggleClass(class=active)"
+      role="button">
+    {% do doc.icons.useIcon('icons/questionmark-outline.svg') %}
+    <svg><use xmlns:xlink="http://www.w3.org/1999/xlink" xlink:href="#questionmark-outline"></use></svg>
+  </div>
+
+  <div class="ap-m-tooltip-tip">
+    <div class="ap-m-tooltip-close"
```
matthiasrohmer, line 14: Also make this a `button`.

matthiasrohmer (approved):
This way it's a lot more concise, thanks!

PR #4448, opened 2020-09-08
pixi loading status banner

tharders:
Add pixi loading status banner with check count
Also improve reset of views for next check 
and avoid error when page has no field data (specific view for that case is open)


robinvanopstal (approved):

frontend/templates/views/partials/pixi/status-intro.j2
```diff
@@ -15,11 +15,11 @@
       <p>That recommendations will come up Lorem ipsum dolor sit amet, consectetur adipis cing elit, sed do eiusmod tempor incididunt.</p>
     </div>
   </div>
-  <div id="status-intro-banner" class="ap-m-pixi-status-intro-banner">
+  <div id="status-intro-banner-loading" class="ap-m-pixi-status-intro-banner">
     <div class="ap-m-pixi-status-intro-banner-icon"></div>
     <div>
-      <h3>&nbsp;</h3>
-      <p>&nbsp;</p>
+      <h3>Please wait a moment<br>We are currently analyzing your page{% include 'views/partials/pixi/analyzing-dots.j2' %}</h3>
```
matthiasrohmer, line 21: Enclose translated texts in `{{ _('...') }}`

pixi/src/checks/MobileFriendlinessCheck.js
```diff
@@ -17,6 +17,10 @@ const API_ENDPOINT = API_ENDPOINT_MOBILE_FRIENDLINESS;
 const API_URL = `${API_ENDPOINT}?key=${AMP_DEV_PIXI_APIS_KEY}`;
 
 export default class MobileFriendlinessCheck {
+  static getCheckCount() {
+    return 2;
```
matthiasrohmer, line 21: Doesn't this only render one card?

matthiasrohmer (commented):
Look good so far. Wondering if the check count should refer to either the API requests made or the cards that the user is able to see at the end...

frontend/templates/views/partials/pixi/status-intro.j2
tharders: We should take all translated texts out of the pixi/index.md.
That is something we need to change for all components later.

pixi/src/checks/MobileFriendlinessCheck.js
tharders: We have the mobileFriendly and resourcesLoadable checks.

PR #4455, opened 2020-09-08
Refactor page experience check to run cache test separately

tharders:
Cache check is not run if page is invalid amp.

This is also the basis for the last missing conditions in the statusBanner logic.
Localization of the texts will be done in a separate PR.


pixi/src/checks/PageExperienceCacheCheck.js
```diff
@@ -0,0 +1,61 @@
...
+        this.addImprovement(cacheLabData, originLabData, 'fid');
+      }
+    }
+    return cacheResult;
+  }
+
+  addImprovement(cacheMetric, originMetric, key) {
+    const cacheData = cacheMetric[key].data;
+    const originData = originMetric[key].data;
+    cacheData.improvement = originData.numericValue - cacheData.numericValue;
```
caroqliu, line 59:
```suggestion
    cacheData.improvement = Math.max(0, originData.numericValue - cacheData.numericValue);
```

pixi/src/checks/PageExperienceCacheCheck.js
```diff
@@ -0,0 +1,61 @@
...
+  async run(originUrl, originPageExperiencePromise) {
+    const cacheUrl = await ampToolboxCacheUrl.createCacheUrl(
+      AMP_PROJECT_CDN_URL,
+      originUrl
+    );
+
+    const cacheResult = await this.pageExperienceCheck.run(cacheUrl);
+    const originResult = await originPageExperiencePromise;
+    if (cacheResult.data && originResult.data) {
+      const cacheFieldData = cacheResult.data.pageExperience.fieldData;
```
caroqliu, line 38: Do we know that if `cacheResult.data` is defined, `cacheResult.data.pageExperience` will also be defined?

caroqliu (approved):

pixi/src/checks/PageExperienceCacheCheck.js
tharders: Yes, the PageExperienceCheck returns either `error`, or `data` with `pageExperience`

pixi/src/checks/PageExperienceCacheCheck.js
tharders: Done

pixi/src/checks/PageExperienceCacheCheck.js
```diff
@@ -0,0 +1,61 @@
...
+      const originLabData = originResult.data.pageExperience.labData;
+      if (cacheFieldData && originFieldData) {
+        this.addImprovement(cacheFieldData, originFieldData, 'lcp');
+        this.addImprovement(cacheFieldData, originFieldData, 'cls');
+        this.addImprovement(cacheFieldData, originFieldData, 'fid');
+      }
+      if (cacheLabData && originLabData) {
+        this.addImprovement(cacheLabData, originLabData, 'lcp');
+        this.addImprovement(cacheLabData, originLabData, 'cls');
+        this.addImprovement(cacheLabData, originLabData, 'fid');
```
caroqliu, line 50: Is this read at all from the response? Missed this, but the metric for lab data is "tbt" in lieu of "fid".

tharders: Currently the check still returns the "tbt" value as "fid", so this should work.

PR #4456, opened 2020-09-08
invalidate cached static assets on deploy

patrickkettner:
a recent PR that was merged updated an image, but the change was not immediately visible, due to caching. this PR is trying top prevent that from happening

### DIFF
--- a/gulpfile.js/deploy.js
+++ b/gulpfile.js/deploy.js
@@ -21,6 +21,7 @@ const {join} = require('path');
 const {sh} = require('@lib/utils/sh.js');
 const mri = require('mri');
 const {ROOT, THUMBOR_ROOT} = require('@lib/utils/project').paths;
+const {execSync} = require('child_process');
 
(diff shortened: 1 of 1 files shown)


gulpfile.js/deploy.js
```diff
@@ -21,6 +21,7 @@ const {join} = require('path');
 const {sh} = require('@lib/utils/sh.js');
 const mri = require('mri');
 const {ROOT, THUMBOR_ROOT} = require('@lib/utils/project').paths;
+const {execSync} = require('child_process');
```
matthiasrohmer, line 24: Try to consistently use the `sh` util.

matthiasrohmer (approved):
Thanks for giving this a try, only one nit. For more context also see my Slack response: images get cached inside Thumbor and are cache-busted by query param.

patrickkettner:
closing, as it ended up being a wrong path

PR #4464, opened 2020-09-09
Pixi: Give tooltip elements unique ids

caroqliu:
This fixes a bug where the lab data tooltips for `LCP` and `CLS` are toggling the tooltip elements associated with their field data counterparts.


matthiasrohmer (approved):
`tooltip.id|replace(".", "-")  + tooltip.type|default('')`, with all what we did with jinja2 on amp.dev this library definitely deserves the badge "battle-tested"...

Issue #4438, opened 2020-09-04
Pixi: do not allow cache URLs for testing


matthiasrohmer:
When validating the input URL it should throw a validation error for cache URLs as we are checking the cache via "Opportunity to improve"

caroqliu:
The cache URLs should have their own invalid status banner, I'll add the `md` for this shortly.

PR #4479, opened 2020-09-10
Pixi: Add amp-cache-url.md

caroqliu:
This is the status banner for URLs that are cached AMP pages. Partial for #4438


tharders (approved):
Cool, I will implement the check for that

PR #4485, opened 2020-09-10
fix banner offset

### LINKED ISSUE #4483: Menu button misaligned on mobile devices
The menu is about 30px below the rest of the header on mobile devices. I see this on my Pixel 2, and I see it in Chrome's emulator:

amp dev_(iPhone 6_7_8)

I don't know this CSS that well, but I do wonder why the `31px` is in this rule:

```
.has-banner+.ap-o-burger-menu-label {
    top: calc(49px + 31px);
}
```

patrickkettner:
fix #4483

### DIFF
--- a/frontend/templates/views/partials/header.j2
+++ b/frontend/templates/views/partials/header.j2
@@ -73,7 +73,7 @@ to provide its state to other components #}
 </nav>
 {% endmacro %}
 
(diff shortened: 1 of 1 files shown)


morsssss (approved):
Tried it... lgtm!

I've hit issues with the header before too. Maybe at some point we should rework the CSS....

PR #4491, opened 2020-09-10
Pixi: Display invalid AMP status after other tests have run.

caroqliu:
It seems kind of confusing to show the invalid AMP status while tests are still running, as well as with recommendations still on the way. This way, even though we know a page is invalid, the user can still keep up with progress.


tharders (approved):
Goof point, thanks

PR #4495, opened 2020-09-11
Implement isCacheUrl check and rule for status banner

### LINKED ISSUE #4438: Pixi: do not allow cache URLs for testing

tharders:
The check is implemented in the linter backend api as isCacheUrl.
The frontend linter check component translates this into isNotCacheURL to have all successful checks "true".
The difference is mostly due to the fact that we use the prod api on stage and with this setup pixi will not show the "cache url" status when the new flag is not in the linter api response.

So having said that, **please remember the linter will only show the "amp-cache-url" status in production mode (local or on stage) when this is deployed on prod.**

Fixes #4438

### DIFF
--- a/package.json
+++ b/package.json
@@ -60,6 +60,7 @@
     "not dead"
   ],
   "dependencies": {
+    "@ampproject/toolbox-cache-list": "2.6.0",
     "@ampproject/toolbox-cors": "2.6.0",
     "@ampproject/toolbox-linter": "2.7.0-alpha.3",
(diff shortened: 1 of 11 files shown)


caroqliu (approved):
Thanks for working on this! :) Can we rename `isNotCacheUrl` as `isOriginUrl`?

We are also safe to deploy this on prod since amp.dev/page-experience only shows the "Coming soon" ui in prod.

PR #4496, opened 2020-09-11
Pixi: Implement "Take action" texts

caroqliu:
Also cleans up a few related i18n bits. :)

### DIFF
--- a/pixi/src/ui/recommendations/RecommendationsView.js
+++ b/pixi/src/ui/recommendations/RecommendationsView.js
@@ -44,10 +44,10 @@ export default class RecommendationsView {
     }
   }
 
-  render(recommendationIds) {
+  render(recommendationIds, metricUis) {
     this.container.classList.remove('pristine');
     const recommendations = i18n.getSortedRecommendations(recommendationIds);
-    const tagIds = [];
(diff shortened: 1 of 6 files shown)


pixi/src/ui/recommendations/RecommendationsView.js
```diff
@@ -83,14 +83,15 @@ export default class RecommendationsView {
...
+        const count = tagIdCounts[tagId] || 0;
+        tagIdCounts[tagId] = count + 1;
       }
 
       this.recommendationNodes.push(recommendation);
       this.container.appendChild(recommendation);
     }
 
-    for (const tagId of new Set(tagIds).values()) {
+    for (const tagId of Object.keys(tagIdCounts)) {
```
patrickkettner, line 94: why this change?

patrickkettner (approved):
LGTM!

pixi/src/ui/recommendations/RecommendationsView.js
caroqliu: The Set was used to get unique values from an array with duplicates. I replaced the array with duplicates with an object that keeps count of values instead.

PR #4497, opened 2020-09-11
Pixi: Fully reset loading UI between runs

### LINKED ISSUE #4478: Pixi: status colors should be reset between runs
if you scan a failing url, then scan a good one, the visual results are confusing

image

caroqliu:
- Resets colors of performance summaries
- Fixes bug in filter UI resetting

Fixes #4478

### DIFF
--- a/pixi/src/ui/recommendations/RecommendationsView.js
+++ b/pixi/src/ui/recommendations/RecommendationsView.js
@@ -39,8 +39,9 @@ export default class RecommendationsView {
       this.container.removeChild(items[i]);
     }
 
-    for (let i = 1; i < this.filter.children.length; i++) {
-      this.filter.removeChild(this.filterPills[i]);
+    const length = this.filter.children.length;
+    for (let i = 1; i < length; i++) {
+      this.filter.removeChild(this.filter.lastChild);
     }
   }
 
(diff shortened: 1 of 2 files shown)


pixi/src/ui/recommendations/RecommendationsView.js
```diff
@@ -39,8 +39,9 @@ export default class RecommendationsView {
       this.container.removeChild(items[i]);
     }
 
-    for (let i = 1; i < this.filter.children.length; i++) {
-      this.filter.removeChild(this.filterPills[i]);
+    const length = this.filter.children.length;
```
patrickkettner, line 42: Any chance filter would be undefined?

pixi/src/ui/report/CoreWebVitalsReportView.js
```diff
@@ -134,7 +134,7 @@ class CoreWebVitalView {
   }
 
   reset() {
-    this.container.classList.remove(...Object.values(CATEGORIES));
+    this.container.classList.remove(...Object.keys(CATEGORIES));
```
patrickkettner, line 137: At one point removing multiple classes at once didn't work and worker-dom, I'm assuming this was fixed?

pixi/src/ui/recommendations/RecommendationsView.js
caroqliu: Not unless it gets removed from the j2 fragment, but as it stands it's always defined on construction of this class.

patrickkettner (approved):

PR #4503, opened 2020-09-11
Fix CSS error

morsssss:
Fixes CSS error in amp-script samples.

This actually doesn't affect the main samples page. But if you pull an individual sample into the playground, the error isn't corrected.


patrickkettner (approved):

patrickkettner:
great catch!

morsssss:
Thanks for focusing on the positive part - not the "dumb mistake" bit 😜

PR #4512, opened 2020-09-14
Pixi - Implement new indicator styling for lab data 

robinvanopstal:


frontend/scss/components/atoms/pixi-scale.scss
```diff
@@ -122,6 +122,65 @@
...
+      transform: translateX(7px);
+
+      &:after {
+        content: '';
+        width: 0;
+        height: 0;
+        margin: 0 0 0 6px;
+        border: 7.5px solid transparent;
+        border-bottom: 0;
+        border-top-color: color('malachite');
```
matthiasrohmer, line 147: Couldn't this use `currentColor` instead? This would also save you the additional `&:after` definitions in 153ff and 160ff

matthiasrohmer (changes requested):

matthiasrohmer (approved):

PR #4519, opened 2020-09-14
Pixi: Update message when no field data is available

caroqliu:
Adds link to the Google Search Console Performance report FAQ.

### DIFF
--- a/frontend/templates/views/partials/pixi/primary-checks.j2
+++ b/frontend/templates/views/partials/pixi/primary-checks.j2
@@ -45,7 +45,12 @@
     {% endwith %}
 
     <div class="ap-o-pixi-primary-checks-data-error-label">
-      <label>{{ _('Unfortunately, we do not have sufficient real-world performance data for this page. Please have a look at the lab data or come back later in a view days.') }}</label>
+      <label>
+        {{ _('Unfortunately, we do not have sufficient real-world performance data for this page. Please have a look at the lab data or use the ') }} 
+        <a href="https://support.google.com/webmasters/answer/7576553?hl=en&ref_topic=9384513">
+          {{ _('Search Console Performance report.') }}
+        </a>
+      </label>
     </div>
 
   </div>


frontend/templates/views/partials/pixi/primary-checks.j2
```diff
@@ -45,7 +45,12 @@
     {% endwith %}
 
     <div class="ap-o-pixi-primary-checks-data-error-label">
-      <label>{{ _('Unfortunately, we do not have sufficient real-world performance data for this page. Please have a look at the lab data or come back later in a view days.') }}</label>
+      <label>
+        {{ _('Unfortunately, we do not have sufficient real-world performance data for this page. Please have a look at the lab data or use the ') }} 
+        <a href="https://support.google.com/webmasters/answer/7576553?hl=en&ref_topic=9384513">
+          {{ _('Search Console Performance report.') }}
```
patrickkettner, line 51: ￼could this be slightly more descriptive? I don't think the majority of visitors would immediately understand what this is

nainar: Suggestions:

`Unfortunately, we do not have sufficient real-world performance data for this page. Please have a look at the lab data or use the Search Console Performance report. Our data is based on the Chrome User Experience Report which aggregates real-world speed data from opted-in users and requires that a URL must be public (crawlable and indexable) and have sufficient number of distinct samples that provide a representative, anonymized view of performance of the URL.`

`Unfortunately, we do not have sufficient real-world performance data for this page. Please have a look at the lab data or use the Search Console Performance report. Our data is based on the Chrome User Experience Report which does not have sufficient real-world speed data for this page. Chrome User Experience Report aggregates real-world speed data from opted-in users and requires that a URL must be public (crawlable and indexable) and have sufficient number of distinct samples that provide a representative, anonymized view of performance of the URL.`

tharders: @caroqliu I have extracted all static texts into the markdown files in an other PR. I have used the first suggestion from @nainar in the new file `pages/content/pixi/no-field-data.md`. If that is fine we can close this ticket.

caroqliu: Sounds great to me, thank you!

PR #4520, opened 2020-09-14
Pixi - Fix lab data scale

robinvanopstal:


tharders:
The bars for field tests and lab tests are now positioned at different heights.
Please check the layout and move both at the same height next to "Opportunity to improve".
But please also check the font size of  "Opportunity to improve" (and around). Should be 14 instead of 16px - the same as in the graph.

matthiasrohmer (approved):

PR #4521, opened 2020-09-14
Pixi: Display messages for 0 recommendations

### LINKED ISSUE #4505: [Accessibility] Multiple issues with "Take action: N recommendations" buttons
Environment:

Steps to reproduce:
2. Enter a valid URL and press Analyze
3. Wait for the analysis to finish
4. Press tab repeatedly to move keyboard focus through controls
5. Attempt to move keyboard focus to the "Take action: N recommendations" button

Expected:
It should be possible to focus these buttons with the keyboard and activate them by pressing enter.

Actual:
These buttons are only useable with the mouse.

6. Turn on a screen reader
7. Navigate to one of the "Take action: N recommendations" buttons

Expected:
Screen readers should read the exact same text that is shown visually.

Actual:
Screen readers read these buttons as "Filter bylcp" or "Filter byfid button", for example.

8. Use a screen reader to activate the button

Expected:
Something happens that is understandable and meaningful to a screen reader user.

Actual:
The page scrolls to the "Take action. Improve your AMP site" section
Screen reader focus is not affected, so a screen reader user will think nothing happened or that something isn't working.

Once keyboard focusability is implemented on these buttons, keyboard focus should also do something meaningful when the button is pressed, rather than remaining on the button when the page scrolls.

The way I typically see this sort of "jump to heading" functionality implemented is by using a link with href="/#heading-name". In any case, this button should do something useful to and understandable by keyboard and screen reader users.

image

caroqliu:
The "Take action" text updates with the following when there are 0 recommendations surfaced for a given category:
- Good status => "Nothing to do"
- Else => "File an issue with AMP" + link to GH issue template

This PR also:
- Removes AMP action `scrollTo` in favor of header hyperlink. This fixes #4505

### DIFF
--- a/frontend/scss/components/molecules/pixi-primary-metric.scss
+++ b/frontend/scss/components/molecules/pixi-primary-metric.scss
@@ -99,11 +99,10 @@
         margin-right: 20px;
       }
 
-      span,
+      a,
       svg {
         white-space: nowrap;
-        opacity: 0.3;
-        transition: opacity .3s ease;
+        transition: opacity 0.3s ease;
 
         .fast &,
         .average &,
--- a/frontend/templates/views/partials/pixi/primary-metric.j2
+++ b/frontend/templates/views/partials/pixi/primary-metric.j2
@@ -37,12 +37,12 @@
         {% include 'views/partials/pixi/analyzing-dots.j2' %}
       </div>
 
-      <div on="tap:recommendations.scrollTo(duration=400),
-              recommendations.toggleClass(class={{ primary_metric.title.id|slug }})"
-            aria-label="{{ _('Filter by') + primary_metric.title.id|slug }}"
-            role="button">
+      <div>
         <strong>{{ _('Take action') }}:</strong>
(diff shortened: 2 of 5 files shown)


frontend/scss/components/molecules/pixi-primary-metric.scss
```diff
@@ -99,11 +99,10 @@
         margin-right: 20px;
       }
 
-      span,
+      a,
```
patrickkettner, line 102: ￼this is a super generic selector, are you sure there are no side effects? Would it be possible to add a class rather than using a tag?

caroqliu: Yeah I can do that.

caroqliu: Actually, this one is only on `a` descendants of the `ap-m-pixi-primary-metric-body-result` class, so should be safe.

patrickkettner: ah, I missed this was a descendant selector-thanks!

patrickkettner (approved):

Issue #4486, opened 2020-09-10
[Accessibility] Not possible to activate Analyze button with keyboard or screen reader


PeterSaathoff-Harshfield:
Environment:
Any desktop OS/browser (Mac OS, Windows, Linux, Chrome OS)

Steps to reproduce:
1. Open https://amp-dev-staging.appspot.com/page-experience/
2. Press tab to move keyboard focus to the "Enter a valid URL" field
3. Paste in a valid URL
4. Press tab to move keyboard focus to the Analyze button
5. Press enter

Expected result:
The button is activated

Actual result:
The Analyze button is disabled unless the mouse clicks in the field, so pressing tab skips the button

It must be possible to enter the field and activate the button with the keyboard only or with a screen reader.

caroqliu:
I was able to reproduce this, it looks like the logic conducted by the amp-script src is disabled due to "lack of user action" -- is Tab `focus` able to be construed as user action? @kristoferbaxter @samouri 

![image](https://user-images.githubusercontent.com/10456171/93092075-f83efd80-f66c-11ea-930a-d72bcc921c8d.png)

samouri:
@caroqliu, I don't see why not. We just need to add "focus" to this list:

https://github.com/ampproject/amphtml/blob/7910e6c254b61301a81badbcf3f1e1de0925066a/extensions/amp-script/0.1/user-activation-tracker.js#L19-L26

samouri:
Hmm. Something is fishy here. Is the goal to enable the analyze button on input focus?
Why not wait for validation of the input text value before enabling the Analyze button?

caroqliu:
The input checks if its value is a valid URL before enabling the "Analyze" button here: 
https://github.com/ampproject/amp.dev/blob/6b3a080f199f8ee0236242524046179cc649cf6f/pixi/src/ui/InputBar.js#L26-L32

It looks like the logic that validates the input as a requirement to enabling the button is what is not able to run here.

samouri:
Right. And there's no reason to check for a valid url on the initial TAB `keyup` event right? Its probably safe to completely ignore TAB within this event handler

caroqliu:
Let me clarify, this is currently the keyboard-only user flow:
1. `Tab` until input field is focused. amp-script fires error: `Blocked 1 attempts to modify DOM element attributes or styles. For variable-sized <amp-script> containers, a user action has to happen first.` Followed by message that amp-script was terminated `due to illegal mutation.`
[...]

PR #4526, opened 2020-09-14
analyze button: dont attempt to validate on focus

samouri:
Fixes https://github.com/ampproject/amp.dev/issues/4486

@caroqliu

### DIFF
--- a/pixi/src/ui/InputBar.js
+++ b/pixi/src/ui/InputBar.js
@@ -29,7 +29,7 @@ export default class InputBar {
       }
 
       // Ignore tab key which only happens on focus because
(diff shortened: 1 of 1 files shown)


caroqliu (commented):
Thanks for this fix, Jake! If possible let's actually switch the event we're listening to from `keyup` to `input` so we don't have to special case `Tab`

samouri:
> Thanks for this fix, Jake! If possible let's actually switch the event we're listening to from keyup to input so we don't have to special case Tab

Done!

caroqliu (approved):

PR #4532, opened 2020-09-15
Put all pixi static texts into markdown files

tharders:
Also fix layout of intro text

### DIFF
--- a/frontend/templates/views/partials/pixi/status-intro.j2
+++ b/frontend/templates/views/partials/pixi/status-intro.j2
@@ -2,44 +2,34 @@
 
 <section id="status-intro" class="ap-m-pixi-status-intro">
   <div class="ap-m-pixi-status-intro-text">
-    <div class="ap-m-pixi-status-intro-text-item">
-    <p>
-      The AMP page experience guide is a tool that shows AMP developers how 
-      their AMP pages are performing against the Google Search page experience 
-      ranking signal, and provides actionable feedback on how they can improve.
-    </p>
-
-    <p>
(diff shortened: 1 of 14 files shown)


robinvanopstal (approved):

frontend/templates/views/partials/pixi/status-intro.j2
```diff
@@ -2,44 +2,34 @@
...
+      <h3>
+        {{ pixi.staticText.statusIntro.headline }}
+        {% if pixi.staticText.statusIntro.headline2 %}
+          <br>
+          {{ pixi.staticText.statusIntro.headline2 }}
+        {% endif %}
+        {% include 'views/partials/pixi/analyzing-dots.j2' %}
+      </h3>
+      <p>
+        {{ pixi.staticText.statusIntro.loadingCopy|replace('${finishedChecks}', '<span id="finished-checks">0</span>')|replace('${totalChecks}','<span id="total-checks">0</span>')|safe }}
```
matthiasrohmer, line 19: Can we instead use proper [`format`](https://jinja.palletsprojects.com/en/2.11.x/templates/#format) syntax here?

matthiasrohmer (approved):
Only one small request.

frontend/templates/views/partials/pixi/status-intro.j2
tharders: The format placeholders cause problems in the markdown frontmatter.
Since we currently do not use `format` elsewhere there is no need to align things.
So we keep it like it is for now.

PR #4539, opened 2020-09-15
Add amp-optimizer redirect as well.

nainar:


matthiasrohmer (commented):
Has this URL already been communicated? Otherwise I'd vote to move this to the go-links.yaml instead and then have it usable via go.amp.dev/amp-optimizer - originally amp-dev-redirects.yaml was only intended for moved pages and URLs that users might expect to exist as they have been existing for ampproject.org for example.

Just wanted to give some context, nothing critical... 🙂

nainar:
Lol I thought I replied here: So I accidentally used this short link 🤦‍♀️  in the AMP fest talk recording. We can go back and change the URL on screen to be the go link if needed?

matthiasrohmer (approved):
Oh no that's fine, don't want to be too strict about it 🙂

PR #4549, opened 2020-09-16
Feature/pixi extract dynamic texts

tharders:
The texts in the index.md are now separated into staticText for the templates and scriptText for the scripts via amp-state.
There are two texts that are used in script and template. I opted in for simply using the scriptText in the template, since the main point in the separation is to only put the text in the amp-state that are needed by the script.

I also translated the intro text in the german index.md to avoid having mixed languages in one file.
(Unfortunately it is not possible to keep the body of a translated markdown file empty to use the body of the main language.)


matthiasrohmer (approved):
Looks good, needs to be rebased though.

PR #4553, opened 2020-09-16
🐛 Do not use origin fallback for field data in Pixi

### LINKED ISSUE #4544: Pixi: Reporting origin summary accidentallly
For the following page: https://www.axios.com/joe-biden-kamala-harris-vice-president-e2109284-944c-4053-b506-9735e261e3ff.html

PSI doesn't have field data: 
Screen Shot 2020-09-15 at 12 44 10 PM

However, Pixi is taking the origin summary and applying it here: 
Screen Shot 2020-09-15 at 12 44 36 PM

It should show: no field data available

matthiasrohmer:
Fixes #4544.


pixi/mocks/pageExperienceCheck/reportDataErrors.json
```diff
@@ -1,44 +1,6 @@
...
-				},
-				"data": {
-					"numericValue": 0,
-					"category": "FAST",
-					"proportion": 0.932653831634579
-				}
-			},
-			"isAllFast": false
-		},
+		"fieldData": null,
```
tharders, line 3: We should introduce a new test series here to check the case when origin_fallback is set and change the existing error case to not have origin_fallback set.
Otherwise we never check the "not fast" cases in the field data together with "isAllFast: false"

pixi/src/checks/PageExperienceCheck.js
```diff
@@ -153,7 +155,7 @@ export default class PageExperienceCheck {
                   this.isFastData(fieldData, 'lcp'),
                 ...fieldData,
               }
-            : undefined,
+            : null,
```
tharders, line 158: Any reason for using null here?
I think there are other places where we should change it if we prefer null over undefined.

tharders (changes requested):
Please ensure we do not reduce the test coverage (see comments)

pixi/src/checks/PageExperienceCheck.js
matthiasrohmer: `null` is valid JSON while `undefined` is not, that was my only reasoning. In my opinion this also makes debugging easier: having a value of `null` means that a variable has been evaluated while `undefined` could also mean it has not been evaluated at all.

pixi/mocks/pageExperienceCheck/reportDataErrors.json
matthiasrohmer: Good point, reset the changes to the error test case.

tharders (approved):
As far as I see the "origin_fallback" logic is now not fully covered by the unit tests.
Would not hurt to add an extra test for "origin_fallback: true".

But otherwise Ok.

matthiasrohmer:
Sorry, was confused about the test cases. Added the missing one.

PR #4555, opened 2020-09-16
✨ Cache mobile friendliness and safe browsing check for Pixi

### LINKED ISSUE #4529: Pixi: We should cache successful results
We should cache results if all the tests finished. This allows users to rerun if a test failed.

matthiasrohmer:
Resolves #4529. This only caches Safe Browsing and Mobile Friendliness checks as they are the ones that take the longest. PSI is not cached to prevent confusing users who develop against Pixi results and the AMP linter API is our USP and runs fast enough to run on every request in my opinion.

We could think about switching from `sessionStorage` to `localStorage` we would only need to decide on a cache timeout then, @nainar and @caroqliu.


caroqliu (approved):
Thanks, Matthias!

IMO we can cache PSI field data (if present) for up to a day since it is a lagging metric. I think it would be reasonable not to store lab data if i.e. someone wanted to run many times. I'll let @nainar and @patrickkettner weigh in here too.

matthiasrohmer:
Thought about this too - it doesn't really matter though as both are coming from the same request. We would only benefit from caching field data (and fetching lab data in the background) if we would refactor the CoreWebVitalsView to be able to render only partially.

caroqliu:
Yeah, I was just about to comment that same thought -- we don't get anything by caching only part of the response. Are we mainly optimizing for speed of results or also minimizing for # of queries? In any case this PR is good to merge and if we want to continue with PSI we can do that separately.

matthiasrohmer:
Sounds good to me 🙂

PR #4546, opened 2020-09-15
Pixi: Surface specific statuses for different API failures

caroqliu:
Fixes #4473

Status for when an API fails adds a new `Investigate` button, which links either to the `Core Web Vitals` header or the `Additional Checks` header: 
Desktop
image
Mobile
image

Example with Safe Browsing analysis failed:
image

PR #4556, opened 2020-09-16
Pixi - Fix metric loading indicator

### LINKED ISSUE #4551: pixi loading indicator hidden to early for "Take action" status
When the core web vitals data is filled all the loading indicators inside are hidden.
But the "Take action" is still in status "Analyzing" (now without the needed loading indicator dots) because the recommendations are not available yet.

robinvanopstal:
Fixes #4551.


matthiasrohmer:
Hm... let's think about a more generic solution to communicate state between the various components.

matthiasrohmer:
When we work on this tomorrow we should also consider https://github.com/ampproject/amp.dev/pull/4546/files#r488965528

matthiasrohmer (approved):

PR #4559, opened 2020-09-16
Pixi: Hide invisible text from screen readers

### LINKED ISSUE #4490: [Accessibility] Screen readers are able to navigate to "Please wait a moment..." text before it is visible
Environment:

Steps to reproduce:
2. Turn on a screen reader
3. Linear navigate through the page, past the URL field and "Why is it important?" and related sections
4. Notice that the following text is accessible with the screen reader:
>Please wait a moment
>We are currently analyzing your page . . .
>0 out of 0 checks are completed
This text is supposed to appear once the Analyze button has been pressed. It should not be possible to navigate to this text with a screen reader until it is intended to be read.

### LINKED ISSUE #4492: [Accessibility] Screen readers are able to navigate to "Why is it important..." text before it is visible
Environment:

Steps to reproduce:
2. Turn on a screen reader
3. Linear navigate with the screen reader to the URL field, enter a URL, and press Analyze
4. Continue linear navigating down the page
5. Notice that the following text is still accessible to the screen reader:
>Why is it important
>Web Vitals is an initiative by Google to provide unified guidance for quality signals that are essential to delivering a great user experience on the web..

This text is supposed to disappear once the Analyze button has been pressed. It should not be possible to navigate to this text with a screen reader at this point.

caroqliu:
Fixes #4492, fixes #4490

### DIFF
--- a/frontend/scss/components/molecules/pixi-status-intro.scss
+++ b/frontend/scss/components/molecules/pixi-status-intro.scss
@@ -50,7 +50,6 @@
     position: absolute;
     top: 0;
     left: 15px;
-    display: flex;
     width: calc(100% - 30px);
     min-height: 100%;
     padding: 30px 15px;
@@ -59,7 +58,7 @@
     background-position: 60px 50px;
     background-color: color('white');
     pointer-events: none;
-    opacity: 0;
+    display: none;
     transition: opacity 0.2s ease-in-out, transform 0.3s ease-in-out;
 
     @media (min-width: 768px) {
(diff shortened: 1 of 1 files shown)


patrickkettner (approved):

frontend/scss/components/molecules/pixi-status-intro.scss
```diff
@@ -59,7 +58,7 @@
     background-position: 60px 50px;
     background-color: color('white');
     pointer-events: none;
-    opacity: 0;
+    display: none;
     transition: opacity 0.2s ease-in-out, transform 0.3s ease-in-out;
```
matthiasrohmer, line 62: All the transitions have no effect anymore with this change, we should remove the related properties as well or think about toggling `aria-hidden` instead.

caroqliu: Ah good catch. We can toggle `aria-hidden` instead if preferred.

matthiasrohmer: Depends on wether we want to keep fancyness 😁

PR #4560, opened 2020-09-16
🐛 Various fixes for Pixi

matthiasrohmer:
- This fixes the exception that occurred by recommendations without body 
- Removes the button disabled state and validation on key-up
- Only caches successful API responses for Safe Browsing and Mobile Friendliness


caroqliu:
Can you clarify what is the error for recommendations without body? I thought some intentionally do not have body so they can fallback to PSI response texts.

matthiasrohmer:
@tharders fixed the initial problem with #4578. I'd still leave the safety net of gracefully handling non-existing body values in. Without `marked` it would not throw a NPE anymore but it would still render a expandable recommendation that doesn't show anything when expanded.

caroqliu (approved):

PR #4562, opened 2020-09-16
Pixi: Improve accessibility for status buttons

### LINKED ISSUE #4493: [Accessibility] "Fix it now" button has unclear behavior for keyboard/screen reader users
Environment:

Steps to reproduce:
2. Enter a valid URL and press Analyze
3. Wait for the analysis to finish
4. Press tab to move keyboard focus to the "Fix it now" button, or navigate there with a screen reader
5. Press enter to activate the "Fix it now" button

What happens:
The page scrolls to the "Take action. Improve your AMP site" section

Keyboard focus has not moved from the "Fix it now" button, so if a keyboard-only user presses tab, the page will scroll back up to where their focus is.

Screen reader focus is also not affected, so a screen reader user will think nothing happened or that something isn't working.

The way I typically see this sort of "jump to heading" functionality implemented is by using a link with href="/#heading-name". In any case, this button should do something useful to and understandable by keyboard and screen reader users.

### LINKED ISSUE #4498: [Accessibility] Focus issues with Share dialog
Environment:

Steps to reproduce:
2. Enter a valid URL and press Analyze
3. Turn on a screen reader
4. Navigate with the screen reader to the Share button
5. Activate the Share button
6. Attempt to navigate with the screen reader to the dialog

In my testing, NVDA, iOS VoiceOver, macOS VoiceOver, and ChromeVox screen readers were all unable to navigate to the dialog.

7. Turn off the screen reader and use the keyboard (tab, shift-tab, enter) to navigate the dialog

When the dialog is open, pressing tab should not close the dialog. It should cycle keyboard focus through focusable controls in the dialog, until the close button or escape is pressed. All controls in the dialog should also be useable with the keyboard. Currently the close button, text field containing URL, and Copy to clipboard button are all not keyboard focusable.

[Creating An Accessible Modal Dialog](https://bitsofco.de/accessible-modal-dialog/)

caroqliu:
This PR will update the "Fix It Now" button to an anchor link instead of an AMP `scrollTo` action to the same header element. The share button dialog contents will also now be appropriately focusable.

Fixes #4498, fixes #4493

### DIFF
--- a/frontend/templates/views/partials/pixi/status-intro.j2
+++ b/frontend/templates/views/partials/pixi/status-intro.j2
@@ -19,12 +19,10 @@
         {{ pixi.staticText.statusIntro.loadingCopy|replace('${finishedChecks}', '<span id="finished-checks">0</span>')|replace('${totalChecks}','<span id="total-checks">0</span>')|safe }}
       </p>
       <div class="ap-m-pixi-status-intro-banner-ctas">
-        <button class="ap-a-btn"
-            on="tap:recommendations.scrollTo(duration=400)"
-            type="button"
-            name="button">
+        <a href="#recommendations" class="ap-a-btn"
+            type="button">
           {{ pixi.staticText.statusIntro.buttonFixIt }}
-        </button>
+        </a>
         <button class="ap-a-btn ap-a-btn-light"
             on="tap:share-dialog"
             type="button"
--- a/frontend/templates/views/partials/pixi/status-intro.j2
+++ b/frontend/templates/views/partials/pixi/status-intro.j2
@@ -24,6 +24,7 @@
           {{ pixi.staticText.statusIntro.buttonFixIt }}
         </a>
         <button class="ap-a-btn ap-a-btn-light"
+            id="share-button"
             on="tap:share-dialog"
             type="button"
             name="button">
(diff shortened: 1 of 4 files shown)


frontend/templates/views/partials/pixi/status-intro.j2
```diff
@@ -19,13 +19,12 @@
         {{ pixi.staticText.statusIntro.loadingCopy|replace('${finishedChecks}', '<span id="finished-checks">0</span>')|replace('${totalChecks}','<span id="total-checks">0</span>')|safe }}
       </p>
       <div class="ap-m-pixi-status-intro-banner-ctas">
-        <button class="ap-a-btn"
-            on="tap:recommendations.scrollTo(duration=400)"
-            type="button"
-            name="button">
+        <a href="#recommendations" class="ap-a-btn"
```
matthiasrohmer, line 22: Can we keep `on="tap:recommendations.scrollTo(duration=400)"`?

matthiasrohmer (approved):
Only one question looks fine otherwise.

frontend/templates/views/partials/pixi/status-intro.j2
caroqliu: Unfortunately even with that there it will jump to the anchor tag. Another option though is to add `scroll-behavior: smooth;` on the document CSS, but I believe this has to be on the outer document and I am not sure how to add this for just the pixi page.

Issue #4522, opened 2020-09-14
[pixi] String escaping doesn't work correctly


nainar:
See below screenshot

![Screen Shot 2020-09-14 at 12 11 58 PM](https://user-images.githubusercontent.com/7919520/93127930-99d14980-f683-11ea-8c73-8a5699ff486d.png)

nainar:
cc @kristoferbaxter @caroqliu

tharders:
The problem with the escaped html entities in the recommendation text and status banner are related to this bug: https://github.com/ampproject/amphtml/issues/27153

tharders:
The provided fix keeps some problem open.
We still need to convert markdown for the text we get from Lighthouse.
The problem there is that we cannot even ensure that the text does not contain certain characters.
The other problem is that there is another [issue](https://github.com/ampproject/amphtml/issues/30273) with amp-script and innerHTML when we use `<pre>` tags which we get when we have code examples in the recommendations.

I have created a PR to work around these problems at the pixi side, but I am not very happy with it.
I had to replace the angle brackets with similar unicode characters in order to display them. 
The problem with that is the user cannot copy a working piece of code.
And it is still possible to create markup that would result in display errors like code examples with highlights.

caroqliu:
Marking this as fixed with followup issue #4591 once https://github.com/ampproject/worker-dom/pull/947 is available.

PR #4578, opened 2020-09-17
Fix/pixi text problems

tharders:
Fix test display problems:
- fast-font-display body not displayed when fired by Linter check
- Code blocks not displayed correctly in recommendations
- Markdown recommendations from PSI where not converted to html

This PR adresses problems described here:
https://github.com/ampproject/amp.dev/issues/4522

But as soon as amp-script problems are solved things done here should be changed!

### DIFF
--- a/pixi/src/ui/I18n.js
+++ b/pixi/src/ui/I18n.js
@@ -1,3 +1,19 @@
+// Copyright 2020 The AMPHTML Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import marked from 'marked';
+
 const DEFAULT_LANGUAGE = 'en';
 
 class I18n {
@@ -45,7 +61,7 @@ class I18n {
         const {body, ...props} = recommendation;
         result.push({
           id,
-          body: body || item.description,
+          body: body || marked(item.description),
           ...props,
         });
       } else {
--- a/pixi/src/utils/texts.js
+++ b/pixi/src/utils/texts.js
@@ -1,0 +1,49 @@
+// Copyright 2020 The AMPHTML Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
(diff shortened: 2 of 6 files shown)


caroqliu (approved):
Thank you for taking a look at this, especially for fixing fast font display. Could you please also file a follow up issue to undo these changes once support is added in amp-script?

pixi/src/ui/I18n.js
```diff
@@ -45,7 +61,7 @@ class I18n {
         const {body, ...props} = recommendation;
         result.push({
           id,
-          body: body || item.description,
+          body: body || marked(item.description),
```
samouri, line 64: Where does `item.description` come from? Can we not easily convert this on the server as well?

matthiasrohmer: That's coming from the Page Speed Insights API... I thought of multiple solutions, none of them really clever: a) round-trip to an amp.dev API to convert, b) proxy PSI API somewhere and convert response there, c) do not rely on the PSI texts but mirror them inside the amp.dev project

Any other ideas?

samouri: something close to (b) makes sense to me: have our api also make the psi request

pixi/src/utils/texts.js
```diff
@@ -0,0 +1,49 @@
...
+// limitations under the License.
+
+const htmlEntityMapping = {
+  // we will not use the html inside attributes
+  // so here we can use the normal characters...
+  '&quot;': '"',
+  '&#39;': "'",
+  // since the following caracters can not be used unescaped
+  // we use unicode replacements...
+  '&gt;': '\uFE65', // small greater than sign
```
samouri, line 22: I'm confused, why couldn't we replace `&gt` with `>`?

pixi/src/ui/I18n.js
tharders: By rendering the html on the server side we only fixed the problem that the apostrophe and the quote signs where not shown as entities. Since we do not use the HTML inside of an attribute we do not need to use entities at all, so simply converting them back on the client also works for us.
So I see no benefit of converting the PSI request on the server side aside from having smaller JS code on the client side. In each case we still have the problem that the code examples in some recommendations which use &lt; and &gt; are not shown correctly. Because of this issue ampproject/amphtml#27153 all html entities we have in the html code will be shown as entities on the site and not as their respective character.
That is why I also replaced those entities with other unicode characters that look the same. It's just an ugly workaround. Without some improvement on the amp side regarding the mentioned issue I see no possibility to display all our recommendations correctly without really rendering the recommendations on the server side and not using DOM manipulation to add them.

pixi/src/utils/texts.js
tharders: see comment above

tharders:
> Could you please also file a follow up issue to undo these changes once support is added in amp-script?

Done with https://github.com/ampproject/amp.dev/issues/4591

PR #4579, opened 2020-09-17
Pixi - Adapt colors in favor of  a11y

robinvanopstal:
+ add list of a11y safe colors

### DIFF
--- a/frontend/scss/components/molecules/input-bar.scss
+++ b/frontend/scss/components/molecules/input-bar.scss
@@ -115,10 +115,6 @@
       }
     }
 
-    &[disabled] {
-      background: color('silver');
-    }
-
     &.loading {
       pointer-events: none;
 
--- a/frontend/scss/components/molecules/input-bar.scss
+++ b/frontend/scss/components/molecules/input-bar.scss
@@ -116,11 +116,11 @@
     }
 
     &[disabled] {
-      background: color('boulder');
+      background: safeColor('oslo-gray');
     }
 
     &.loading {
-      background: color('boulder');
+      background: safeColor('oslo-gray');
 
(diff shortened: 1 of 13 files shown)


frontend/scss/components/molecules/pixi-primary-metric.scss
```diff
@@ -17,7 +17,7 @@
   }
 
   & + & {
-    border-top: solid 1px color('boulder');
+    border-top: solid 1px color('mercury');
```
caroqliu, line 20: Should this be restored to 2px?

frontend/scss/components/molecules/pixi-recommendations-item.scss
```diff
@@ -7,7 +7,7 @@
   $root: &;
   margin-top: 30px;
   padding: 30px 18px;
-  border: solid 1px color('boulder');
+  border: solid 1px color('mercury');
```
caroqliu, line 10: Same here, is this supposed to be 2px?

caroqliu (approved):

frontend/scss/components/molecules/input-bar.scss
```diff
@@ -108,19 +108,19 @@
...
+      fill: safeColor('white');
 
       @media (min-width: 600px) {
         display: none;
       }
     }
 
     &[disabled] {
-      background: color('boulder');
+      background: color('silver');
```
caroqliu, line 119: Is `silver` or `mercury` the light grey of choice? Or are we using three different shades of grey?

caroqliu: Also I thought we were not visually disabling the input button anymore?

caroqliu (approved):

PR #4580, opened 2020-09-17
✨ Add tooltip for field- and lab-data

robinvanopstal:


pages/content/pixi/info-texts/field-and-lab-tests.md
```diff
@@ -0,0 +1,9 @@
+---
```
caroqliu, line 1: Let's consolidate these texts into the core-web-vitals.md and use that, since the tooltip is next to the Core Web Vitals headline, correct?

caroqliu (commented):
Thanks for working on this!

caroqliu (approved):

PR #4588, opened 2020-09-18
Sample for protocol adapter

morsssss:


examples/source/1.components/amp-script/index.html
```diff
@@ -219,6 +219,45 @@
       </script>
     </div>
 
+    <!--
+      ## Data source for `amp-list`
+      An `<amp-script>` function can serve as the data source for an `<amp-list>`.
+        - Use `exportFunction()` to make the function visible to the `<amp-list>`.
+        - Specify the script and the function in the `src` attribute of the `<amp-list>`. Use the form `src="amp-script:{scriptID}:functionName"`, where `{scriptID}` is the `id` of the `<amp-script>` and `{functionName}` is the name of the exported function.
+        - You can use the `nodom` attribute in the `<amp-script>` to indicate that the `<amp-script>` won't need a DOM. This improves performance.
```
patrickkettner, line 227: neat! 
if possible to do so succinctly, it would great to mention in what way performance is improved

morsssss: I got this from @samouri , so I'd like to ask him to comment here!

* I know that when `nodom` is used, `amp-script` loads a smaller bundle ~ 2K instead of 15K. I believe this is because it won't need any of the logic WorkerDOM uses to handle the DOM API. Does that smaller bundle not include WorkerDOM at all?
* I imagine it also might mean that the JS that builds the virtual DOM doesn't need to execute (and might not even be loaded), saving time... perhaps even more JS doesn't need to execute.

Anything else?

morsssss: https://github.com/ampproject/amphtml/blob/master/extensions/amp-script/0.1/amp-script.js#L324

samouri: > Does that smaller bundle not include WorkerDOM at all?

Pretty accurate! It technically includes the `worker-dom` library still, but it removes the vdom implementation

morsssss: Ok! Put in a bit about no vdom.

samouri (approved):

examples/source/1.components/amp-script/index.html
```diff
@@ -219,6 +218,45 @@
...
+    <!--
+      ## Data source for `amp-list`
+      An `<amp-script>` function can serve as the data source for an `<amp-list>`.
+        - Use `exportFunction()` to make the function visible to the `<amp-list>`.
+        - Specify the script and the function in the `src` attribute of the `<amp-list>`. Use the form `src="amp-script:{scriptID}:functionName"`, where `{scriptID}` is the `id` of the `<amp-script>` and `{functionName}` is the name of the exported function.
+        - You can use the `nodom` attribute in the `<amp-script>` to indicate that the `<amp-script>` won't need a DOM. This improves performance, because `amp-script` doesn't need to load or execute its virtual DOM implementation.
+    -->
+    <div class="sample">  
+      <amp-script id="dataFunctions" script="protocol-adapter-script" nodom></amp-script>
+      <script id="protocol-adapter-script" type="text/plain" target="amp-script">
```
samouri, line 230: nit: I'm not sure if we want to keep the name "protocol adapter". It constrains the capability to `fetch`/xhr, whereas the JS can basically do anything (localStorage, websockets, math computations, etc.). Maybe `<script id="dataFunctions-script">`

morsssss: made id even more explicit!

samouri (approved):

morsssss:
@patrickkettner it looks like I need your approval too to merge this sucker...

patrickkettner (approved):

patrickkettner:
@morsssss sorry for blocking!

morsssss:
No problem! You have been... busy. To say the least.

Issue #4458, opened 2020-09-08
Pixi: Flaky Share URL construction


caroqliu:
Haven't dug into this further, but occasionally all tests are complete with a final status banner update:
![image](https://user-images.githubusercontent.com/10456171/92523565-93a81c80-f1ee-11ea-8f0e-29184eadcb45.png)

Note this does not happen consistently.

caroqliu:
Closing as this appears fixed

caroqliu:
Reopening, seeing this again with https://amp.reddit.com/t/coronavirus/
This also appears in the console: 
![image](https://user-images.githubusercontent.com/10456171/92737792-6de35a80-f349-11ea-8d1c-e4868d87cd3b.png)

cc @tharders any ideas?

caroqliu:
New info thanks to #4601:
![image](https://user-images.githubusercontent.com/10456171/93820857-c52bd980-fc2b-11ea-83a1-686b3ad479ff.png)

I believe the original issue no longer surfaces but we should keep this open for next steps.

caroqliu:
@matthiasrohmer 
Another where the error comes from share:
![Screen Shot 2020-09-22 at 9 50 07 AM](https://user-images.githubusercontent.com/10456171/93891310-4aa99b00-fcb9-11ea-8983-d776a8268e8d.png)

It also appears the share field is empty: 
![Screen Shot 2020-09-22 at 9 51 19 AM](https://user-images.githubusercontent.com/10456171/93891326-4e3d2200-fcb9-11ea-98e7-9ccf78496d6c.png)

Note that this error leads to the undesired status banner for `amp.dev` input: all tests have passed so we should be able to give a conclusion on page experience; instead we surface "something went wrong"
![Screen Shot 2020-09-22 at 9 50 15 AM](https://user-images.githubusercontent.com/10456171/93891348-51d0a900-fcb9-11ea-8d82-ee69d479b5d6.png)

PR #4601, opened 2020-09-21
🐛 Do not take AMP.getState returning something for granted in Pixi

matthiasrohmer:
Had no luck reproducing #4458 specifically but my best guess is that `AMP.getState` sometimes throws instead of returning something. We've guarded this for the `InputBar` but didn't for the other occurrences until now.

### DIFF
--- a/pixi/src/ui/I18n.js
+++ b/pixi/src/ui/I18n.js
@@ -26,12 +26,18 @@ class I18n {
   }
 
   async init() {
-    const pixiContent = await Promise.all([
-      AMP.getState('pixi.i18n'),
-      AMP.getState('pixiStatusBanners'),
-      AMP.getState('pixiRecommendations'),
-      AMP.getState('pixiInfoTexts'),
-    ]);
+    let pixiContent = [];
+    try {
+      pixiContent = await Promise.all([
+        AMP.getState('pixi.i18n'),
+        AMP.getState('pixiStatusBanners'),
+        AMP.getState('pixiRecommendations'),
+        AMP.getState('pixiInfoTexts'),
+      ]);
+    } catch (e) {
+      console.error('Failed to initialize translations', e);
+    }
+
     const i18nConfig = JSON.parse(pixiContent[0]);
     this.language = i18nConfig.language;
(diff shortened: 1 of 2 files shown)


pixi/src/ui/I18n.js
```diff
@@ -26,12 +26,18 @@ class I18n {
...
-    ]);
+    let pixiContent = [];
+    try {
+      pixiContent = await Promise.all([
+        AMP.getState('pixi.i18n'),
+        AMP.getState('pixiStatusBanners'),
+        AMP.getState('pixiRecommendations'),
+        AMP.getState('pixiInfoTexts'),
+      ]);
+    } catch (e) {
```
tharders, line 37: This will help nothing, since two lines later we will access the pixi state elements that will not be there.
You should write a unit test for I18n, where you can produce all the different error cases and test that the API will handle them gracefully.

tharders: Also when the init function handles all error cases, the question is what will happen later when the get... functions are called and the expected content is not there?

I think it makes more sense to handle errors at a higher level and show a static status banner text with some "internal error" message.
Of course we have to ensure that all errors that cannot really be handled bubble up.
We have to double check the error handling in all those async functions.

tharders (changes requested):
Please have a look at the comments in the code

pixi/src/ui/I18n.js
matthiasrohmer: I could have bet that you are going to comment this... these are valid points but my intention for now is just to track down if that's really the problem with #4458.

Created https://github.com/ampproject/amp.dev/issues/4604 for further discussion, I am not intending to fix this in this PR.

tharders: Would have been nice to know the intention of this PR right from the start ;-)

tharders (approved):

pixi/src/ui/I18n.js
matthiasrohmer: Then there would have been no bet to win 😁 Thanks though for taking a look!

PR #4609, opened 2020-09-21
Pixi: Add inline explainer for lagging field data

### LINKED ISSUE #4543: Pixi: Clarify Field test and Lab test more
- [x] Information tips for Field and Lab data - See here: https://docs.google.com/document/d/15KTgFoQnmPEDhHyfANe-Ln5_qG20rTK1QOe4F6IfejQ/edit
- [ ] Field data should have a info banner inline explaining that this is 28 days aggregated data hence data will lag.

caroqliu:
Fixes #4543

Field data:
image
image

**Optional**: Also added the following note to lab data. (This was originally planned for the "Lab data" info tip copy. If undesired in this format I will restore it to the info .md where it should be:
image
image


pages/content/pixi/index.md
```diff
@@ -25,6 +25,8 @@ staticText:
     tabsAriaLabel: Field and lab data tabs
     fieldData: Field data
     labData: Lab data
+    fieldDataExplainer: This data updated daily for the trailing 28-day period.
+    labDataExplainer: Please note that lab tests aren't reliable as they may vary per run.
```
patrickkettner, line 29: Not sure what the point being made here is?

caroqliu: Taken from https://docs.google.com/document/d/15KTgFoQnmPEDhHyfANe-Ln5_qG20rTK1QOe4F6IfejQ/edit#
I believe the original intention was to caveat that lab data is a controlled/simulated test, especially if field data is not available.

However I'm all good with removing this entirely from the current proposed UI.

caroqliu: We could also update to `Please note that lab tests may vary per run.` @nainar

caroqliu: via Naina: We can also call out specifically that they are synthetic tests and do not impact page experience.

patrickkettner (approved):
I think this is good!￼For what it's worth, it is not the most recent 28 days, but rather data is released every 28 days. So the data could literally be from yesterday, or 28 days ago.￼

PR #4613, opened 2020-09-21
Pixi: Correctly identify CTA buttons

### LINKED ISSUE #4602: Pixi: do not show CTAs for non-AMP pages
Bildschirmfoto 2020-09-21 um 15 20 41

caroqliu:
Add `data-cta` attribute to the targeted anchor elements we want to modify. This fixes #4602, which was a consequence of the inline text having a hyperlink of its own (to PSI).


matthiasrohmer (commented):
Should we move to `id`s instead? And then have distinct variables for each of the buttons in the JS? This would make the part thats currently accessing `anchors[0]` and `anchors[1]` more readable.

caroqliu:
> Should we move to `id`s instead? And then have distinct variables for each of the buttons in the JS? This would make the part thats currently accessing `anchors[0]` and `anchors[1]` more readable.

Great idea! Done, PTAL. :)

matthiasrohmer (approved):
Looks great now, thanks 🙂

PR #4618, opened 2020-09-22
components/reference/amp-3d-gltf: fix spelling of glTF

zhuowei:
It's glTF, with a lowercase L, and not gITF with an uppercase I.


matthiasrohmer (approved):
Thanks!

PR #4627, opened 2020-09-22
💄 Pixi - Input error styling

robinvanopstal:


frontend/scss/components/molecules/input-bar.scss
```diff
@@ -135,14 +136,27 @@
...
-    padding-left: 25px;
-    color: color('red');
+    position: absolute;
+    width: calc(100% - 30px);
+    padding: 10px 15px 2px 20px;
+    border-bottom-left-radius: 6px;
+    border-bottom-right-radius: 6px;
+    font-weight: bold;
+    color: safeColor('white');
+    background: safeColor('guardsman-red');
```
matthiasrohmer, line 148: Wondering if we should move away from fancy color names for the safe colors. In an ideal world safe colors would not hold more than three shades of a color so something like `dark-red`, `light-red` and `red` would work again.

matthiasrohmer (approved):

PR #4647, opened 2020-09-22
Redirects for pixi

nainar:
Should we add any other?


caroqliu (commented):
Please also add `/vitals` and `/webvitals`

nainar:
@caroqliu done and added web-vitals too

matthiasrohmer (approved):

PR #4649, opened 2020-09-22
♻️ Stop replacing HTML entities in Pixi

matthiasrohmer:
This fixes AMP to RTV 002009190410000 where https://github.com/ampproject/amphtml/issues/27153 should be fixed. Quick test I've done with `<test>` in one of the recommendations worked fine.


caroqliu (approved):
Thank you!! 🎉

Issue #4608, opened 2020-09-21
Pixi: Mobile view status covers "Core Web Vitals" headline


caroqliu:
![image](https://user-images.githubusercontent.com/10456171/93783942-55e9c180-fbfa-11ea-8aa8-b3a13ed6a796.png)

Issue #4470, opened 2020-09-09
Pixi: Tablet view too narrow


caroqliu:
![image](https://user-images.githubusercontent.com/10456171/92653593-96714300-f2bc-11ea-9d05-5fc245b23ab4.png)

nainar:
Can we make the URL bar smaller on tablet?

caroqliu:
> Can we make the URL bar smaller on tablet?

@nainar Do you mean smaller in width?

nainar:
yes

PR #4656, opened 2020-09-23
Pixi - Fix layout bugs

robinvanopstal:
Resolves: #4608 and resolves: #4470


caroqliu (approved):

caroqliu (commented):
I'm seeing a white margin to the left but not the right of the blue gradient, is this expected? 
image

PR #4658, opened 2020-09-23
Pixi - Add next advice button 

robinvanopstal:


robinvanopstal:
@googlebot I consent.

matthiasrohmer:
@googlebot I consent.

pages/content/pixi/index.md
```diff
@@ -45,6 +45,7 @@ staticText:
     checkManually: Check manually
   recommendations:
     headline: Take action. Improve your AMP site
+    nextAdvice: Next advise
```
caroqliu, line 48: "Next advice"

pixi/src/ui/recommendations/RecommendationsView.js
```diff
@@ -182,4 +197,9 @@ export default class RecommendationsView {
       recommendation.hidden = !commonValues.length;
     }
   }
+
+  onClickNext(recommendation) {
+    recommendation.classList.remove('expanded');
+    recommendation.nextSibling.classList.add('expanded');
```
caroqliu, line 203: We also want to set `aria-expanded` to `"true"` (note it's a string).

By the way, the button is not available when there is not a `nextSibling` correct?

pixi/src/ui/recommendations/RecommendationsView.js
```diff
@@ -83,11 +87,22 @@ export default class RecommendationsView {
...
+            this.onClickNext(recommendation);
+          });
+        } else {
+          nextButton.remove();
+        }
       }
 
       recommendation.style = null;
-      recommendation.id = `recommendation-${value.id}`;
+      title.id = `recommendation-${value.id}`;
```
caroqliu, line 105: Is it OK if we restore it to set/link to the `recommendation.id` instead of `title.id`? I understand we need more padding to account for the top banners, however as it is the screen reader will get trapped on the title element after clicking "Next advise"

frontend/templates/views/partials/pixi/recommendations-item.j2
```diff
@@ -1,12 +1,21 @@
...
+      role="button"
+      tabindex="0"
+      aria-expanded="false">
     <div class="ap-m-pixi-recommendations-item-header-title"></div>
     <div class="ap-m-pixi-recommendations-item-header-toggle">
     </div>
   </header>
-  <div class="ap-m-pixi-recommendations-item-body" role="region">
+  <div class="ap-m-pixi-recommendations-item-body">
+    <div class="ap-m-pixi-recommendations-item-body-text" role="region">
```
caroqliu, line 13: Let's keep `role="region"` on the `.ap-m-pixi-recommendations-item-body`

pixi/src/ui/recommendations/RecommendationsView.js
```diff
@@ -182,4 +197,9 @@ export default class RecommendationsView {
       recommendation.hidden = !commonValues.length;
     }
   }
+
+  onClickNext(recommendation) {
+    recommendation.classList.remove('expanded');
```
caroqliu, line 202: This element also needs `aria-expanded` set to `false`

pixi/src/ui/recommendations/RecommendationsView.js
robinvanopstal: Correct :)

caroqliu (commented):
Ah it looks like this comment never got sent, I'll send a PR with this suggested change as well. Sorry about that!

Issue #4687, opened 2020-09-24
Pixi: Next advice button should set "aria-expanded" on header


caroqliu:
The `aria-expanded` attributes set here: 
https://github.com/ampproject/amp.dev/blob/cc2b5862a7a290be82e0a843aaa0b9cfb3284c4e/pixi/src/ui/recommendations/RecommendationsView.js#L201-L206

Should actually be set on the `header` element (`#header-${id}`) rather than the `recommendation` element (`#recommendation-${id}`) -- sorry about this mixup!

PR #4700, opened 2020-09-25
♿️ Set aria expanded on recommendation header

robinvanopstal:
Resolves: #4687


matthiasrohmer (approved):

pixi/src/ui/recommendations/RecommendationsView.js
```diff
@@ -202,6 +202,8 @@ export default class RecommendationsView {
     recommendation.classList.remove('expanded');
     recommendation.setAttribute('aria-expanded', 'false');
```
caroqliu, line 203: Want it for this one as well

PR #4702, opened 2020-09-25
Pixi - Ltr scales

robinvanopstal:


matthiasrohmer:
@googlebot I consent.

caroqliu:
Thanks for working on this Robin! I should have clarified more - I don't think we need to reverse the direction of the graph components, we mainly just want the text and scales to be on opposite sides, like this:
image

Most of this is doable by placing `dir="rtl"` on the `.ap-m-pixi-primary-metric` element.

robinvanopstal:
Can we close this one in favor of #4757?

matthiasrohmer:
Sounds good!

PR #4703, opened 2020-09-25
🔧  Enable AGGRESSIVE_BUNDLE for Pixi build

matthiasrohmer:
This disables babel transpilations for Pixi as all JS features we are using (`async`/`await` mainly) are gracefully handled by Closure Compiler already. This brought down bundle size from 110 KiB to 98.7 KiB.

Additionally this PR enables CCs' `AGGRESSIVE_BUNDLE` mode which surprisingly didn't require much work. Savings aren't that big though, it brought down bundle size to 98.2 KiB with slight DX degredations.

We could even further bring it down to by only supporting ES2015 capable browsers (82.4KB) but this would exclude IE11.

Biggest optimization would be to get rid of `marked` though, without it and ES2015+ browser support we would land at 48.6 KiB. Creating a follow-up issue to discuss this.

### DIFF
--- a/pixi/src/checks/constants.js
+++ b/pixi/src/checks/constants.js
@@ -17,9 +17,4 @@ const UNIT_DEC = {name: '', conversion: 100, digits: 2};
 const UNIT_SEC = {name: 's', conversion: 1000, digits: 1};
 const UNIT_MS = {name: 'ms', conversion: 1, digits: 0};
 
-export {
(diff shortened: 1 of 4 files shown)


caroqliu (approved):

matthiasrohmer:
Created https://github.com/ampproject/amp.dev/issues/4704 for further discussion.

PR #4706, opened 2020-09-25
Pixi - Recommendation aria expanded

robinvanopstal:


pixi/src/ui/recommendations/RecommendationsView.js
```diff
@@ -204,7 +204,9 @@ export default class RecommendationsView {
 
   onClickNext(recommendation) {
     recommendation.classList.remove('expanded');
-    recommendation.setAttribute('aria-expanded', 'false');
+    recommendation
```
caroqliu, line 207: When this is event handler is registered on the `recommendation` element, can we also pass in its `header` element as a parameter to use here? Could save us an additional call with `querySelector`.

caroqliu (approved):

Issue #4675, opened 2020-09-23
Pixi: Add Analytics support


nainar:
Add analytics support for the following
 - [ ] How many people used this page
 - [ ] How many many people viewed the recommendations section overall - GA viewability trigger
 - [ ] How often Share button is used
 - [ ] Clicking on take action: x recommendations option
 - [ ] Clicking on Filing GH issue
 - [ ] HOW TO TRACK THAT PEOPLE ARE SEEING SUCCESS WITH PIXI?

nainar:
We should also track API failure

caroqliu:
via #4634 - we should also track tooltip engagement.

caroqliu:
via @matthiasrohmer:
> We could think about tracking the analyzed URL as well - though this might be hard to normalize. Another problem is that the trigger would also fire for invalid URLs...

caroqliu:
Note to add a trigger for when the "Fix it now" and "Investiate" buttons are clicked.

caroqliu:
Update:
- Merge first visibility triggers on field data, lab data, and recommendations. Investigate for a work around to get nth visibility too.
- Add click triggers for "Fix it Now", "Investigate", "Field data" tab, "Lab data" tab, Recommendation filter categories, and toggle button for every recommendation (on open)

Open question:
- How to track API failures?

caroqliu:
Bumping priority down as we have the following click triggers:
- Field data tab
- Lab data tab
- Expand recommendation item
- "Next advice" on recommendation item
- Open tooltip
- "File a GH issue"
- "Share"
- "Fix it Now" / "Make Improvements"
- "Investigate" (API failure has occurred)
- "Take Action: X Recommendations"
- Recommendation filter category

And the following visibility triggers:
- pageview
- nth view of recommendations
- 1st view of loaded field data
- 1st view of loaded lab data

With the following remaining open tasks: 
- nth view of field and lab data that either differentiates between or discards impressions on loading or error states of these UIs
- Track API failures - One option is to do this on the server with support of #4758

PR #4709, opened 2020-09-25
Pixi: Add amp-analytics triggers

caroqliu:
Partial for #4675


matthiasrohmer (approved):
Seems to work as intended 🙂  The selectors are a little fishy but I have no quick idea how to improve them.

We could think about tracking the analyzed URL as well - though this might be hard to normalize. Another problem is that the trigger would also fire for invalid URLs...

caroqliu:
Merging as is, will add your recommendation to the open issue, thanks!

Issue #4712, opened 2020-09-25
[Accessibility] Expandable sections aren't expandable/collapsible with enter/space


PeterSaathoff-Harshfield:
Environment:
Any OS/browser

Steps to reproduce:
1. Open https://amp-dev-staging.appspot.com/page-experience/
2. Enter a valid URL and press Analyze
3. Press tab to move keyboard focus to the expandable sections below the "Take action. Improve your AMP site" heading
4. When a section has focus, press enter and space

Expected:
These buttons should accept both enter and space to expand/collapse them

Actual:
Enter has no effect and space pages down

![image](https://user-images.githubusercontent.com/30604285/94316502-4f7c8200-ff39-11ea-8df1-f82a83ea66d6.png)

robinvanopstal:
Hi @PeterSaathoff-Harshfield, this will only work on press `enter` since `e.preventDefault()` is not allowed in `amp-script`.

PR #4727, opened 2020-09-28
Pixi - Expand/collapse recommendation on press enter

robinvanopstal:
Resolves: #4712

### DIFF
--- a/frontend/scss/components/templates/pixi.scss
+++ b/frontend/scss/components/templates/pixi.scss
@@ -31,7 +31,7 @@
     }
 
     @media (min-width: 1024px) {
-      padding: 120px 35px 0 65px;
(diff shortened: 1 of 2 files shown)


caroqliu (commented):
Ah thanks for taking a look at this! I should have clarified -- while the event listener is a good approach to add functionality for "Enter", I don't think it will work for "Space" because by default that key will make the page scroll down. Seeing as we cannot `preventDefault` in amp-script this will cause problems. 

I would suggest implementing with a native `<button>` element which has browser support for `Enter` and `Space` built in -- would it be okay to prefer #4728 instead?

matthiasrohmer:
Closing in favor of #4728.

PR #4729, opened 2020-09-28
Add local storage sample for amp-script

morsssss:


examples/source/1.components/amp-script/index.html
```diff
@@ -484,5 +484,37 @@ <h2 id="translated-greeting"></h2>
...
+        const input = document.getElementById('username-input');
+        const display = document.getElementById('username-display');
+
+        const oldUsername = localStorage.getItem('username');
+        display.innerHTML = oldUsername ? `Your username is <b>${oldUsername}</b>.` : 'Username is not set';
+
+        function setUsername() {
+          const newUsername = input.value;
+          localStorage.setItem('username', newUsername);
+          display.innerHTML = `Username changed to <b>${newUsername}</b>.`;
```
westonruter, line 513: Bad thing can happen if someone puts illegal HTML as the input value, e.g. `<script>hello</script>`:

image

Instead of using `innerHTML`, probably better to use `textContent`.

examples/source/1.components/amp-script/index.html
```diff
@@ -484,5 +484,37 @@ <h2 id="translated-greeting"></h2>
...
     </div>  
+    <!--
+        ## Local storage
+        Your script has access to the Web Storage API. This lets you persist user information between browser sessions.
+        Here, we use `localStorage` to track a username.
+        If you set the username, then reload this page, the username will remain set.
+    -->
+    <div>
+      <amp-script layout="fixed-height" script="local-storage-script" height="110" class="sample">
+        <div id="username-display"></div>
```
westonruter, line 495:
```suggestion
        <div id="no-username-display">Username is not set.</div>
        <div id="username-display" hidden>Your username is <b id="persisted-username"></b>.</div>
```

examples/source/1.components/amp-script/index.html
```diff
@@ -484,5 +484,37 @@ <h2 id="translated-greeting"></h2>
...
+        const oldUsername = localStorage.getItem('username');
+        display.innerHTML = oldUsername ? `Your username is <b>${oldUsername}</b>.` : 'Username is not set';
+
+        function setUsername() {
+          const newUsername = input.value;
+          localStorage.setItem('username', newUsername);
+          display.innerHTML = `Username changed to <b>${newUsername}</b>.`;
+        }
+
+        submit.addEventListener('click', setUsername);
```
westonruter, line 516:
```suggestion
        const submit = document.getElementById('username-submit');
        const input = document.getElementById('username-input');
        const display = document.getElementById('username-display');
        const noDisplay = document.getElementById('no-username-display');
        const persistedUsername = document.getElementById('persisted-username');
  
        function setUsername() {
          const newUsername = input.value || localStorage.getItem('username');
          if ( newUsername ) {
            persistedUsername.textContent = newUsername;
            localStorage.setItem('username', newUsername);
            } else {
              localStorage.removeItem('username');
            }
          display.hidden = (newUsername === '');
          noDisplay.hidden = (newUsername !== '');
        }
  
        setUsername();
        submit.addEventListener('click', setUsername);
```

westonruter (changes requested):

examples/source/1.components/amp-script/index.html
morsssss: This is true. But I'm not sure I want to discourage `innerHTML` for that reason... well, ok, I was conflicted about using it because it's not supported for arbitrary HTML, just in simple cases like mine.

I added `innerHTML` because I wanted to make the username boldface. Your solution reflects a better separation of concerns, but it's substantially longer - and we try to keep these samples short. (Sebastian would probably already say mine is too long.)

morsssss: Ok! I've changed this sample to use `textContent` instead. Could you take another look?

examples/source/1.components/amp-script/index.html
```diff
@@ -484,5 +488,40 @@ <h2 id="translated-greeting"></h2>
...
+        ## Local storage
+        Your script has access to the Web Storage API. This lets you persist user information between browser sessions.
+        Here, we use `localStorage` to track a username.
+        If you set the username, then reload this page, the username will remain set.
+    -->
+    <div>
+      <amp-script layout="fixed-height" script="local-storage-script" height="110" class="sample">
+        <div>
+          <span>Current username: </span>
+          <span id="username-display"></span>
```
westonruter, line 501: Any reason to not use a [`b`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/b) tag here for this? Seems ideally suited semantically as well:

> The **HTML Bring Attention To element** (`<b>`) is used to draw the reader's attention to the element's contents, which are not otherwise granted special importance.

```suggestion
          <b id="username-display"></b>
```

Then the style change wouldn't be needed either.

examples/source/1.components/amp-script/index.html
```diff
@@ -98,6 +98,10 @@
         color: #666;
         font-weight: bold;
       }
+
+      #username-display {
+        font-weight: bold;
+      }
```
westonruter, line 104: Per below:

```suggestion
```

westonruter (approved):

morsssss:
Interesting! I never use `<b>` semantically like this... done!

PR #4730, opened 2020-09-28
Pixi: Add visibility analytics

caroqliu:
To track engagement with loaded field/lab data, as well as recommendations suggested. Partial for #4675

This PR adds a second instance of `amp-analytics` solely to track visibility of elements that get dynamically added by `amp-script`. The `amp-analytics` element is instantiated with `display: none` until the analysis has completed, at which point the element is shown via CSS, lays out, and triggers appropriate visibility events.

Open question:
- How often can visibility analytics be triggered? Are they throttled at all? Context: On subsequent runs of Pixi, I do not see the analytics event firing for visibility of the targets (field data, lab data, recommendations) elements even though they did the first time. This could mean that if someone runs multiple URLs and views these elements each time, we would only count once total. 

Another option:
To track engagement with the data, we can see when a user clicks on the "Field" and "Lab" data Tabs, as well as interacts with any recommendation filters and expand/collapse buttons. This could be more comprehensive than tracking the first visible event. @nainar 

/cc @micajuine-ho to review as well if you have any advices here :)


caroqliu:
This is ready for review @matthiasrohmer :)

pages/content/amp-dev/page-experience.html
```diff
@@ -43,6 +43,48 @@
   {% set x=info_texts.__setitem__(node.base, nodeDict) %}
 {% endfor %}
 
+{% set config = {
```
matthiasrohmer, line 46: Nit: rename to `analyticsConfig`

matthiasrohmer (approved):
Interesting... great you've got it working 🙂

PR #4745, opened 2020-09-29
Pixi: Add additional click engagements

caroqliu:
Partial for #4675

Also allows `repeat` visibility events on the recommendations while keeping only first-visibility events on the field data and lab data sections, deferring to click events on their tabs for engagements on subsequent runs of Pixi. This is because repeat visibility on field data and lab data currently does not differentiate between if they are showing loading states or error states.


caroqliu:
This is now ready for review @matthiasrohmer :)

matthiasrohmer (approved):

matthiasrohmer:
You've used `ap-m-pixi-recommendations-item-next-advice` for styling in #4750 but use `ap-m-pixi-next-advice` here for tracking. Can we unify this? 🙂  This also causes merge conflicts currently.

caroqliu:
> You've used `ap-m-pixi-recommendations-item-next-advice` for styling in #4750 but use `ap-m-pixi-next-advice` here for tracking. Can we unify this? 🙂 This also causes merge conflicts currently.

Good idea! Fixed this with the recent merge with future. :)

Issue #4379, opened 2020-08-25
[Pixi] Internationalization: Support right to left languages


nainar:
UI: Add support for RTL languages
Page layout: Add support for RTL languages.

matthiasrohmer:
via @patrickkettner: especially consider this for the PSI scales.

patrickkettner:
After checking with some speakers, Arabic would be RTL for numbers, Hebrew should be LTR

caroqliu:
#4702 fixes the orientation of the scales for RTL. 

Also need to do:

- [ ] Field and lab data tabs should be right-aligned on the CWV container
- [ ] Recommendation items and filter should be right-aligned, with "Next advice" available on the left side (dir="rtl" shoulddo most of the automatically)


Open question:
- Should the visual itself be RTL in addition to be oriented left of the text? i.e. Should the green page loads bar be right-aligned over the grey bar? Should the lab data visual start with the green region on the right, yellow region in the middle, and red region on the left?

caroqliu:
Now that #4750 and #4757 are merged this issue is fixed.

PR #4750, opened 2020-09-30
Pixi: Support RTL direction on full page

caroqliu:
Partial for #4379, which will be fixed once both this and #4702 are merged.

Desktop:
image
image
image

Mobile: 
image
image
image
image
image


matthiasrohmer (approved):
Thanks a lot!

PR #4753, opened 2020-09-30
Run static images thru lossless image conversion

kristoferbaxter:
Ran the static image directory thru lossless image compression tools and found some minor improvements to filesize.


matthiasrohmer (approved):
Thanks Kris! Not quite sure if this has effect for end-users though as most images will be served through [thumbor](https://github.com/thumbor/thumbor) which might increase file sizes in some circumstances.

Issue #4765, opened 2020-10-01
Pixi: Copy changes tracking issue


caroqliu:
Related to: #4523

caroqliu:
Tagging #4741 as this PR also recently changed copy.

PR #4766, opened 2020-10-01
Pixi: Surface generic error before other linter information

caroqliu:
Otherwise, in the case that the linter response fails, we were originally surfacing "This page is invalid AMP". Tagging #4765 as this is a change in copy.

Now we surface:

image


nainar:
Let's remove the share button too

matthiasrohmer (approved):

PR #4235, opened 2020-07-28
add #BlackLivesMatter banner

patrickkettner:
ping @nainar

Issue #4665, opened 2020-09-23
Remove position: sticky from AMP logo in header


kevinkimball:
### Current behavior
The anchor tag wrapping AMP logo within the sticky header is also marked as sticky. This causes the browser to lay it out as a separate layer. Having too many layers can degrade performance and increase rendering time. 

Please also provide:
- Page link: amp.dev
- (Optional) Screenshots/videos:
<img width="230" alt="Screen Shot 2020-09-23 at 9 42 45 AM" src="https://user-images.githubusercontent.com/6886483/94043624-08985c00-fd82-11ea-9565-714cc0ee4de6.png">
<img width="213" alt="Screen Shot 2020-09-23 at 9 42 51 AM" src="https://user-images.githubusercontent.com/6886483/94043633-0cc47980-fd82-11ea-880d-35b677385e71.png">

### Expected behavior/code
Position: sticky is not necessary as the element is within the sticky header, so it should be removed 

### Reproduction procedure
Please provide the steps to reproduce the problem:

1. Visit amp.dev
2. Open rendering tab in CDT and click "layer borders"
3. remove position: sticky to see that the layer is removed and behavior is not impacted 

- What browser are you using?
Chrome 87
- What O/S are you using?
OSX

kevinkimball:
I actually see that the hamburger menu (which is not currently located inside of the header) is also affected by this since it is using position: fixed. This also causes a jump during loading where the hamburger moves up and down independently of the sticky header.  

<img width="379" alt="Screen Shot 2020-09-23 at 9 55 49 AM" src="https://user-images.githubusercontent.com/6886483/94044850-8ad55000-fd83-11ea-97f4-410d67b3da16.png">

The ideal solution is to make them all sibling elements. I believe flexbox is already used in the page so that might be a suitable option here to retain the existing display logic.

matthiasrohmer:
Thanks for your research on this! I guess the weird layering behavior is a relict from when we used `input`s to toggle the burger-menu state.

I'll have a closer look next week to see how things could be simplified.

kevinkimball:
@matthiasrohmer No problem. Just to give you some context, Kristofer asked me to help ensure amp.dev will pass CWV before the Pixi launch.

Issue #4466, opened 2020-09-09
Make sure survey banner doesn't violate CLS scores


nainar:
The current survey banner causes CLS issues. 

We should fix this.

nainar:
Reopening because this is a fundamental issue with the way the banner is implemented. This issue will persist next quarter when we bring the survey back.

matthiasrohmer:
Thanks for bringing this up! When Sebastian and I were implementing the banner we thought that having the banner dismissible is bigger UX win than what we lose by the CLS... but we both weren't really happy how it's built since we are abusing the `.amp-active` CSS class.

Any advise on how to proceed would be much appreciated 🙂

patrickkettner:
@kevinkimball would you be able to weigh in on a better pattern to use for a dismiss-able banner?

kevinkimball:
@patrickkettner Since it's a notification and dismissable, could use a fixed position below the navigation bar? Not sure how desirable it would be from the UI perspective but it would resolve the layout shift issue

PR #4770, opened 2020-10-02
Banner for AMP Fest

morsssss:
This puts up the banner for AMP Fest.

Note that, to avoid layout problems, I had to undo the work that was done to put up the BLM banner in https://github.com/ampproject/amp.dev/pull/4235 - as that PR wasn't compatible with other banner sizes. Going forward, we need to make the header more flexible so that it can accommodate banners of different sizes!

### DIFF
--- a/frontend/templates/views/partials/banner.j2
+++ b/frontend/templates/views/partials/banner.j2
@@ -1,7 +1,23 @@
 {% do doc.styles.addCssFile('/css/components/molecules/banner.css') %}
-
+<!--
 <a id="blm-banner"
    target="_blank"
    href="https://blacklivesmatter.com">
   #BlackLivesMatter
 </a>
+-->
+<div class="ap-m-banner">
+  <a id="top-banner"
+     class="ap-m-banner-link"
+     target="_blank"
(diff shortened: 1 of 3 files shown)


frontend/templates/views/partials/banner.j2
```diff
@@ -1,7 +1,23 @@
 {% do doc.styles.addCssFile('/css/components/molecules/banner.css') %}
-
+<!--
```
kristoferbaxter, line 2: Do you need to comment this out or can you just delete it? (it's a git revert away)

morsssss: You're right. I just didn't feel good about removing it :)

matthiasrohmer (approved):
This should work for now. When working on https://github.com/ampproject/amp.dev/issues/4665 and https://github.com/ampproject/amp.dev/issues/4466 we should also think about implementing the banner properly and flexible once instead of always poking this Frankenstein.

Do we also want to implement the BLM banner permanently then?

morsssss:
I think we want to make it so that amp.dev can accommodate banners of different heights, be they BLM or other!

(I like "poking this Frankenstein" :) )

PR #4775, opened 2020-10-02
Change shortlink for optimize-amp

morsssss:
We have about 4 shortlinks for the optimizer guide, but none for the really useful [guide to speeding up AMP pages yourself](https://amp.dev/documentation/guides-and-tutorials/optimize-and-measure/optimize_amp). I need a shortlink for a talk... I hope it's ok to switch this one! It's named ideally for the optimize_amp page....


patrickkettner:
its generally bad practice to swap out urls, do we have any traffic data on if it is used at all? or where it was/is listed?

morsssss:
Yeah, you're probably right. I frankly doubt people are using the link I changed... but they certainly could be. I don't know whether we track traffic on these go links.

We need this shortlink for a video that's about to be complete, so I just changed it to something previously unused. Thanks for keeping me honest!

patrickkettner (approved):

patrickkettner:
@sebastianbenz @matthiasrohmer any idea if there is analytics for go links?

PR #4777, opened 2020-10-03
⚡️ Use amp-carousel for news slider

robinvanopstal:


frontend/scss/components/organisms/news.scss
```diff
@@ -77,13 +68,21 @@
       padding-right: 0;
     }
 
+    div {
```
matthiasrohmer, line 71: This would add the `transition` property to all `ap-o-news-card-title`, `ap-o-news-card-image` and `ap-o-news-card`. Can we use a more specific selector here?

matthiasrohmer (changes requested):

patrickkettner:
@robinvanopstal could you sign the new cla when you have a chance?

matthiasrohmer (approved):
One small request left, besides that LGTM.

PR #4801, opened 2020-10-06
✅  Add Pixi to smoke tested pages

matthiasrohmer:
This would have prevented https://github.com/ampproject/amp.dev/pull/4792. Before a build goes live on amp-dev-staging.appspot.com or amp.dev itself it now needs to be able to serve a AMP-valid version of Pixi, otherwise it blocks build.


caroqliu (approved):
Thank you!

Issue #4818, opened 2020-10-08
Pixi status banner not shown in ar version


tharders:
When checking a page in the AR version of pixi the status banner is not shown.
https://amp-dev-staging.appspot.com/ar/page-experience/

![image](https://user-images.githubusercontent.com/6458408/95420567-f69ae980-093b-11eb-8d2f-117140ba7cb7.png)

I have checked different pages with different status results.
In an other language this works fine.
Maybe it is an LTR issue.

patrickkettner:
it doesn't appear that the amp-script is ever being loaded. The blob is not in chrome's devtools as a context option, and none of the code in it appears to be running.

patrickkettner:
this seems like potentially an issue with worker-dom?

@kristoferbaxter I stepped through the code, and in the version @tharders linked, we are hitting [this](https://github.com/ampproject/worker-dom/blob/1d4762aff3591f39f2f40bf7153a95ee0a48c7a1/src/main-thread/worker.ts#L104) post, but [the handler](https://github.com/ampproject/worker-dom/blob/1d4762aff3591f39f2f40bf7153a95ee0a48c7a1/src/main-thread/install.ts#L64) is never being entered.

nainar:
Pinging @kristoferbaxter and @samouri

samouri:
I don't see any issue with `worker-dom`. Running amp.dev locally and root caused the issue to be the translation itself. We are relying upon the `type` to be success or error in order to apply a necessary class:

https://github.com/ampproject/amp.dev/blob/90b488a8271ce07edec6abda234131814889300c/pixi/src/ui/StatusIntroView.js#L27-L30

For Arabic, it seems like the `type` is set to `نجاح`. This problem exists in other languages too, for example French.

Potential fixes:
1. Rely upon the ID of the returned status as opposed to `type`
2. Don't translate the type. I don't know if it is used in translated form elsewhere.

patrickkettner:
Thanks for tracking this down, @samouri!

PR #4819, opened 2020-10-08
✨ Render amp-script hash to Pixi template

### LINKED ISSUE #4813: Pixi: amp-script hash missing
https://amp.dev/vi/page-experience/
image
image

cc @matthiasrohmer could you have a look since you setup the amp-script src?

matthiasrohmer:
Sorry, had a misconception that the AMP optimizer will add the hash, but it only does so for inline-scripts.

Fixes #4813, probably related to https://github.com/ampproject/amp.dev/issues/4818#issuecomment-705389563.

### DIFF
--- a/pixi/src/ui/page-experience.hbs
+++ b/pixi/src/ui/page-experience.hbs
@@ -2,3 +2,4 @@
 {{#htmlWebpackPlugin.files.js }}
 {% set src = '{{ . }}' %}
 {{/htmlWebpackPlugin.files.js }}
+{% set cspHash = '{{ cspHash }}' %}
--- a/frontend/templates/views/custom.j2
+++ b/frontend/templates/views/custom.j2
@@ -1,10 +1,5 @@
 {% extends '/layouts/default.j2' %}
 
-{% block head %}
-{{ super() }}
-
-{% endblock %}
-
 {% block styles %}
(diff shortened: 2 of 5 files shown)


pixi/src/ui/page-experience.hbs
```diff
@@ -2,3 +2,4 @@
 {{#htmlWebpackPlugin.files.js }}
 {% set src = '{{ . }}' %}
 {{/htmlWebpackPlugin.files.js }}
+{% set cspHash = '{{ cspHash }}' %}
```
patrickkettner, line 5: would  there ever a case for multiple scripts on the page?

patrickkettner (approved):

patrickkettner:
LGTM!

pixi/src/ui/page-experience.hbs
matthiasrohmer: In this context yes, only one hash, as Pixi's webpack config will only ever emit one bundle.

matthiasrohmer: ... and yes, rendering `src` in a loop then is confusing 😅

patrickkettner:
fwiw, it doesn't appear to be the issue in #4818

PR #4820, opened 2020-10-08
use selected pixi language for doc links

patrickkettner:
the different languages is making github go wild with spurious spacing and capitalization, but if you look at the raw diff it is all lowercase and without random whitespace


matthiasrohmer (approved):
Nice! Thanks for taking care of this.

PR #4867, opened 2020-10-16
Switch AMP Fest page to post-event state

morsssss:
Added carousels of videos, removed almost everything else.

Also removed the AMP Fest banner.

Note that we need to restore the BLM banner! I just wanted to get this PR out quickly this evening.


morsssss:
Note also that I removed the `ap-o-stage-content-subline` class... which I liked on desktop... but on mobile it makes a very tiny font.

matthiasrohmer (approved):

PR #4922, opened 2020-11-02
🐛 Update AMP URL API sample iframe URL

### LINKED ISSUE #4920: The AMP URL API example/demo is broken
### Problem
The page at https://amp.dev/documentation/examples/guides/using_the_amp_url_api/?format=websites describes how the AMP URL API can be used to request the AMP equivalent of a URL.

It contains an interactive example/demo, which is broken due to the deprecation of Go 1.9 (since ~July).

The demo should be 'repaired' if migration is feasible and straightforward, or, removed (and the associated documentation updated) if that's unfeasible.

### Location
https://amp.dev/documentation/examples/guides/using_the_amp_url_api/?format=websites

matthiasrohmer:
Fixes #4920

/cc @patrickkettner, @caroqliu: the API key is still scoped to ampbyexample.com and prevents the sample from working. I can't seem to find the matching API in the Cloud Console though. Can one of you advise?

### DIFF
--- a/examples/source/2.guides/Using_the_AMP_URL_API.html
+++ b/examples/source/2.guides/Using_the_AMP_URL_API.html
@@ -44,7 +44,7 @@ author: sebastianbenz
               sandbox="allow-scripts allow-same-origin allow-popups"
               allowfullscreen
               frameborder="0"
-              src="<% hosts.api %>/iframe/amp-url-api.html">
+              src="/static/samples/files/amp-url-api.html">
               <amp-img src="/static/samples/img/amp-url-api-placeholder.png"
                layout="fixed-height"
                height="645"
--- a/examples/static/samples/files/amp-url-api.html
+++ b/examples/static/samples/files/amp-url-api.html
@@ -69,7 +69,7 @@ function sendAmpUrlRequest() {
  */
 function sendApiRequest(urls, callback) {
   try {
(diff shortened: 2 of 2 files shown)


caroqliu:
From looking into the [AMP URL API](https://developers.google.com/amp/cache/reference/acceleratedmobilepageurl/rest) and [the sample page ](https://amp.dev/documentation/examples/guides/using_the_amp_url_api/) that discusses using it, the endpoint likely shows an iframe which takes an input and pings something like this: 
```
https://developers.google.com/amp/cache/reference/acceleratedmobilepageurl/rest/v1/ampUrls/batchGet?apix_params=%7B%22resource%22%3A%7B%22urls%22%3A%5B%22 + URL + %22%5D%7D%7D
```

It looks like this sample needs both an API key and a valid endpoint for the `amp-iframe` that uses it. I'm not seeing anything like this currently, but perhaps @sebastianbenz knows more? Did this sample used to work before #4912?

matthiasrohmer:
Sorry, I've been linking the wrong issue: #4920 is correct. The [iframe endpoint](http://amp.dev/static/samples/files/amp-url-api.html) is also there actually, it's really just the key that's not working on amp.dev 🙂

patrickkettner:
the key would exist on the old amp-by-example dash, which I never had access to. They API just wasn't enabled for amp.dev. I turned on the acceleratedmobilepageurl.googleapis.com api, I added to the PR to include the new key after confirming it works locally

patrickkettner:
since I PRed a change, could @caroqliu or @sebastianbenz approve (if approvable)

caroqliu (approved):

PR #4929, opened 2020-11-03
💬  Add VOGSY success story.

matthiasrohmer:
Bildschirmfoto 2020-11-03 um 18 36 53
localhost_8080_success-stories_vogsy_(iPad)


alexduran4:
Can we make the quote at the bottom a pull-out quote like we do on our other case studies?

alexduran4:
And can we do the same for the quote in the Solution section? The one that says "With the right information..."

PR #4958, opened 2020-11-12
🚸 Show passing recommendations if there are details

### LINKED ISSUE #4955: Pixi: Recommendations not always shown for image size, srcset, next gen formats
### Current behavior
Recommendations is not always shown for image size, use srcset, and next gen formats even when the page could have savings with smaller images, doesn't use srcset, or doesn't use next gen formats. 

I believe this is because PSI reports the page as "passing" even though the page is not adopting the recommendations.

### Expected behavior/code
If the page could have smaller image sizes, expect to see a recommendation.
If the page is not using srcset, expect to see a recommendation to use srcset. 
If the page is not using webp, expect a recommendation for webp. 

### Reproduction procedure
Visit the tool and put in a page url. For example [here](https://www.eleconomista.com.mx/amp/mercados/Peso-mexicano-se-aprecia-tras-triunfo-de-Biden-dolar-cierra-en-20.34-unidades-20201109-0078.html)

Observe there are no recommendations re: srcset, webp, smaller image size. 

Visit the [PageSpeed Insights](https://developers.google.com/speed/pagespeed/insights) tool for the same URL and observe reservations (that are still marked green/passing). 

- What browser are you using? Chrome 87
- What O/S are you using? OSX

cc @nainar  @caroqliu

matthiasrohmer:
Fixes #4955.

This force-renders recommendations for Pixi even if the underlying PSI check finishes with a score of 1 as long as it provides further details like response times, resource sizes, etc.

### DIFF
--- a/pixi/src/checks/PageExperienceCheck.js
+++ b/pixi/src/checks/PageExperienceCheck.js
@@ -97,7 +97,6 @@ export default class PageExperienceCheck {
         result.data[resultName] = !details.items.length;
         result.details[resultName] = details;
       }
-
     }
   }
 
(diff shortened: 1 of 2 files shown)


kevinkimball (approved):
LGTM

PR #4974, opened 2020-11-17
Fix invalid transformed example

schlessera:
The example for the `transformed` flag is not using a version string (`;v=1`), which seems to be invalid according to the AMP validator.


matthiasrohmer (approved):
Thanks for keeping an eye on details 🙂

PR #4956, opened 2020-11-12
Translation pt br - Finding the Right Component

alvarocamillont:

PR #4977, opened 2020-11-18
Fixed captions

morsssss:
The HTML output contained the word "code" and htmlentities. Removed markdown code markers, and replaced hyphen with something similar to avoid line break.

@matthiasrohmer , what I've done here is pretty hacky... "<amp-carousel>" was breaking at the hyphen in the Portuguese version, and when I tried an htmlentity for a nonbreaking hyphen, that was output literally. So I used a Unicode character that resembles a hyphen. This is not what we'd call a best practice :)

I wonder if we'd consider letting picture captions be wider than 300 px. That would solve this problem too.

### DIFF
--- a/pages/content/amp-dev/documentation/courses/beginning-course/finding-the-right-component.md
+++ b/pages/content/amp-dev/documentation/courses/beginning-course/finding-the-right-component.md
@@ -27,7 +27,7 @@ When we look at the [documentation](../../../documentation/components/reference/
 
 - What layouts does this component support?
 
-{{ image('/static/img/courses/beginner/image25.png', 512, 386, caption='AMP documentation page for `<amp-carousel>`.') }}
+{{ image('/static/img/courses/beginner/image25.png', 512, 386, caption='AMP documentation page for <amp‑carousel>.') }}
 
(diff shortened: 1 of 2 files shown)


morsssss:
Also, um, is there a way to re-run that Travis build?

alvarocamillont:
> /cc @alvarocamillont - I noticed this after you submitted #4956 !

Humm thats odd... Do you need me to upload a fix?

morsssss:
@alvarocamillont - not at all! It's a bug that existed before you did your translation. I just hadn't noticed it before.

This PR presents one possible fix, once we unblock the Travis build.

CrystalOnScript (approved):

morsssss:
@matthiasrohmer , I fear I need your help here. When I first pushed this, I got this error from Travis:

`This job was cancelled because the "Auto Cancellation" feature is currently enabled, and a more recent build (#15490) for branch fix/captions came in while this job was waiting to be processed.`

I just pushed an empty commit to re-trigger the build. Sadly I got the same result 😭 😭 😭

morsssss:
And... now the build has passed! I don't know why. But I am grateful.

PR #5000, opened 2020-11-28
Use the `python2` command to run Python 2.7 when it's available

TomerAberbach:
On Arch Linux (which I use) Python 2.7 is installed as [`python2`](https://wiki.archlinux.org/index.php/python#Python_2) while Python 3 is installed as [`python`](https://wiki.archlinux.org/index.php/python#Python_3). As a result, the
`test:grow-extensions` npm script tries to use Python 3 on my machine and fails. As a quick fix I've edited the script locally to use `python2`, but I have to be careful not to commit this file.

This PR makes it so that the `test:grow-extensions` npm script uses the `python2` command if it's available, but uses the `python` command otherwise.


patrickkettner:
the script seems fine, however my instinct would just be to use `python2` directly - I can't recall a system used in the last decade that didn't include it. its certainly on gcloud, macos, and the default python installs on debian and windows. I would love @matthiasrohmer or @sebastianbenz's input on this, as I haven't touched our python infra much

matthiasrohmer:
I had a brief look yesterday morning but couldn't make up my mind and hence did not commented but I was thinking in the same direction as Patrick. Let's switch to `python2` as a command name. @TomerAberbach, do you want to take this on?

TomerAberbach:
Thanks for the feedback! Yup, I can take this on

patrickkettner (approved):

Issue #4999, opened 2020-11-28
Use of non-code quality quotation marks in code example


ayumi-cloud:
Code example location in doc's:

https://amp.dev/documentation/components/amp-consent/#prompt-ui-for-stories

Current code:

```html
<amp-consent layout="nodisplay" id="consent-element">
  <script type="application/json">
    {
      "checkConsentHref": "https://foo.com/api/show-consent",
      "promptUI": "consent-ui"
    }
  </script>
  <amp-story-consent id="consent-ui" layout="nodisplay">
    <script type="application/json">
      {
        “title”: "My title",
        “message”: "My example message.",
        “vendors”: ["Item 1", "Item 2", "Item 3", "Item 4"]
      }
    </script>
  </amp-story-consent>
</amp-consent>
```

Correct code:

```html
<amp-consent layout="nodisplay" id="consent-element">
  <script type="application/json">
    {
      "checkConsentHref": "https://foo.com/api/show-consent",
      "promptUI": "consent-ui"
    }
  </script>
  <amp-story-consent id="consent-ui" layout="nodisplay">
    <script type="application/json">
      {
        "title": "My title",
        "message": "My example message.",
        "vendors": ["Item 1", "Item 2", "Item 3", "Item 4"]
      }
    </script>
  </amp-story-consent>
</amp-consent>
```

If you missed it - see screenshot:

![image](https://user-images.githubusercontent.com/57409060/100518818-e78e2600-318b-11eb-9ce4-969e8d634273.png)

patrickkettner:
thanks for the report! this is actually part of the imported documentation, whose data exists over in ampproject/amphtml. I submitted a PR over there to fix this.

ayumi-cloud:
Thank you, I did search both repos.

patrickkettner:
thanks! to be clear, I wasn't chastising - more apologizing for the unusually complex structure

ayumi-cloud:
No need to apologize! Thanks for your quick help.

PR #5001, opened 2020-11-30
force docs to rewrite curly quotes to straight

patrickkettner:
re: #4999 
not confident that this is the right place to do the transformation


matthiasrohmer (approved):
Yep, this does the trick for imported documents. We should probably `grep` over the docs inside this repository to make sure we are in good shape here.

patrickkettner:
> Yep, this does the trick for imported documents. We should probably `grep` over the docs inside this repository to make sure we are in good shape here.

done in https://github.com/ampproject/amphtml/pull/31352

PR #5002, opened 2020-11-30
remove unused package

patrickkettner:
package was added in a4fc6ce2d4046581005385ba9f4e855c8d030401, but was never actually used.


matthiasrohmer (approved):
Thanks, Patrick! 🙂

PR #5023, opened 2020-12-04
strip trailing slashes for go links

### LINKED ISSUE #3859: go.amp.dev links don't support trailing slash
## 🚀 Feature Request

I am a fan of go.amp.dev links! URLs like this work great:

[http://go.amp.dev/learn-script](http://go.amp.dev/learn-script)

However, if you happen to add a trailing slash, as people sometimes do, you get a 404:

[http://go.amp.dev/learn-script/](http://go.amp.dev/learn-script/)

If it's simple to just strip any trailing slash before rewriting the URL, that would be wonderlicious!

patrickkettner:
fixes #3859

### DIFF
--- a/platform/lib/routers/go.js
+++ b/platform/lib/routers/go.js
@@ -81,6 +81,7 @@ function initGoLinks(config) {
       });
     } else {
       simple[key] = value.url;
+      simple[`${key}/`] = value.url;
     }
   }
   return {
--- a/platform/lib/routers/go.js
+++ b/platform/lib/routers/go.js
@@ -32,8 +32,8 @@ const go = express.Router();
 const goLinks = initGoLinks(yaml.safeLoad(readFileSync(GO_LINKS_DEFINITION)));
 
 go.use((request, response, next) => {
+  const requestPath = request.path.replace(/\/?$/, '');
   let target;
-  let requestPath = request.path.replace(/\/?$/, '');
 
   if (goLinks.simple[requestPath]) {
     target = goLinks.simple[requestPath];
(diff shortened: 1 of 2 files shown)


sebastianbenz:
Can you please add a test in go.test.js.

patrickkettner:
jawoll captain

platform/lib/routers/go.js
```diff
@@ -81,6 +81,7 @@ function initGoLinks(config) {
       });
     } else {
       simple[key] = value.url;
+      simple[`${key}/`] = value.url;
```
sebastianbenz, line 84: what if the key already has a trailing slash?

patrickkettner: These are static strings, it would be `/foo/` and `/foo//`. Handling it felt like the complexity outweighed the use since there are no uses of it currently. It would allow for `/foo/` and `/foo//`. 

That being said, I do something like

```js 
} else {
  key =  key.replace(/\/?$/, '/?');
  simple[key] = value.url;
}
```

which would force an optional trailing slash (e.g. `/?`) on all `simple` routes

sebastianbenz: I think it'd be good to consistently support both schemes with and without trailing slash. Stripping trailing slashes when loading the input and when handling a request (like in your first version) might be the easiest way to do this.

patrickkettner: > I think it'd be good to consistently support both schemes with and without trailing slash.

makes sense

pushed an update, wdyt

sebastianbenz (approved):
Looks good once build passes. Thanks!

Issue #3553, opened 2020-02-17
Make git util lazy and use sh over execSync


matthiasrohmer:
Isolated from #3330.

https://github.com/ampproject/amp.dev/blob/551ee7fb30b9778cc68c3f70319ac7d876192258/platform/lib/utils/git.js#L18-L26

git.js currently does three `execSync` operations as soon as it is required. That's bad. Especially because not all of the computed values are needed for development or are even available. 

Therefore we should refactor the util to be Promise-based, cache the computed values where it makes sense and use our `sh` util instead of `execSync` directly.

sebastianbenz:
Good find!

>

patrickkettner:
closing in favor of #5125

PR #5035, opened 2020-12-07
Feature/async git helper

bezoerb:
Solves #3553


gulpfile.js/build.js
```diff
@@ -529,18 +529,22 @@ function collectStatics(done) {
...
-    'by': git.user,
-    'environment': config.environment,
-    'commit': {
-      'sha': git.version,
-      'message': git.message,
-    },
-  };
-
-  fs.writeFile(project.paths.BUILD_INFO, yaml.safeDump(buildInfo), done);
+  Promise.all([git.user(), git.version(), git.message()])
```
patrickkettner, line 532: we are using async/await elsewhere in the file - any reason you aren't here?

gulpfile.js/deploy.js
```diff
@@ -35,7 +35,7 @@ const TIMESTAMP = Date.now();
 // We tag docker images by the current git commit SHA + TIMESTAMP,
 // this makes it easy to identify and reproduce builds.
 // Pass a specific tag via commandline using `gulp updateStart --tag ABCDE...`
-const TAG = argv.tag || `${require('@lib/utils/git').version}-${TIMESTAMP}`;
+const TAG = argv.tag || `${require('@lib/utils/git').version()}-${TIMESTAMP}`;
```
patrickkettner, line 38: this would resolve as a `string || Promise`, no?

platform/lib/utils/sh.js
```diff
@@ -44,7 +44,9 @@ function sh(commandLine, ...options) {
   const command = fragments[0];
   const args = fragments.splice(1);
 
-  console.log(`$ ${command} ${args.join(' ')}`);
+  if (!opts.quiet) {
+    console.log(`$ ${command} ${args.join(' ')}`);
```
patrickkettner, line 48: would it make sense to use the [util logger](https://github.com/ampproject/amp.dev/blob/future/platform/lib/utils/log.js)?

gulpfile.js/build.js
bezoerb: Didn't want to change the the function declaration ;)
I'll give it a try

gulpfile.js/deploy.js
bezoerb: good catch :)

platform/lib/utils/sh.js
bezoerb: The logger util is used to provide scoped logs. I would keep the `console.log` statements as we don't have a specific scope here

gulpfile.js/build.js
```diff
@@ -528,23 +535,23 @@ function collectStatics(done) {
...
-      fs.writeFile(project.paths.BUILD_INFO, yaml.safeDump(buildInfo), done);
-    })
-    .catch((error) => done(error));
+async function persistBuildInfo() {
+  const version = await git.version();
+  const user = await git.user();
+  const message = await git.message();
+
+  const buildInfo = {
+    'number': travis.build.number || null,
```
patrickkettner, line 544: when would `travis.build.number` not be defined? If it isn't, are we sure `travis.build` exists?

patrickkettner: ping @bezoerb

matthiasrohmer: Yep, `travis.build` (and even `.number`) will always exist though may be undefined:

https://github.com/ampproject/amp.dev/blob/4f010c325688cf2c1037f2e56d1b0d3ed159fd4e/platform/lib/utils/travis.js#L51-L54

patrickkettner (approved):

PR #5043, opened 2020-12-08
Force jest to exit when finished

bezoerb:
Adds options `--runInBand --detectOpenHandles --forceExit` to jest calls so it won't stay alive forever & preventing pushes

CleanShot 2020-12-08 at 00 08 02@2x


sebastianbenz (approved):
Looks good. Wondering about the formatting change. Do we already run prettier on JSON files?

bezoerb:
> Wondering about the formatting change. Do we already run prettier on JSON files?
Think this is related to my VS Code. But prettier on JSON files could easily be integrated in the `husky` config

PR #5047, opened 2020-12-08
build localized versions of the site on travis when PR touches localization files

### LINKED ISSUE #5038: localization PR's should trigger a CI build for that locale
when .po files are modified, that locale should be triggered for build. some bugs have crept since PRs only seem to be testing english

patrickkettner:
fixes #5038

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,7 +34,20 @@ jobs:
     - scripts/unbuffer.sh gulp buildComponentVersions
     - scripts/unbuffer.sh npm test
     script:
-    - if [ $TRAVIS_BRANCH == "future" ] && [ $TRAVIS_PULL_REQUEST == "false" ]; then scripts/unbuffer.sh gulp buildPrepare; fi
+    - |
+      if [ $TRAVIS_BRANCH == "future" ] && [ $TRAVIS_PULL_REQUEST == "false" ]; then scripts/unbuffer.sh gulp buildPrepare; fi
+      if [ $TRAVIS_PULL_REQUEST == "true" ]; then
+        # extract list of files being modified that touch markdown localized files (e.g. example@ru.md)
+        MD_LANGS=$(git diff --name-only $TRAVIS_COMMIT_RANGE | grep @ | awk -F'[@.]' '{print $2}' | uniq)
+        # extract list of files being modified that touch PO files
+        PO_LANGS=$(git diff --name-only $TRAVIS_COMMIT_RANGE | grep LC_MESSAGES | awk -F'/' '{print $3}' | uniq)
+        # combine the two lists into one unqiue list
+        LANGS=(`for L in "${MD_LANGS[@]}" "${PO_LANGS[@]}" ; do echo "$L" ; done | sort -du`)
+        # join the list of languages with a comma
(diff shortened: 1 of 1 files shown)


patrickkettner:
@sebastianbenz would you be able to take a look?

sebastianbenz:
I was hoping for @matthiasrohmer to take a look as he's the master of our travis files.

matthiasrohmer (commented):
Sorry for letting this sit for so long, Patrick! This seems to basically work - not to sure if the imported pages are handled gracefully though as `buildPrepare` is not called. Probably not:

https://github.com/ampproject/amp.dev/blob/cd5cc1e13f827dc5a95e4c6c2b7ff9694e169d9b/.travis.yml#L36-L37

Some thoughts though:

In https://github.com/ampproject/amp.dev/pull/1881#discussion_r275148066 and ff Sebastian and I agreed to let JS be the orchestrating technology. So I'd suggest to extra the bit to get the list of changed files to the git util:

https://github.com/ampproject/amp.dev/blob/cd5cc1e13f827dc5a95e4c6c2b7ff9694e169d9b/platform/lib/utils/git.js#L18

while using the sh util

https://github.com/ampproject/amp.dev/blob/cd5cc1e13f827dc5a95e4c6c2b7ff9694e169d9b/platform/lib/utils/sh.js#L97

Next bit is building the pages. Grow allows [filtering](https://grow.io/reference/podspec/#filtering) the build output. We already utilize this for the test content, see:

https://github.com/ampproject/amp.dev/blob/cd5cc1e13f827dc5a95e4c6c2b7ff9694e169d9b/gulpfile.js/build.js#L350-L359

This is where it gets a little diffuse though. We need to find an elegant way to pass the paths then filtered by the git util to the `buildPages` task through this obscure configuration method:

https://github.com/ampproject/amp.dev/blob/cd5cc1e13f827dc5a95e4c6c2b7ff9694e169d9b/platform/lib/config.js#L247-L261

This way we would only build the actually changed pages instead of the full locale which would speed things up significantly.

I know the logic for this is scattered a lot of places - let me know if I can help with any of this or if you favor a different approach at all.

patrickkettner:
@matthiasrohmer do we want to build the changed pages only, though? Since they could have touched the po files

patrickkettner:
@matthiasrohmer another questions on this

> Sebastian and I agreed to let JS be the orchestrating technology..

It seems like this is only being used in the gulp pipeline. Are you suggesting I add a gulp task that outputs the locales and then call it to get the information when I am in travis from within the travis yaml?

matthiasrohmer:
>do we want to build the changed pages only, though? Since they could have touched the po files

Oh this is fine actually. The PO-files get validated with every run of `grow run` and/or `grow build`.

matthiasrohmer:
>It seems like this is only being used in the gulp pipeline. Are you suggesting I add a gulp task that outputs the locales and then call it to get the information when I am in travis from within the travis yaml?

Yup, a gulp task is what I am looking for. This should call a method in the git util to get the changed locales and then call the grow util to actually run a build.

patrickkettner:
closing as RIP Travis usage

PR #5054, opened 2020-12-09
use data-css-strict on amp4email

### LINKED ISSUE #5053: Add data-css-strict to all email samples
This enables stricter CSS validation required for Gmail compatibility.

//cc @patrickkettner

patrickkettner:
fixes #5053

### DIFF
--- a/examples/source/1.components/amp-autocomplete.html
+++ b/examples/source/1.components/amp-autocomplete.html
@@ -1,9 +1,9 @@
 <!--
   ## Introduction
 
-  An autocomplete-enabled input field suggests completed results corresponding to the user input as they type into the input field. 
+  An autocomplete-enabled input field suggests completed results corresponding to the user input as they type into the input field.
   This feature can help the user to carry out their task more quickly.
-  
+
   Data can either be fetched from a JSON endpoint or locally from amp-state.
 -->
 <!-- -->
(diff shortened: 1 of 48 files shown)


examples/source/1.components/amp-autocomplete.html
```diff
@@ -8,7 +8,7 @@
 -->
 <!-- -->
 <!doctype html>
-<html ⚡ lang="en">
+<html ⚡ lang="en" data-css-strict>
```
sebastianbenz, line 11: This is not an email only sample, we should put this behind a template variable that will only inject the flag if the email mode is enabled.

sebastianbenz (changes requested):
Thanks for the quick fix!

examples/source/1.components/amp-autocomplete.html
patrickkettner: this was just a mistake - email has [it's own autocomplete example](https://github.com/ampproject/amp.dev/blob/use-data-css-strict/examples/source/1.components/amp-autocomplete-email.html)

sebastianbenz: Afaik, some examples are rewritten for email.

patrickkettner: I believe `amp-bind-macro` is the only example of that, I will get that one sorted shortly

sebastianbenz: Found it, we need to add the flag here as well: https://github.com/ampproject/amp.dev/blob/future/platform/lib/format-transform/formats/amp4email.js#L26

patrickkettner: beauty, thanks. I would've taken me a long while to find that.

pushed and update for that and for the test suite

sebastianbenz (approved):
Thanks!

PR #5071, opened 2020-12-11
Revert "Add CSP Headers"

sebastianbenz:
Reverts ampproject/amp.dev#5052

PR #5072, opened 2020-12-11
use CSP reporting endpoint

patrickkettner:
Turn on the [reporting](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy-Report-Only), rather than blocking, in CSP, so we can audit potential issues

https://csp-collector.appspot.com/csp/amp is listed as a the endpoint, which is the csp report collector for @ampproject, probably should use a different endpoint?

### DIFF
--- a/platform/lib/middleware/csp.js
+++ b/platform/lib/middleware/csp.js
@@ -50,7 +50,7 @@ module.exports = (req, res, next) => {
       'https://cdn.ampproject.org/rtv/',
       ...getDynamicHosts(),
     ],
-    reportUri: ['https://csp-collector.appspot.com/csp/amp'],
+    reportUri: ['/csp-report'],
   };
 
   // Allow unsafe-inline for examples
--- a/platform/lib/middleware/csp.js
+++ b/platform/lib/middleware/csp.js
@@ -1,0 +1,66 @@
+/**
+ * Copyright 2020 The AMP HTML Authors. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS-IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+'use strict';
+
+const contentSecurityPolicy = require('helmet-csp');
+const config = require('@lib/config');
+
+const getDynamicHosts = () =>
+  ['playground', 'preview', 'go', 'log']
+    .map((key) => {
+      const {[key]: hostConfig} = config.hosts || {};
+      if (hostConfig) {
+        return `${config.getHost(hostConfig)}/`;
+      }
+    })
+    .filter((v) => v);
+
+module.exports = (req, res, next) => {
+  const directives = {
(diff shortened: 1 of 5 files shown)


patrickkettner:
@sebastianbenz this just logs out to [CSP Report] in the cloud log explorer. We would change [this](https://github.com/ampproject/amp.dev/pull/5072/files#diff-5e1f4ebf545d527c05556c742255a2c4dc54fa31e6167579d8cd6a47552c302fR64) line to false after N days once we have confirmed that there aren't any issues with CSP being hit in production

platform/lib/middleware/csp.js
```diff
@@ -0,0 +1,66 @@
...
+      ...getDynamicHosts(),
+    ],
+    reportUri: ['https://csp-collector.appspot.com/csp/amp'],
+  };
+
+  // Allow unsafe-inline for examples
+  if (/\/documentation\/examples\//.test(req.path)) {
+    directives.scriptSrc = [...directives.scriptSrc, `'unsafe-inline'`];
+  }
+
```
bezoerb, line 60: We should add the `unsafe-eval` directive for playground to fix #5071. Maybe something like this:
```js
const {playground} = config.hosts || {};
const {host: playgroundHost} = playground || {};
if (req.hostname.includes(playgroundHost)) {
  directives.scriptSrc = [...directives.scriptSrc, `'unsafe-eval'`];
}
```

platform/lib/middleware/csp.js
```diff
@@ -0,0 +1,66 @@
...
+      'https://www.googletagmanager.com/gtag/js',
+      ...getDynamicHosts(),
+    ],
+    objectSrc: [`'none'`],
+    styleSrc: [
+      `'unsafe-inline'`,
+      'https://cdn.ampproject.org/rtv/',
+      ...getDynamicHosts(),
+    ],
+    reportUri: ['https://csp-collector.appspot.com/csp/amp'],
```
bezoerb, line 53: Shouldn't the [`/csp-report`](https://github.com/ampproject/amp.dev/pull/5072/commits/2a3158a2bbb031e7ac13254edc826940605c9531#diff-2b890bf0ec18da7e4c03e8bf4568a4c7c850411978807b0980a9338fd64a0050R27) uri be used here?

platform/lib/routers/cspReport.js
```diff
@@ -0,0 +1,31 @@
...
+const express = require('express');
+const log = require('@lib/utils/log')('CSP Violation Report')
+
+// eslint-disable-next-line new-cap
+const cspReportRouter = express.Router();
+
+cspReportRouter.use('/csp-report', express.json({ type: 'application/csp-report' }));
+
+cspReportRouter.post('/csp-report', (req, res) => {
+  log.error(req.body)
```
bezoerb, line 28: Are these logs persisted somewhere or are they only available at runtime?

platform/lib/middleware/csp.js
patrickkettner: totally - changing this to reportOnly will give us a view on domains that are hitting that and therefore we can scope anything like that accordingly

platform/lib/middleware/csp.js
patrickkettner: 🤦‍♂️

platform/lib/routers/cspReport.js
patrickkettner: They are persisted in the Google Cloud logger

platform/lib/middleware/csp.js
patrickkettner: fixed

bezoerb (approved):
LGTM

sebastianbenz (approved):
Thanks!

PR #5081, opened 2020-12-14
optimize 404 pages

### LINKED ISSUE #5080: search button throws an error on on 404 page
image

Error is thrown when you tap on the magnifying glass icon.

Strangely, _doesn't_ occur on https://amp.dev/404, but does on any other page that is a 404

patrickkettner:
probably a bad idea, but fixes #5080

### DIFF
--- a/platform/lib/routers/notFound.js
+++ b/platform/lib/routers/notFound.js
@@ -18,13 +18,21 @@
 
 const {pagePath} = require('@lib/utils/project');
 const {setMaxAge} = require('../utils/cacheHelpers.js');
+const AmpOptimizer = require('@ampproject/toolbox-optimizer');
+const {readFileSync} = require('fs');
+const {join} = require('path');
+const FourOhFour = readFileSync(join(pagePath(), '404.html'), 'utf-8');
+
+const optimizer = AmpOptimizer.create();
 
 // eslint-disable-next-line no-unused-vars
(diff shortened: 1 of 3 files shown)


platform/lib/routers/notFound.js
```diff
@@ -18,13 +18,21 @@
...
+
+const optimizer = AmpOptimizer.create();
 
 // eslint-disable-next-line no-unused-vars
 module.exports = (req, res, next) => {
   setMaxAge(res, 60 * 10); // ten minutes
   if (req.headers.accept && req.headers.accept.includes('text/html')) {
-    res.status(404).sendFile('404.html', {root: pagePath()});
-    return;
+    optimizer.transformHtml(FourOhFour).then((optimized) => {
```
sebastianbenz, line 32: few things:

1. please cache the optimized page in memory
2. Ideally, we'd be sharing the optimizer config with growPages.js
3. please use async await.

patrickkettner: thanks! I wasn't sure if optimizing the 404 page is the actual "correct" thing, as opposed to just including the `<amp-lightbox>` code in the search partial.
Will update with your requests

patrickkettner: updated

sebastianbenz (approved):
thanks!

PR #5026, opened 2020-12-04
fix invalid html in playground demos

patrickkettner:
fixes #5024

PR #5083, opened 2020-12-14
fix more invalid html in playground demos

patrickkettner:
The rest of the demos missed in #5026


sebastianbenz (approved):
Thanks for fixing!

PR #5106, opened 2020-12-16
Fix mission & vision headings on low res

alanorozco:
| Before | After |
|-|-|
|image|image|


sebastianbenz (approved):
Good one! Thanks for fixing!

alanorozco:
@sebastianbenz No problem!

I would merge this but I lack powers.

Issue #5146, opened 2020-12-24
Amp4Email: amp-autocomplete-email component not working


anbarasiu:
## 🐛 Bug Report

I'm following the documentation [here](https://amp.dev/documentation/examples/components/amp-autocomplete-email/), but the amp-autocomplete-email isn't working, even after importing the required amp-autocomplete component.

I tried to open the demo on [playground](https://playground.amp.dev/?url=https%3A%2F%2Fpreview.amp.dev%2Fdocumentation%2Fexamples%2Fcomponents%2Famp-autocomplete-email&format=email&_gl=1*ix9bcx*_ga*YW1wLWNaX0tpMkg1ZmFwWHhuOU5DbDI1SWc.) on the same doc page, which throws this error in the console. 
`Resource: failed to build: amp-autocomplete#1: <amp-autocomplete> is not currently available in AMP4Email.`

This is misleading, since the documentation states that the component is available.

caroqliu:
@sebastianbenz @matthiasrohmer Would it be possible to inject `data-amp-autocomplete-opt-in` attribute to `amp-autocomplete` elements displayed in the AMP playground in the email format? This is a partially-supported feature in email -- valid to send in emails, but off by default when viewing emails. Individual email clients can turn it on by default in their environments when they add support by injecting the above attribute at runtime, so it would be helpful if the AMP playground replicated a supported environment behavior rather than an unsupported one.

patrickkettner:
yep! PRed a change - thanks @caroqliu!!

caroqliu:
thank you!!

patrickkettner:
this was fixed by #5165

PR #5165, opened 2021-01-05
add data-amp-autocomplete-opt-in to email playground samples

patrickkettner:
fix for #5146

### DIFF
--- a/examples/source/1.components/amp-autocomplete-email.html
+++ b/examples/source/1.components/amp-autocomplete-email.html
@@ -81,8 +81,12 @@ formats:
 
     Note: Because server-side filtering is mandatory in the email format, it is wise to pair the `src` and `query` attributes
     in order to passed user inputs to a statically generated endpoint. For example if a `src="https://example.com` and `query="q"`, then a user who types in `abc` will get the fetch JSON result from `https://example.com?q=abc`.
+
+    Note: that the `data-amp-autocomplete-opt-in` attribute has
+    to be included to amp4email amp documents
     -->
     <amp-autocomplete
+      data-amp-autocomplete-opt-in
       min-characters="1"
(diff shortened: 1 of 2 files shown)


examples/source/1.components/amp-autocomplete-email.html
```diff
@@ -81,8 +81,12 @@ <h2>Basic usage</h2>
 
     Note: Because server-side filtering is mandatory in the email format, it is wise to pair the `src` and `query` attributes
     in order to passed user inputs to a statically generated endpoint. For example if a `src="https://example.com` and `query="q"`, then a user who types in `abc` will get the fetch JSON result from `https://example.com?q=abc`.
+
+    Note: that the `data-amp-autocomplete-opt-in` attribute has
+    to be included to amp4email amp documents
     -->
     <amp-autocomplete
+      data-amp-autocomplete-opt-in
```
caroqliu, line 89: Sorry, should have been more clear - using `data-amp-autocomplete-opt-in` is not valid AMP - users should not be the ones to include it.
Rather, email clients inject it into their email environment when displaying/sending the email, after the document has passed the AMP validator. Is it possible to replicate that behavior in the displayed document in the playground without it being part of the manual code sample?

patrickkettner: my mistake - I should have checked the source - thank you @caroqliu! I updated the code to add it at runtime, and added @sebastianbenz to review the method of adding it.

sebastianbenz (approved):

caroqliu (approved):

nainar:
Do we know when this will land @patrickkettner ?

patrickkettner:
@nainar the build is broken at the moment, but once its fixed ill push it

PR #5182, opened 2021-01-09
📌  Update to js-yaml 4 and replace safe prefix from yaml methods

matthiasrohmer:
Replaces #5155 following the [migration guide](https://github.com/nodeca/js-yaml/blob/master/migrate_v3_to_v4.md).


patrickkettner (approved):
thanks!

PR #5183, opened 2021-01-09
✨ Add badge to bento component docs

### LINKED ISSUE #5094: Render Bento label in component reference docs
Once `bento` has been add to the supported formats in bento component readme files (see https://github.com/ampproject/amphtml/issues/31609), we should add a bento label to the respective component doc pages. 

* For the design we should re-use the `experimental` label design (color `#f15a5a`).  
* The label should link to the Bento intro guide (to be written).

Mockup: 

Screen Shot 2020-12-15 at 19 27 27

matthiasrohmer:
Resolves #5094.

Bildschirmfoto 2021-01-09 um 16 58 48

### DIFF
--- a/frontend/templates/views/partials/component-headline.j2
+++ b/frontend/templates/views/partials/component-headline.j2
@@ -13,15 +13,26 @@
 {% endif %}
 </h1>
 
-{% if doc.experimental or doc.deprecated %}
+{% if doc.experimental or doc.deprecated or doc.bento %}
 <div class="ap--component-tags">
 {% if doc.experimental %}
   {% do doc.styles.addCssFile('css/components/atoms/pill.css') %}
-  <div class="ap-a-pill ap-a-pill-small ap-a-pill-meta">{{ _('Experimental') }}</div>
+  <div class="ap-a-pill ap-a-pill-small ap-a-pill-meta">
+    {{ _('Experimental') }}
+  </div>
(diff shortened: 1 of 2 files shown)


frontend/templates/views/partials/component-headline.j2
```diff
@@ -13,15 +13,26 @@
 {% endif %}
 </h1>
 
-{% if doc.experimental or doc.deprecated %}
+{% if doc.experimental or doc.deprecated or doc.bento %}
```
sebastianbenz, line 16: Would it make sense to add bento to the `formats` field?

sebastianbenz (approved):
Thanks Matthias!

frontend/templates/views/partials/component-headline.j2
matthiasrohmer: Could make sense. Is there a specific plan how bento component docs will be structured? Or do they have API parity with the classic components and therefore don't need their own docs but really just a marker?

Wasn't comfortable making assumptions about this and therefore simply added the flag 😁

sebastianbenz: Bento and AMP component docs are going to be the same (Bento and AMP version of a component is the same JS script). So a flag is fine.

sebastianbenz (approved):

PR #5192, opened 2021-01-11
Extract available versions from filesystem

sebastianbenz:
This is in preparation for the bento doc integration. Bento versions are
not yet in the validation rules requiring us to parse the versions from
the file system. This PR also makes sure that all bento component docs
are always fetched from master.


platform/lib/pipeline/componentReferenceImporter.js
```diff
@@ -224,11 +223,21 @@ class ComponentReferenceImporter {
...
+    for (const file of extension.files) {
+      const path = file.substring(`extensions/${spec.name}/`.length);
+      const match = path.match(/^(\d+\.\d+)\//);
+      if (match) {
+        versions.add(match[1]);
+      }
+    }
+    spec.version = Array.from(versions);
+
+    // Calculate the latest version
```
caroqliu, line 237: There is a concept of a "latest version" of a component that is specified here in `bundles.config.js`. amp-carousel is an example of when the `latestVersion` specified is not always the larger: https://github.com/ampproject/amphtml/blob/91aaf6f3e28c28117fc6ca0eb660d39161e387ee/build-system/compile/bundles.config.js#L460-L464

Is this relevant at all here?

caroqliu (approved):

platform/lib/pipeline/componentReferenceImporter.js
sebastianbenz: Thanks! Afaik, we use the experimental flag instead. But we should use this information instead of auto calculating the latest version.

//cc @matthiasrohmer

matthiasrohmer: Created #5200 to track this, thanks for the hint Caroline!

platform/config/imports/componentReference.json
```diff
@@ -1,4 +1,31 @@
 {
 	"only": [],
-	"fetchFromMaster": ["amp-youtube", "amp-access-laterpay", "amp-app-banner", "amp-script", "amp-carousel", "amp-consent", "amp-bind", "amp-story", "amp-subscriptions", "amp-list", "amp-script"]
+	"fetchFromMaster": [
```
matthiasrohmer, line 3: Is somebody already taking care of adding `bento: true` to their frontmatter in ampproject/amphtml?

matthiasrohmer (approved):

platform/config/imports/componentReference.json
sebastianbenz: yes, all bento docs will be updated by the team

PR #5198, opened 2021-01-12
Added Bento guide

CrystalOnScript:
Bento guide is here! We still need:
- Link to blog
- Link to amp-stream-gallery
- Possibly a larger code sample// @sebastianbenz @dvoytenko 

Thanks all for your patience and help! 

screencapture-localhost-8080-documentation-guides-and-tutorials-start-bento-guide-2021-01-11-17_12_56


pages/content/amp-dev/documentation/guides-and-tutorials/start/bento_guide.md
```diff
@@ -0,0 +1,102 @@
...
+$order: 1
+description: 'Use a subset of AMP components as standalone elements in non-AMP pages with Bento AMP.'
+tutorial: false
+toc: true
+formats:
+  - websites
+author: CrystalOnScript
+---
+
+Use a subset of AMP components as standalone elements in non-AMP pages with Bento AMP. Bento AMP is a project that embraces use of AMP components without the need to commit to fully valid AMP. This creates copy and past-able code for common UI widgets, allows testing if AMP is the right switch for your website, and using Bento in a migration plan.
```
patrickkettner, line 12: Are we framing bento as a stepping stone to AMP?

pages/content/amp-dev/documentation/guides-and-tutorials/start/bento_guide.md
```diff
@@ -0,0 +1,102 @@
...
+
+Use a subset of AMP components as standalone elements in non-AMP pages with Bento AMP. Bento AMP is a project that embraces use of AMP components without the need to commit to fully valid AMP. This creates copy and past-able code for common UI widgets, allows testing if AMP is the right switch for your website, and using Bento in a migration plan. 
+
+[tip type="important"]
+ Bento is only available [experimentally](../learn/experimental.md). Read our [Blog Post](TODO) for information on future plans and how to get involved! 
+[/tip]
+
+# Use Bento components in non-AMP pages
+
+Bento components are available for experimental use in the developer preview. After enabling the Bento experimental
```
patrickkettner, line 20: After enabling the Bento experiment<b>al</b>?

pages/content/amp-dev/documentation/guides-and-tutorials/start/bento_guide.md
```diff
@@ -0,0 +1,102 @@
...
+[/tip]
+
+# Use Bento components in non-AMP pages
+
+Bento components are available for experimental use in the developer preview. After enabling the Bento experimental 
+
+
+## Enable Bento Experiment 
+
+Bento components are available experimentally and you must [enable experimental features](../learn/experimental.md).
```
patrickkettner, line 25: is this not repeating the previous sentence?

pages/content/amp-dev/documentation/guides-and-tutorials/start/bento_guide.md
```diff
@@ -0,0 +1,102 @@
...
+```
+
+
+
+## Import AMP runtime and Bento component logic
+
+You must include the AMP runtime script and import the logic for each individual Bento component desired.
+
+```
+<script async src="https://cdn.ampproject.org/v0.js">&lt;/script>
```
patrickkettner, line 57: is the &amp;lt; intentional?

pages/content/amp-dev/documentation/guides-and-tutorials/start/bento_guide.md
```diff
@@ -0,0 +1,102 @@
...
+```
+<script>
+  (self.AMP = self.AMP || []).push(function (AMP) {
+    AMP.toggleExperiment('bento', true);
+  });
+</script>
+```
+
+
+Or enable a single Bento component by including the script below. Ensure you replace `bento-foo` with the desired supported component name given on its documentation page:
```
patrickkettner, line 39: would it make sense to link "supported component" to the "Available Bento components" section?

pages/content/amp-dev/documentation/guides-and-tutorials/start/bento_guide.md
CrystalOnScript: Removed, thanks!

pages/content/amp-dev/documentation/guides-and-tutorials/start/bento_guide.md
```diff
@@ -0,0 +1,102 @@
...
+*   [amp-social-share](../../../documentation/components/reference/amp-social-share.md)
+*   [amp-timeago](../../../documentation/components/reference/amp-timeago.md)
+*   Video components
+    *   [amp-video](../../../documentation/components/reference/amp-video.md)
+    *   [amp-youtube](../../../documentation/components/reference/amp-youtube.md)
+    *   [amp-video-iframe](../../../documentation/components/reference/amp-video-iframe.md)
+
+# Working with experiments
+
+Bento AMP implementation is still in experimental mode and available through the developer preview. The AMP team welcomes developer feedback through [GitHub](https://github.com/ampproject/amphtml/issues) and our [Slack](https://docs.google.com/forms/d/e/1FAIpQLSd83J2IZA6cdR6jPwABGsJE8YL4pkypAbKMGgUZZriU7Qu6Tg/viewform?fbzx=4406980310789882877) channel. Please reach out with any questions or issues.
```
patrickkettner, line 92: Is "Bento AMP implementation" the correct phrasing/ordering?

pages/content/amp-dev/documentation/guides-and-tutorials/start/bento_guide.md
CrystalOnScript: It is, sort of - users tend to skip banners of things unintentionally, so repeating that it's experimental seems okay for now.

patrickkettner: SGTM!

pages/content/amp-dev/documentation/guides-and-tutorials/start/bento_guide.md
CrystalOnScript: nope, nice catch again - something must have gone wrong in my doc conversion

pages/content/amp-dev/documentation/guides-and-tutorials/start/bento_guide.md
CrystalOnScript: good suggestion - done!

pages/content/amp-dev/documentation/guides-and-tutorials/start/bento_guide.md
CrystalOnScript: Last I heard it was a possible path - cc @caroqliu @dvoytenko

sebastianbenz: I think we should frame bento bigger. My (badly phrased) suggestion for this paragraph would be:

Use AMP components everywhere with Bento AMP. Bento AMP allows you to use AMP components in non-AMP pages without the need to commit to fully valid AMP. For example. you can use AMP's well-tested and accessible carousel implementation with frameworks or CMSes that don't support AMP.

pages/content/amp-dev/documentation/guides-and-tutorials/start/bento_guide.md
```diff
@@ -0,0 +1,102 @@
...
+
+Use a subset of AMP components as standalone elements in non-AMP pages with Bento AMP. Bento AMP is a project that embraces use of AMP components without the need to commit to fully valid AMP. This creates copy and past-able code for common UI widgets, allows testing if AMP is the right switch for your website, and using Bento in a migration plan. 
+
+[tip type="important"]
+ Bento is only available [experimentally](../learn/experimental.md). Read our [Blog Post](TODO) for information on future plans and how to get involved! 
+[/tip]
+
+# Use Bento components in non-AMP pages
+
+Bento components are available for experimental use in the developer preview.
```
sebastianbenz, line 20: Can you elaborate on what this means? Is it OK to use these in production? What are the risks? 

//cc @dvoytenko @caroqliu

pages/content/amp-dev/documentation/guides-and-tutorials/start/bento_guide.md
```diff
@@ -0,0 +1,102 @@
...
+
+
+## Import AMP runtime and Bento component logic
+
+You must include the AMP runtime script and import the logic for each individual Bento component desired.
+
+```
+<script async src="https://cdn.ampproject.org/v0.js"></script>
+
+<script async custom-element="amp-bento" src="https://cdn.ampproject.org/v0/amp-bento-1.0.js"></script>
```
sebastianbenz, line 59: suggestion, rephrase `amp-bento` to `amp-bento-component-name` otherwise it looks like a generic bento runtime.

pages/content/amp-dev/documentation/guides-and-tutorials/start/bento_guide.md
caroqliu: Bento _can be_ a stepping stone to AMP, but it should be valuable to anyone regardless of their investment level in AMP. @nainar likely has good copy for how to frame this.

pages/content/amp-dev/documentation/guides-and-tutorials/start/bento_guide.md
caroqliu: They are guarded behind an experimental flag which can be enabled on the document or in the console, and this is no different from other [experiments in AMP](https://amp.dev/documentation/guides-and-tutorials/learn/experimental/).

The risks are also the same - it is experimental because we reserve the right for the API to change drastically if given feedback. This is not likely IMO but the purpose of the preview is to get feedback that can include making breaking changes if it makes the features better.

pages/content/amp-dev/documentation/guides-and-tutorials/start/bento_guide.md
```diff
@@ -0,0 +1,102 @@
...
+Bento supported AMP components are listed below:
+
+
+
+*   [amp-accordion](../../../documentation/components/reference/amp-accordion.md)
+*   [amp-base-carousel](../../../documentation/components/reference/amp-base-carousel.md)
+*   [amp-inline-gallery](../../../documentation/components/reference/amp-inline-gallery.md)
+    *   amp-inline-gallery-pagination
+    *   amp-inline-gallery-thumbnails
+*   [amp-stream-gallery](TODO)
```
caroqliu, line 76: Note that this is in AMPHTML: https://github.com/ampproject/amphtml/blob/master/extensions/amp-stream-gallery/amp-stream-gallery.md

caroqliu (approved):

pages/content/amp-dev/documentation/guides-and-tutorials/start/bento_guide.md
nainar: Agree that a stepping stone is one of possible use cases. It's perfectly fine for people to stop at Bento AMP. 

I like Sebastian's phrasing, also from the blog:

`Developers can use Bento components for one-off cases such as adding a carousel to an otherwise non AMP page, or as a way of testing out AMP components on the path to being valid AMP. `

pages/content/amp-dev/documentation/guides-and-tutorials/start/bento_guide.md
```diff
@@ -0,0 +1,102 @@
...
+*   [amp-fit-text](../../../documentation/components/reference/amp-fit-text.md)
+*   [amp-instagram](../../../documentation/components/reference/amp-instagram.md)
+*   [amp-lightbox](../../../documentation/components/reference/amp-lightbox.md)
+*   [amp-selector](../../../documentation/components/reference/amp-selector.md)
+*   [amp-social-share](../../../documentation/components/reference/amp-social-share.md)
+*   [amp-timeago](../../../documentation/components/reference/amp-timeago.md)
+*   Video components
+    *   [amp-video](../../../documentation/components/reference/amp-video.md)
+    *   [amp-youtube](../../../documentation/components/reference/amp-youtube.md)
+    *   [amp-video-iframe](../../../documentation/components/reference/amp-video-iframe.md)
```
caroqliu, line 88: Please remove, I don't see a `1.0` registered in AMPHTML for `amp-video-iframe`.

pages/content/amp-dev/documentation/guides-and-tutorials/start/bento_guide.md
CrystalOnScript: got it, thank you!

pages/content/amp-dev/documentation/guides-and-tutorials/start/bento_guide.md
CrystalOnScript: updated the wording, lmkwyt!

pages/content/amp-dev/documentation/guides-and-tutorials/start/bento_guide.md
CrystalOnScript: updated the intro, would love to hear what ya'll think!

pages/content/amp-dev/documentation/guides-and-tutorials/start/bento_guide.md
```diff
@@ -0,0 +1,388 @@
...
+<html ⚡>
+  <head>
+    <meta charset="utf-8" />
+    <title>AMP Public Radio</title>
+    <link rel="canonical" href="self.html" />
+    <meta
+      name="viewport"
+      content="width=device-width,minimum-scale=1,initial-scale=1"
+    />
+    <style amp-boilerplate>
```
caroqliu, line 108: Mind shrinking the boilerplate here? Looks like prettier might have expanded it.

pages/content/amp-dev/documentation/guides-and-tutorials/start/bento_guide.md
```diff
@@ -0,0 +1,388 @@
...
+
+
+## Page experience
+
+Bento components are designed to be highly performant and contribute to an excellent page experience. Developers are highly encouraged to [file issues](https://github.com/ampproject/amphtml/issues) if they see contradictory results.  
+
+
+## AMP caches and validation
+
+Using Bento components during the limited developer preview makes your web page an invalid AMP page. Therefore, your doc and any Bento components are not added to any AMP caches.
```
caroqliu, line 388: Are these intended to be subheadings of `Working with experiments`?

pages/content/amp-dev/documentation/guides-and-tutorials/start/bento_guide.md
```diff
@@ -0,0 +1,388 @@
...
+    ></script>
+    <script
+      async
+      custom-element="amp-fit-text"
+      src="https://cdn.ampproject.org/v0/amp-fit-text-1.0.js"
+    ></script>
+    <script>
+      (self.AMP = self.AMP || []).push(function(AMP) {
+        AMP.toggleExperiment("amp-base-carousel-bento", true);
+        AMP.toggleExperiment("amp-fit-text-bento", true);
```
caroqliu, line 180:
```suggestion
        AMP.toggleExperiment("bento", true);
```

caroqliu: Otherwise:
```suggestion
        AMP.toggleExperiment("bento-carousel", true);
        AMP.toggleExperiment("bento-fit-text", true);
```

caroqliu: Sorry, we updated these experiment names since I made this sample 😇

pages/content/amp-dev/documentation/guides-and-tutorials/start/bento_guide.md
CrystalOnScript: I was thinking of removing it - is that still valid for the experiment, to just have the runtime script?

pages/content/amp-dev/documentation/guides-and-tutorials/start/bento_guide.md
CrystalOnScript: good catch - fixed!

pages/content/amp-dev/documentation/guides-and-tutorials/start/bento_guide.md
caroqliu: Yes, that works!

sebastianbenz: Is the boilerplate needed with bento?

pages/content/amp-dev/documentation/guides-and-tutorials/start/bento_guide.md
sebastianbenz: Thanks Caroline. Might be worth mentioning in the guide.

pages/content/amp-dev/documentation/guides-and-tutorials/start/bento_guide.md
CrystalOnScript: No worries, thanks for catching!

pages/content/amp-dev/documentation/guides-and-tutorials/start/bento_guide.md
```diff
@@ -0,0 +1,343 @@
...
+
+[tip type="important"]
+ Bento is only available [experimentally](../learn/experimental.md). Read our [Blog Post](TODO) for information on future plans and how to get involved! 
+[/tip]
+
+# Use Bento components in non-AMP pages
+
+Bento components are available for experimental use in the developer preview. This means you must enable the experimental flag either in the document or the console. This is no different from other [experiments in AMP](../learn/experimental.md). 
+
+The risk of using the Bento experiment is also the same. The AMP teams reserves the right to change the API drastically based on feedback. Changes are necessary to improve features.
```
morsssss, line 24: I wouldn't talk about "the risk" or "drastically". I think this paragraph should be replaced by a sentence that just says that, while Bento's in developer preview, APIs may change. Otherwise it kind of sounds like we could keep changing the API indefinitely.

pages/content/amp-dev/documentation/guides-and-tutorials/start/bento_guide.md
```diff
@@ -0,0 +1,343 @@
...
+    </div>
+  </body>
+</html>
+```
+
+See the [demo in action on Glitch](TODO).
+
+# Managing events
+
+Fully valid AMP installs event listeners on elements via the [`on` attribute](../learn/amp-actions-and-events.md) with the event and responding action as values. Bento AMP does not adhere to this rules. Instead, the AMP team will provide a JS API to manage and react to events.
```
morsssss, line 315: note "this rules"

I don't quite understand this paragraph as written. I think you're saying that valid AMP supports actions and events via the "on" attribute, but Bento AMP doesn't? And that soon we'll provide a special API for this in Bento?

pages/content/amp-dev/documentation/guides-and-tutorials/start/bento_guide.md
```diff
@@ -0,0 +1,343 @@
...
+
+
+# Page experience
+
+Bento components are designed to be highly performant and contribute to an excellent page experience. Developers are highly encouraged to [file issues](https://github.com/ampproject/amphtml/issues) if they see contradictory results.  
+
+
+# AMP caches and validation
+
+Using Bento components during the limited developer preview makes your web page an invalid AMP page. Therefore, your doc and any Bento components are not added to any AMP caches.
```
morsssss, line 343: "are not added to" - I'd say instead "won't be served by"

morsssss (changes requested):
Nice! Just a few comments........ a few....

pages/content/amp-dev/documentation/guides-and-tutorials/start/bento_guide.md
```diff
@@ -0,0 +1,343 @@
...
+    AMP.toggleExperiment('bento-foo', true);
+  });
+</script>
+```
+
+
+
+## Import AMP runtime and Bento component logic
+
+You must include the AMP runtime script and import the logic for each individual Bento component desired.
```
dvoytenko, line 58: Might be good to do an aside that AMP team is planning to eventually eliminate the need for v0.js.

pages/content/amp-dev/documentation/guides-and-tutorials/start/bento_guide.md
dvoytenko: I think we should just tend to advise to use a single "bento" experiment in all guides. There's another reason for it: some of the runtime is geared to this experiment a bit better.

pages/content/amp-dev/documentation/guides-and-tutorials/start/bento_guide.md
CrystalOnScript: Good call - thanks!

pages/content/amp-dev/documentation/guides-and-tutorials/start/bento_guide.md
```diff
@@ -0,0 +1,343 @@
...
+$order: 1
+description: 'Use a subset of AMP components as standalone elements in non-AMP pages with Bento AMP.'
+tutorial: false
+toc: true
+formats:
+  - websites
+author: CrystalOnScript
+---
+
+Use AMP's well-tested, cross browser compatible and accessible components everywhere with Bento AMP. Bento AMP allows you to use AMP components in non-AMP pages without needing to commit to fully valid AMP! Bento allows you to take components and place them in implementations with frameworks and CMSs that don't support AMP.
```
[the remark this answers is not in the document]
CrystalOnScript: I've added the hyphen, but I like the list of benefits up top for people who may not be familiar with AMP components and why they're great

pages/content/amp-dev/documentation/guides-and-tutorials/start/bento_guide.md
```diff
@@ -63,6 +49,10 @@ You must include the AMP runtime script and import the logic for each individual
 <script async custom-element="amp-bento-component-name" src="https://cdn.ampproject.org/v0/amp-bento-1.0.js"></script>
 ``` 
 
+[tip type="note"]
+Eliminating the need for the AMP runtime script is a high priority for Bento AMP. Follow this progress on the [Bento roadmap](../../../../community/roadmap/?category=bento#status-updates).
```
caroqliu, line 53: Nice!

caroqliu (approved):

dvoytenko (approved):

pages/content/amp-dev/documentation/guides-and-tutorials/start/bento_guide.md
```diff
@@ -0,0 +1,332 @@
...
+$order: 1
+description: 'Use a subset of AMP components as standalone elements in non-AMP pages with Bento AMP.'
+tutorial: false
+toc: true
+formats:
+  - websites
+author: CrystalOnScript
+---
+
+Use AMP's well-tested, cross-browser compatible and accessible components anywhere on the web with Bento AMP. Bento AMP allows you to use AMP components in non-AMP pages without needing to commit to fully valid AMP! Bento allows you to take components and place them in implementations with frameworks and CMSs that don't support AMP.
```
sebastianbenz, line 12: Not sure about this, but can we avoid starting two subsequent sentences with the same "Bento allows you to ..."?

pages/content/amp-dev/documentation/guides-and-tutorials/start/bento_guide.md
```diff
@@ -0,0 +1,332 @@
...
+
+Bento AMP enables you to insert highly performant and tested web components, such as carousels and media embeds, anywhere you need. Developers can use Bento components for one-off cases, such as adding a carousel to an otherwise non-AMP page. Bento lets you test out AMP components to see if the path to valid AMP is right for your use case.
+
+[tip type="important"]
+ Bento is only available [experimentally](../learn/experimental.md). Read our [Blog Post](TODO) for information on future plans and how to get involved! 
+[/tip]
+
+# Use Bento components in non-AMP pages
+
+Bento components are available for experimental use in the developer preview. This means you must enable the experimental flag either in the document or the console. This is no different from other [experiments in AMP](../learn/experimental.md).
```
sebastianbenz, line 22: Suggestion" start with "Bento components are still in development and APIs might change. However, they're available for experimental use in a developer preview".

pages/content/amp-dev/documentation/guides-and-tutorials/start/bento_guide.md
```diff
@@ -0,0 +1,332 @@
...
+Read each Bento component’s reference documentation for implementation details.
+
+
+# Available Bento components  <a name="available-bento-components"></a>
+
+Bento supported AMP components are listed below:
+
+
+
+*   [amp-accordion](../../../documentation/components/reference/amp-accordion.md)
```
sebastianbenz, line 65: These currently link to the current prod version, but not the bento version:

* prod: https://amp.dev/documentation/components/amp-fit-text/?format=websites
* bento: https://amp.dev/documentation/components/amp-fit-test-v1.0/?format=websites

pages/content/amp-dev/documentation/guides-and-tutorials/start/bento_guide.md
```diff
@@ -0,0 +1,332 @@
...
+      async
+      custom-element="amp-fit-text"
+      src="https://cdn.ampproject.org/v0/amp-fit-text-1.0.js"
+    ></script>
+    <script>
+      (self.AMP = self.AMP || []).push(function (AMP) {
+        AMP.toggleExperiment('bento', true);
+      });
+    </script>
+    <style>
```
sebastianbenz, line 109: This demo is super long. Can we strip it to the bare minimum? No CSS etc.

pages/content/amp-dev/documentation/guides-and-tutorials/start/bento_guide.md
```diff
@@ -0,0 +1,332 @@
...
+                Maureen Corrigan's 10 Books That Will Connect You In A Socially
+                Distant Year
+              </amp-fit-text>
+            </a>
+          </div>
+          <img
+            class="article-img"
+            width="700"
+            height="393"
+            src="https://media.npr.org/assets/img/2020/12/03/corrigan2020_wide-26e0b93fd657db854e1ba2bdb7b18539ca72ecfa-s1100-c85.jpg"
```
sebastianbenz, line 291: Avoid linking to external assets. These might change or get deleted.

pages/content/amp-dev/documentation/guides-and-tutorials/start/bento_guide.md
```diff
@@ -0,0 +1,332 @@
...
+        </article>
+      </amp-base-carousel>
+    </div>
+  </body>
+</html>
+```
+
+See the [demo in action on Glitch](TODO).
+
+# Managing events
```
sebastianbenz, line 302: I find this heading a bit confusing as it doesn't really explain what this section is about. Maybe: "Interacting with Bento components. 

This section should also move into the "Use Bento components in non-AMP pages" section.

pages/content/amp-dev/documentation/guides-and-tutorials/start/bento_guide.md
```diff
@@ -0,0 +1,332 @@
...
+
+Fully valid AMP installs event listeners on elements via the [`on` attribute](../learn/amp-actions-and-events.md) with the event and responding action as values. Bento AMP does not rely on this attribute. Instead, the AMP team will provide a JS API to manage and react to events. 
+
+During the developer preview, you can register component events and respond with defined actions using following syntax:
```
sebastianbenz, line 306: I find this a bit confusing: "During the developer preview" - will this go away?

pages/content/amp-dev/documentation/guides-and-tutorials/start/bento_guide.md
```diff
@@ -0,0 +1,332 @@
...
+
+`el.getApi().then(api => api.play())`
+
+The code below triggers [`amp-base-carousel`'s next action](../../../documentation/components/reference/amp-base-carousel.md#next) when executed. 
+
+```js
```
sebastianbenz, line 312: Show HTML and JS together. Otherwise it's unnecessarily hard to reproduce. Also, `$('amp--base-carousel');` is not a browser API.

pages/content/amp-dev/documentation/guides-and-tutorials/start/bento_guide.md
```diff
@@ -0,0 +1,332 @@
...
+const carousel = $('amp--base-carousel');
+const apiPromise = carousel.getApi();
+apiPromise.then(api => {
+  api.next();
+})
+```
```
sebastianbenz, line 318: Maybe add a sentence mentioning where component APIs are documented.

sebastianbenz (commented):
Left a few additional comments. We also need a section describing how layouts work with Bento.

pages/content/amp-dev/documentation/guides-and-tutorials/start/bento_guide.md
CrystalOnScript: This is the current implementation, since there is no JS API

pages/content/amp-dev/documentation/guides-and-tutorials/start/bento_guide.md
CrystalOnScript: > Left a few additional comments. We also need a section describing how layouts work with Bento.

I've ben thinking a bit about this, and talking about AMP layouts and reproducing them seems not useful to non-AMP developers, which I think is a target audience for this document? Mentioning the AMP layouts, and that they can't be used seems like it would be confusing for developers who have never used AMP. 

Is there something specific to these components around dimensions and layout that is standard across the components that isn't dependent on the tech stack the developer is inserting the components into?

sebastianbenz: The target audience is AMP and non-AMP developers. Imo mentioning that AMP layouts are not supported is a must, but it does not need to be the first thing to mention. 

> Is there something specific to these components around dimensions and layout that is standard across the components that isn't dependent on the tech stack the developer is inserting the components into?

//cc @caroqliu @dvoytenko

pages/content/amp-dev/documentation/guides-and-tutorials/start/bento_guide.md
sebastianbenz: But it sounds like it will go away afterwards.

pages/content/amp-dev/documentation/guides-and-tutorials/start/bento_guide.md
CrystalOnScript: Are the API documented anywhere that I could link to? @caroqliu @dvoytenko

dvoytenko: Not yet. That's the same question as we are trying to figure out in the amp-social-share's PR. We need to come up with a structure here that works for us.

jaygray0919:
IMHO you should address these points.

"To achieve {standard AMP benefits} we do several things. For example, we don't allow `!important` {for your reasons}. Also some components depend on specific CSS that cannot be changed (e.g. our nemesis of making a `<template>` a block element vs. an inline element). Bento is designed for developers who depend on those capabilities but also want the features and efficency of core AMP services (from your list)."

On a similar line of reasoning, `<amp-script>` confuses many developers. One has to make frequent trips to the exception page to be reminded of non-compatible JS components. In general, folks don't like being told what they can't do. If, alternatively, you lead with "what you can do" (and show examples on AMP Playground) folks will be more accepting IMHO.

Bottom line: don't take the `<amp-script>` approach with Bento. Clearly show what can be done in the context of what people already are doing. For each Bento component, there's a CodePen example. Why should I use Bento when a CodePen example seems to meet my needs?

In our case, we would like to use Handelbars (which does not have the block-level consequences that we *rage* about WRT `<template>`) and accordion, carousel, sidebar and the other AMP goodies.

morsssss:
@jaygray0919 are these points you think must be addressed in this documentation?

Personally, I don't think we need to mention details about the CSS restrictions associated with valid AMP. On the other hand, I'd join you in advising against using `<amp-script>` in a Bento environment, since developers will have access to regular JavaScript. Anyone who wants to use JS in a Worker can simply do so. On the other other hand, I don't believe there are any plans to offer `<amp-script>` as a Bento component... for the same reasons. On the other other other hand, someone who wanted to use WorkerDOM could simply... use WorkerDOM, minus `<amp-script>`'s shim to AMP.

jaygray0919:
Comments were "in general" and probably not appropriate here. The point about `<amp-script>` is that it is an effort to accommodate developers who want to use their "favorite techniques." But IMHO it increases frustration because many "techniques" are not supported (e.g. we use D3 extensively). My "Bento-point" is to reverse the `<amp-script>` "with limits" approach. Many harsh critics argue: AMP is a prison. Turn the tables. Bento is about improving functions that developers use for much of their work. CodePen is a source of many of those funtions. AMP is ... (your performance story) and Bento is a collection of those high performance functions/utilities that work well with the way current applications work - which now will work better. I can use my substitution routines (e.g. handlebars, since `<template>` breaks my conventional patterns) with carousel, accordion, sidebar et al (Bento components). Let me have my cake and eat it too. Flame off.

morsssss:
@jaygray0919 this is why many of us have been looking forward to this Bento moment for a long time. With this, plus changes planned by Google Search, it will be much easier to have our cakes and eat them too. Enjoy! 🎂🎂🎂

pages/content/amp-dev/documentation/guides-and-tutorials/start/bento_guide.md
```diff
@@ -0,0 +1,274 @@
...
+        controls="never"
+        loop
+        auto-advance
+        auto-advance-interval="5000"
+      >
+      <img
+          class="article-img"
+          width="1024"
+          height="682"
+          src="https://raw.githubusercontent.com/ampproject/amp.dev/future/examples/static/samples/img/green_apple_1_1024x682.jpg"
```
sebastianbenz, line 189: use the amp.dev link as it's shorter

pages/content/amp-dev/documentation/guides-and-tutorials/start/bento_guide.md
```diff
@@ -0,0 +1,274 @@
...
+        >
+      <img
+          class="article-img"
+          width="1024"
```
sebastianbenz, line 193: the images have a different aspect ratio than the carousel is that intended?

pages/content/amp-dev/documentation/guides-and-tutorials/start/bento_guide.md
```diff
@@ -0,0 +1,274 @@
...
+            src="https://raw.githubusercontent.com/ampproject/amp.dev/future/examples/static/samples/img/product2_1024x682.jpg"
+          >
+      </section>
+    </amp-accordion>
+    <button id="toggle">
+    Click
+  </button>
+  <script>
+    const toggleAccordion = async () => {
+    const accordion = document.querySelector('#myAccordion')
```
sebastianbenz, line 249: nit: fix formatting

pages/content/amp-dev/documentation/guides-and-tutorials/start/bento_guide.md
```diff
@@ -0,0 +1,274 @@
...
+      Next Slide
+    </button>
+    <script>
+        const nextButton = document.querySelector('#next-button');
+        const prevButton = document.querySelector('#previous-button');  
+        const carousel = document.querySelector('amp-base-carousel');
+        
+        customElements
+        .whenDefined('amp-base-carousel')
+        .then(() => {
```
sebastianbenz, line 217: nit: shorter `then(() => carousel.getApi())`

pages/content/amp-dev/documentation/guides-and-tutorials/start/bento_guide.md
```diff
@@ -0,0 +1,274 @@
...
+            return carousel.getApi();
+        })
+        .then((api) => {
+            nextButton.addEventListener('click', () => api.next());
+            prevButton.addEventListener('click', () => api.prev());
+        });
+    </script>
+    <amp-accordion id="my-accordion" disable-session-states>
```
sebastianbenz, line 225: TBH, I think we don't need the accordion here. Demonstrating a single component should be sufficient to show how bento works. I think the carousel example works great for this.

sebastianbenz (changes requested):
I think this goes in the right direction. Left a few more comments, mostly nits.

pages/content/amp-dev/documentation/guides-and-tutorials/start/bento_guide.md
```diff
@@ -0,0 +1,274 @@
...
+```
+await customElements.whenDefined('amp-bento-component');
+const api = await component.getApi();
+api.callMethod();
+```
+
+The example below triggers [`amp-base-carousel`'s next action](../../../documentation/components/reference/amp-base-carousel.md#next) when the user clicks the Next Slide button.
+
+```
+<amp-base-carousel>...</amp-base-carousel>
```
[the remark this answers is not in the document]
CrystalOnScript: Absolutely, but it is the same as the example below. I was trying to highlight the code developers need to add outside of the component. But maybe we switch to an example with shorter component code - wdyt?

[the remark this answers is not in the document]
CrystalOnScript: I've removed the accordion example below and placed it here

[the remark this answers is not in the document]
sebastianbenz: perfect!

dvoytenko:
@jaygray0919 @morsssss FWIW, the `!important` requirement will not carry over to Bento. We can't enforce it there. And also components are now additionally protected from CSS collisions.

pages/content/amp-dev/documentation/guides-and-tutorials/start/bento_guide.md
dvoytenko: QQ: should we switch to `await` for these examples? It makes things a bit cleaner and a lot of folk either rely on Babel or can translate back and force.

sebastianbenz: +1 for await

CrystalOnScript: I've updated the example, but I was struggling to assign `next` and `prev` in the same function. Any advice?

pages/content/amp-dev/documentation/guides-and-tutorials/start/bento_guide.md
```diff
@@ -0,0 +1,274 @@
...
+
+[tip type="note"]
+Eliminating need for the AMP runtime script is a high priority for Bento AMP. Follow our progress on the [Bento roadmap](../../../community/roadmap.html?category=bento#status-updates).
+[/tip]
+
+Read each Bento component’s reference documentation for implementation details.
+
+## Layout and styling
+
+Fully valid AMP pages use the AMP layout system to infer sizing of elements to create a page structure before downloading any remote resources. This is important since this significantly reduces rendering and scrolling jank. However, Bento use imports components into less controlled environments, increasing the risk of content shifting.
```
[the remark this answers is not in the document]
CrystalOnScript: I've added what I could in to this section - starting on line 56. I could use help filing in the rest of the CSS properties and figuring out if we're missing any layouts we should advise on.

pages/content/amp-dev/documentation/guides-and-tutorials/start/bento_guide.md
```diff
@@ -0,0 +1,286 @@
...
+
+[/tip]
+
+## Interacting with Bento components
+
+Bento components have individual APIs that grant access to component actions and events. Access to component APIs and invoking events use the following syntax:
+
+```
+await customElements.whenDefined('amp-bento-component');
+const api = await component.getApi();
```
dvoytenko, line 141: Let's rename "component" here to "element".

pages/content/amp-dev/documentation/guides-and-tutorials/start/bento_guide.md
```diff
@@ -0,0 +1,286 @@
...
+Bento components don't support [AMP's layout](https://amp.dev/documentation/guides-and-tutorials/learn/amp-html-layout/). 
+
+[/tip]
+
+## Interacting with Bento components
+
+Bento components have individual APIs that grant access to component actions and events. Access to component APIs and invoking events use the following syntax:
+
+```
+await customElements.whenDefined('amp-bento-component');
```
dvoytenko, line 140: Do you think it's worth commenting that this only needs to be done once per all `amp-bento-component` elements in the document? Or is it self-obvious?

pages/content/amp-dev/documentation/guides-and-tutorials/start/bento_guide.md
```diff
@@ -0,0 +1,286 @@
...
+  </amp-base-carousel>
+  <button id="previous-button">
+    Previous Slide
+  </button>
+  <button id="next-button">
+    Next Slide
+  </button>
+  <script>
+    async function nextSlide() {
+      const carousel = await document.querySelector('amp-base-carousel')
```
dvoytenko, line 254: No `await` here. `document.querySelector` is a synchronous function. We could, though, add `await customElements.whenDefined('amp-base-carousel')` if you'd like.

pages/content/amp-dev/documentation/guides-and-tutorials/start/bento_guide.md
```diff
@@ -0,0 +1,286 @@
...
+      const api = await carousel.getApi();
+      api.next();
+    }
+    async function prevSlide() {
+      const carousel = await document.querySelector('amp-base-carousel')
```
dvoytenko, line 259: Ditto

dvoytenko (changes requested):

pages/content/amp-dev/documentation/guides-and-tutorials/start/bento_guide.md
CrystalOnScript: Thanks for the catch! I've added some clarification - it's always worth clarifying anything that may be confusing.

pages/content/amp-dev/documentation/guides-and-tutorials/start/bento_guide.md
CrystalOnScript: I've removed both `awaits`. Is use of `await customElements.whenDefined('amp-base-carousel')` recommended?

dvoytenko: I think for an example we should give the `await customElements.whenDefined('...')` to make it always functional. Elsewhere we can explain that it's only needed to be run once, but I think it's mostly self-explanatory.

dvoytenko (approved):

pages/content/amp-dev/documentation/guides-and-tutorials/start/bento_guide.md
CrystalOnScript: Could I get a more complete code example? I tried updating the example to include `await customElements.whenDefined('...')` last week and it continued to throw me the `Cannot read property 'getApi' of undefined` error.

pages/content/amp-dev/documentation/guides-and-tutorials/start/bento_guide.md
```diff
@@ -0,0 +1,295 @@
...
+    height: auto;
+    aspect-ratio: width / height;
+}
+```
+
+### Fill layout
+
+```css
+.fill-layout {
+...
```
dvoytenko, line 105: This is either:

```
position: absolute;
top: 0;
left: 0;
right: 0;
bottom: 0;
```

or
```
width: 100%;
height: 100%;
```

Depending on the context. The first one is fairly universal.

pages/content/amp-dev/documentation/guides-and-tutorials/start/bento_guide.md
```diff
@@ -0,0 +1,295 @@
...
+}
+```
+
+### Flex layout
+
+```css
+.flex-layout {
+    display: block;
```
dvoytenko, line 113: Flex should only need this:

```
  display: block;
  flex: 1 1 auto;
```

pages/content/amp-dev/documentation/guides-and-tutorials/start/bento_guide.md
```diff
@@ -0,0 +1,295 @@
...
+...
+}
+```
+
+### Responsive layout with a fixed aspect ratio
+
+```css
+.responsive-layout {
+    display: block;
+    width: 100%;
```
dvoytenko, line 95: No need for `width:100%` and `height:auto` for this.

pages/content/amp-dev/documentation/guides-and-tutorials/start/bento_guide.md
```diff
@@ -0,0 +1,295 @@
...
+    height: auto;
+    aspect-ratio: width / height;
```
dvoytenko, line 97: Worth calling out that this is a new CSS feature. There's an "old" way to do this. I sketched it out in [this bin](https://jsbin.com/yusenup/edit?html,css,output). Do you think it'd be worth showing this "fallback" here? Or too much into weeds?

pages/content/amp-dev/documentation/guides-and-tutorials/start/bento_guide.md
dvoytenko: Hmm. `await customElements.whenDefined('...')` should be always sufficient against `getApi undefined` errors.

pages/content/amp-dev/documentation/guides-and-tutorials/start/bento_guide.md
CrystalOnScript: Thanks for providing this! I'll include it as a note

pages/content/amp-dev/documentation/guides-and-tutorials/start/bento_guide.md
CrystalOnScript: I'm able to get it to work with this:
```
<button id="next-button">
Next Slide
</button>
<script>
let nextButton = document.getElementById('next-button')
    function nextSlide(){
      const carousel = document.querySelector('amp-base-carousel');
      customElements
        .whenDefined('amp-base-carousel')
        .then(() => {
          return carousel.getApi();
        })
        .then((api) => {
          // Make the carousel go to the next slide
          api.next();
      });
    }
 nextButton.addEventListener("click", nextSlide())
```

But this throws an error:

```
  <button id="next-button">
    Next Slide
  </button>
  <script>
    async function nextSlide() {
      const carousel = await customElements.whenDefined('amp-base-carousel');
      const api = await carousel.getApi();
      api.next();
    }
    const nextButton = document.querySelector('#next-button');
    nextButton.addEventListener('click', nextSlide);
  </script>
```

dvoytenko: The first one is correct (or one of possible correct) forms. The second has an error. `customElements.whenDefined` does not return an element, so a further call to `await carousel.getApi()` will fail. If we were to build on the first one with `await`, it'd look like this:

```
async function nextSlide(){
  const carousel = document.querySelector('amp-base-carousel');
  await customElements.whenDefined('amp-base-carousel');
  const api = await carousel.getApi();
  // Make the carousel go to the next slide
  api.next();
}
const nextButton = document.getElementById('next-button')
nextButton.addEventListener("click", nextSlide)
```

(a small nit: notice also that `addEventListener` should just have `nextSlide`, not `nextSlide()`).

pages/content/amp-dev/documentation/guides-and-tutorials/start/bento_guide.md
```diff
@@ -0,0 +1,303 @@
...
+*   [amp-selector](../../../documentation/components/reference/amp-selector-v1.0.md)
+*   [amp-social-share](../../../documentation/components/reference/amp-social-share-v1.0.md)
+*   [amp-timeago](../../../documentation/components/reference/amp-timeago-v1.0.md)
+*   Video components
+    *   [amp-video](../../../documentation/components/reference/amp-video-v1.0.md)
+    *   [amp-youtube](../../../documentation/components/reference/amp-youtube-v1.0.md)
+
+# Bento in action
+
+The example below demonstrates how to include `amp-base-carousel` and `amp-accordion` in a non-AMP HTML page.
```
sebastianbenz, line 226: it doesn't demo amp-accordion anymore.

sebastianbenz: Have you tried embedding this as a live example?

CrystalOnScript: It's live now!

sebastianbenz (approved):
Looks great! Thanks Crystal!

caroqliu (approved):

CrystalOnScript:
> Looks great! Thanks Crystal!

My pleasure!

PR #5206, opened 2021-01-12
🐛 Pixi: Check for falsy value instead of undefined specifically

### LINKED ISSUE #5205: Pixi: Always showing Error state message
Screen Shot 2021-01-12 at 10 39 21 AM

All the results are shown however, this message keeps showing up

matthiasrohmer:
Fixes #5205. `cacheExperience[dataType]` is either an object or `null` but never undefined.

The webpack config changes are not closely related to this bug but any of the latest updates to either the `CleanWebpackPlugin` or `FileManagerPlugin` caused a race condition that delete the created bundles before they can get copied to where express can find them.


sebastianbenz (approved):
Thanks for the quick fix!

PR #5207, opened 2021-01-13
log analytics events based on pixi searches

patrickkettner:
meant to collect information as narrowly as possible, let me know if it should change

### DIFF
--- a/pixi/backend/api.js
+++ b/pixi/backend/api.js
@@ -111,7 +111,7 @@ const logAnalytics = async (url) => {
     };
 
     return fetch(
-      `http://www.google-analytics.com/collect?${new URLSearchParams(data)}`
+      `https://www.google-analytics.com/collect?${new URLSearchParams(data)}`
     );
   } catch (e) {
     log.error('Unable to log', e.stack);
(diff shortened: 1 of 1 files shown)


pixi/backend/api.js
```diff
@@ -88,6 +90,34 @@ const execChecks = async (url) => {
...
+      // Application Name - for easier searching in GA
+      an: 'pixi',
+      // Event Action - required value
+      ea: 'lint',
+      // Event Label - where the URL is stored
+      el: new URL(url).host,
+    };
+
+    return fetch(
+      `http://www.google-analytics.com/collect?${new URLSearchParams(data)}`
```
sebastianbenz, line 114: `http`?

pixi/backend/api.js
```diff
@@ -97,6 +127,7 @@ api.get('/lint', async (request, response) => {
 
   const fetchUrl = request.query.url;
   try {
+    await logAnalytics(fetchUrl);
```
sebastianbenz, line 130: Why block on logging?

pixi/backend/api.js
patrickkettner: whoopsie!

pixi/backend/api.js
patrickkettner: thanks! vestigial code, removed the await in the last update

nainar:
OUR Privacy disclosure covers this given we are not collecting any data about the user other than the timestamp and the domain/URL entered. LGTM from me

nainar (approved):
See comment on thread

sebastianbenz (approved):

PR #5219, opened 2021-01-14
Remove JavaScript from Tink-supported languages

taymonbeal:
Tink for JavaScript only works in a browser environment, not in Node.js. Furthermore, it is in an extremely alpha-quality state, largely undocumented, and people shouldn't be directed to it at this point in its development unless they want to brave a lot of sharp edges.

I added Python to the list instead, since support for that is more mature.

If the AMP Project considers Tink Node.js support important for the AMP ecosystem, let me know and I'll see what we can do. There might not be much movement until March, though.


matthiasrohmer:
@jpettitt and @sebastianbenz for visibility.

Merging early as this might decrease the risk of grey hairs for users. Thanks @taymonbeal!

matthiasrohmer (approved):

Issue #5218, opened 2021-01-14
gulp bootstrap does not generate 404 page


sebastianbenz:
Running `gulp develop` after `gulp bootstrap` will result in the following error:

[21:21:15] 'run' errored after 26 s
[21:21:15] Error: ENOENT: no such file or directory, open 'amp.dev/dist/pages/404.html'
    at Object.openSync (fs.js:462:3)
    at readFileSync (fs.js:364:35)
    at Object.<anonymous> (/Users/sbenz/Projects/amp.dev/platform/lib/routers/notFound.js:26:20)
    at Module._compile (internal/modules/cjs/loader.js:1137:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1157:10)
    at Module.load (internal/modules/cjs/loader.js:985:32)
    at Function.Module._load (internal/modules/cjs/loader.js:878:14)
    at Module.require (internal/modules/cjs/loader.js:1025:19)
    at require (internal/modules/cjs/helpers.js:72:18)
    at Object.<anonymous> (/Users/sbenz/Projects/amp.dev/platform/lib/platform.js:46:13)
    at Module._compile (internal/modules/cjs/loader.js:1137:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1157:10)
    at Module.load (internal/modules/cjs/loader.js:985:32)
    at Function.Module._load (internal/modules/cjs/loader.js:878:14)
    at Module.require (internal/modules/cjs/loader.js:1025:19)
    at require (internal/modules/cjs/helpers.js:72:18)

PR #5221, opened 2021-01-15
dont read 404 file before its written

patrickkettner:
fix for #5218

bisecting the issue got me to  [this commit](https://github.com/ampproject/amp.dev/pull/5081) (mea culpa)

I think this was trying to read the file before it had been written. I hit quota limits from bootstrapping before I was able to test this, but I believe it should work


sebastianbenz (approved):

sebastianbenz:
Thanks!

Issue #4846, opened 2020-10-11
Migrate Pixi and Playground to webpack 5


matthiasrohmer:
Blocks #4842 and #4841.

matthiasrohmer:
I started working on this. Playground is almost up to date, currently running in an issue with [json-tree-view](https://www.npmjs.com/package/json-tree-view) which relies on polyfilled `util`. The WIP is at [chore/webpack5](https://github.com/ampproject/amp.dev/compare/chore/webpack5?expand=1).

PR #5240, opened 2021-01-19
🎨 Remove CleanWebpackPlugin in favor of FileManagerPlugin

matthiasrohmer:
Replaces #5230.


sebastianbenz (approved):
Do we also need this for the playground?

matthiasrohmer:
The playground currently only uses the CleanWebpackPlugin but not the FileManagerPlugin. I am currently working to consolidate plugin usage with https://github.com/ampproject/amp.dev/issues/4846.

PR #5244, opened 2021-01-20
update echo endpoint to work with amp4emails

patrickkettner:


examples/api/echo.js
```diff
@@ -24,7 +24,29 @@ const {setImmutable, setNoCache} = require('@lib/utils/cacheHelpers');
 const examples = express.Router();
 
 examples.get('/echo', (request, response) => {
-  if (request.header('Content-Type') != 'application/json') {
+  if (request.header('AMP-Email-Sender')) {
+    response
+      .set('AMP-Email-Allow-Sender', request.header('AMP-Email-Sender'))
```
sebastianbenz, line 29: I think this is handled by the Toolbox CORS middleware

sebastianbenz: The only thing needed is check if it's either content type JSON or an Email.

lluerich: @patrickkettner Is there any other way?

patrickkettner: sorry, I didn't mean to make this more complex than nessecary. The only thing that is required is checking for `request.query.__amp_source_origin`. If it is there, it is an email

lluerich: No worries, but isn't ```__amp_source_origin``` set on other requests as well?

patrickkettner: checking the amphtml source, it doesn't seem like it is used in many places. That is the only thing that amp4email is actually requiring for the component to function. There are additional CORS responsibilities, but those are enforced on the client and since we can just choose to not enforce them, then it doesn't need to be here.

sebastianbenz: __amp_source_origin used to be set for all AMP CORS requests. You need to check the value for whether it's an email. Let's keep things simple and just remove the JSON check here.

patrickkettner:
Sorry, I feel like this may be an XY problem, I should have asked for more
clarity before suggesting anything.
My understanding is that we were wanting to get the example to function
within the context of an amp4email document. Was that the goal? Or was it
purely to detect the document as amp4email on the server side?
-- 
patrick

examples/api/echo.js
patrickkettner: Sorry, I feel like this may be an XY problem, I should have asked for more clarity before suggesting anything. My understanding is that we were wanting to get the example to function within the context of an amp4email document. Was that the goal? Or was it purely to detect the document as amp4email on the server side?

patrickkettner: @sebastianbenz @lluerich

lluerich: Sorry for the delay. In my understanding the goal was purely to identify the request as amp4email on server side.

patrickkettner: in that case, there is nothing unique that would identify the requests of an amp4email document provided by the runtime.

lluerich: Alright, then I guess we can close this PR. The example is working now, since we switched to a POST request as an interim solution. @sebastianbenz is there anything else from your side?

PR #5247, opened 2021-01-20
Fix minor issues in subscription settings

TomerAberbach:
Fixed a typo.

### DIFF
--- a/examples/source/interactivity-dynamic-content/Subscription_Settings/index.html
+++ b/examples/source/interactivity-dynamic-content/Subscription_Settings/index.html
@@ -19,6 +19,7 @@ formats:
     <meta charset="utf-8">
     <link rel="canonical" href="<% canonical %>">
     <title>Subscription Settings</title>
+    <meta name="viewport" content="width=device-width,minimum-scale=1,initial-scale=1">
 
     <script async src="https://cdn.ampproject.org/v0.js"></script>
     <style amp4email-boilerplate>body{visibility:hidden}</style>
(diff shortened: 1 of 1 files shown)


examples/source/interactivity-dynamic-content/Subscription_Settings/index.html
```diff
@@ -19,7 +19,6 @@
     <meta charset="utf-8">
     <link rel="canonical" href="<% canonical %>">
     <title>Subscription Settings</title>
-    <meta name="viewport" content="width=device-width,minimum-scale=1,initial-scale=1">
```
sebastianbenz, line 22: Should this go away?

TomerAberbach: Yeah, I think so. I noticed that the AMP playground automatically adds the `meta` viewport tag. You can open a snippet in the playground from the [page](https://amp.dev/documentation/examples/interactivity-dynamic-content/subscription_settings/?format=email) to see.

So if we don't remove this, then it'll be added twice

sebastianbenz: Looks like the playground doesn't correctly pre-process email snippets for the playground //cc @patrickkettner. Manually removing the viewport here should not be needed.

TomerAberbach: Ah I see. Let me revert this commit and I'll let @patrickkettner handle the viewport issue

sebastianbenz:
Thanks!

PR #5251, opened 2021-01-24
ampbench has been turned down

ithinkihaveacat:
See https://github.com/ampproject/ampbench#status.


matthiasrohmer (approved):
Thanks for updating!

PR #5257, opened 2021-01-25
🐛 Fix amp-validation error caused by inline custom properties

### LINKED ISSUE #5256: Bento pill invalid AMP: properties are not allowed in inline styles
This is a problem for the bento pill: 
```
<div class="ap-a-pill ap-a-pill-small" style="--background-color:
  #D45A5A; --color: #FFF;"> Bento </div>
```

lluerich:
* Remove inline custom properties
* Create modifier class in CSS
* Fixes #5256

### DIFF
--- a/frontend/scss/_variables.scss
+++ b/frontend/scss/_variables.scss
@@ -59,6 +59,7 @@ $colors: (
   'slack':           #45184B,
   'lav-magenta':     #EB49E1,
   'electric-violet': #5500D7,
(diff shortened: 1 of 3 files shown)


matthiasrohmer (approved):
Thanks, Leo! Why though ... 🙁

sebastianbenz:
We're not the first ones to run into this... https://github.com/ampproject/amphtml/issues/24262

PR #5258, opened 2021-01-26
Add bento js + CSS copy script

sebastianbenz:
This adds a pill to the copy script for toggling between valid AMP and
Bento mode.

Screen Shot 2021-01-26 at 11 44 44

Screen Shot 2021-01-26 at 11 44 48

### DIFF
--- a/frontend/scss/components/atoms/copy-script.scss
+++ b/frontend/scss/components/atoms/copy-script.scss
@@ -7,12 +7,12 @@
   display: flex;
   align-items: center;
 
-  code {
+  pre {
     overflow-x: scroll;
     flex: 1 0;
     margin: 6px 8px 6px 0;
     padding: 4px 6px;
-    white-space: nowrap;
+    white-space: pre;
     background: color('iron');
     font-size: 12.6px;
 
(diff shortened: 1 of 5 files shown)


sebastianbenz:
Ferdi is probably not going to be happy about my pill usage here...

matthiasrohmer:
I am also heavily confused by looking at the screens: does dark blue in this context mean "press me" or "active" 😅 

The second screenshot shows a stringified JSON `{css: ...` where and how would that be used?

sebastianbenz:
Ups - fixed the stringified JSON. 

Screen Shot 2021-01-26 at 12 11 23

Blue means active - isn't this how we're using it everywhere?

matthiasrohmer:
Blue is also our color for the primary call to action 😁   this has also confusing me for the filter bubbles before.

Bildschirmfoto 2021-01-26 um 12 31 49
Bildschirmfoto 2021-01-26 um 12 31 53
Bildschirmfoto 2021-01-26 um 12 32 02

But this will work for now, forwarded it to Ferdi anyway 🙂

matthiasrohmer (approved):

sebastianbenz:
Let's discuss it this afternoon.

PR #5259, opened 2021-01-26
Bento info

sebastianbenz:
This will show a note for component with a bento version. We'll also
generate a JSON file listing all available bento components.

Screen Shot 2021-01-26 at 14 11 34


frontend/templates/views/partials/component-intro.j2
```diff
@@ -1,5 +1,11 @@
 {% do doc.styles.addCssFile('css/components/organisms/component-intro.css') %}
 <div class="ap-o-component-intro">
+  {% if doc.bentoPath and not doc.bento %}
+  {% do doc.icons.useIcon('icons/speech-bubble.svg') %}
+{% do doc.styles.addCssFile('css/components/molecules/tip.css') %}
+  <div class="ap-m-tip"> <div class="ap-a-ico ap-m-tip-icon"> <svg><use xmlns:xlink="http://www.w3.org/1999/xlink" xlink:href="#speech-bubble"></use></svg> </div> <div class="ap-m-tip-content">You can now use this component outside valid AMP documents using the <a href="{{ doc.bentoPath }}">experimental Bento version of this component</a>. Learn more in our <a href="../../guides-and-tutorials/start/bento_guide.md">Bento guide</a>.</div>
```
matthiasrohmer, line 6: Depends on #5198 - otherwise this will cause a broken link which will most probably break build:

Bildschirmfoto 2021-01-26 um 16 37 41

platform/lib/pipeline/componentReferenceImporter.js
```diff
@@ -360,7 +387,7 @@ class ComponentReferenceImporter {
         fileContents,
         extension
       );
-      await doc.save(docPath);
+      return doc;
```
matthiasrohmer, line 390: With this being a return will the `catch` still take action?

platform/lib/utils/project.js
```diff
@@ -62,6 +62,7 @@ const paths = {
   INLINE_EXAMPLES_DEST: absolute('dist/inline-examples'),
   PAGES_SRC: absolute('pages/content/amp-dev'),
   STATICS_DEST: absolute(`${DIST_DIR}/static`),
+  BENTO_COMPONENTS_LIST: absolute(`${DIST_DIR}/static/bento-components.json`),
```
matthiasrohmer, line 65: What's the plan with this?

frontend/templates/views/partials/component-intro.j2
```diff
@@ -1,5 +1,11 @@
...
+</div>
```
caroqliu, line 7: optional: suggested indents to help this section be more readable 
```suggestion
  {% do doc.styles.addCssFile('css/components/molecules/tip.css') %}
  <div class="ap-m-tip">
    <div class="ap-a-ico ap-m-tip-icon">
      <svg><use xmlns:xlink="http://www.w3.org/1999/xlink" xlink:href="#speech-bubble"> </use></svg>
    </div>
    <div class="ap-m-tip-content">
      You can now use this component outside valid AMP documents using the <a href="{{ doc.bentoPath }}">experimental Bento version of this component</a>. Learn more in our <a href="../../guides-and-tutorials/start/bento_guide.md">Bento guide</a>.
    </div>
  </div>
```

platform/lib/pipeline/componentReferenceImporter.js
caroqliu, line 390: Do we need to update the fn signature for this change?

caroqliu (approved):
Thanks for doing this!

Is anyone else seeing this extra whitespace on the local server? Could it just be my local build?
image

CrystalOnScript (approved):

platform/lib/utils/project.js
matthiasrohmer: We might use it to automate this list here: https://github.com/ampproject/amp.dev/pull/5198/files#diff-4184c78462ed1fab4a0530e0ada531051198fa358ce91308ac1e08ad1fc305d6R206-R222 🙂

sebastianbenz: Hah! Now I remember, that was part of the reason why I added it.

frontend/templates/views/partials/component-intro.j2
sebastianbenz: Good point. Removed for now.

westonruter:
Is the `bento-components.json` file stale? I just found that it is missing `amp-video-iframe` and `amp-vimeo`. I want to make use of these two files as part of the WordPress AMP plugin's build process in order to determine that is the `latest` version of each component and what version is Bento (and whether it is `experimental` or not):

* https://amp.dev/static/bento-components.json
* https://amp.dev/static/files/component-versions.json

But both of these files seem to be not entirely accurate or at least they are outdated. See https://github.com/ampproject/amp-toolbox/pull/1235/files#r637225246.

PR #5264, opened 2021-01-26
Bento script polish

sebastianbenz:
* use radio buttons instead of pills
* wrap script text

Screen Shot 2021-01-26 at 20 56 10

### DIFF
--- a/frontend/templates/views/partials/component-intro.j2
+++ b/frontend/templates/views/partials/component-intro.j2
@@ -26,16 +26,13 @@
     {% if doc.bento %}
     <div class="ap-o-component-intro-filter">
       <label>
-      <input type="radio" name="format-switch"
-        class="ap-a-pill ap-a-pill-flat ap-a-pill-small ap-a-pill-link filtered"
-        [class]="bento ? 'ap-a-pill ap-a-pill-flat ap-a-pill-small ap-a-pill-link ' : 'ap-a-pill ap-a-pill-flat ap-a-pill-small ap-a-pill-link filtered'"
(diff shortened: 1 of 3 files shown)


frontend/templates/views/partials/component-intro.j2
```diff
@@ -21,13 +21,23 @@
...
-    </h4>
+  <div class="ap-o-component-intro-row ap-o-component-intro-row-inline">
+    <h4 class="ap-o-component-intro-headline">{{ _('Required Scripts') }}</h4>
+    {% if doc.bento %}
+    <div class="ap-o-component-intro-filter">
+      <label>
+      <input type="radio" name="format-switch"
+        class="ap-a-pill ap-a-pill-flat ap-a-pill-small ap-a-pill-link filtered"
+        [class]="bento ? 'ap-a-pill ap-a-pill-flat ap-a-pill-small ap-a-pill-link ' : 'ap-a-pill ap-a-pill-flat ap-a-pill-small ap-a-pill-link filtered'"
+        on="tap:AMP.setState({ 'bento': false })" checked>AMP
```
caroqliu, line 32: suggestion: [`toggleClass`](https://amp.dev/documentation/guides-and-tutorials/learn/amp-actions-and-events/?referrer=ampproject.org#*-(all-elements)) seems more suitable here.

caroqliu (approved):

frontend/templates/views/partials/component-intro.j2
sebastianbenz: Good point - with the radio button amp-bind is no longer needed. Changed to `hide` `show` and removed amp-bind.

PR #5273, opened 2021-01-28
track cvw in analytics

sebastianbenz:


patrickkettner:
I don't see any issues, but given my lack of experience with amp-analytics it would be great to get @micajuine-ho to give it a glance

patrickkettner (approved):

frontend/templates/views/partials/analytics.j2
```diff
@@ -18,6 +18,18 @@
...
+      'on': 'visible',
+      'request': 'pageview',
+      'vars': {
+        'title': '{{title}}'
+      },
+      'extraUrlParams': {
+        'lcp': '${largestContentfulPaint}',
+        'cls': '${cumulativeLayoutShift}',
+        'fid': '${firstInputDelay}'
+      }
```
micajuine-ho, line 31: I would suggest that we add this to a special request rather than `pageview`, since we have discussed that these macros might take a while to resolve, delaying the firing of the request.

sebastianbenz: What kind of special request would be appropriate here?

micajuine-ho: Special, as in individual requests for each CWV:

```
{
  "triggers": {
    "pageviewCls": {
      "on": "visible",
      "request": "endpoint",
      "extraUrlParams": {
        "cls": "${cumulativeLayoutShift}"
      }
    },
    "pageviewLcp": {
      "on": "visible",
      "request": "endpoint",
      "extraUrlParams": {
        "lcp": "${largestContentfulPaint}"
      }
    },
    "pageviewFid": {
      "on": "visible",
      "request": "endpoint",
      "extraUrlParams": {
        "fid": "${firstInputDelay}"
      }
    }
  }
}
```

sebastianbenz: Ah got it - thanks! Do you happen to know what'd be the correct endpoint to use for Google Analytics?

micajuine-ho: I think it would probably be `event`

patrickkettner:
@micajuine-ho I redid the config per your suggestions. Could you see if I understood you correctly?

Thanks!

frontend/templates/views/partials/analytics.j2
```diff
@@ -20,156 +20,188 @@
...
   },
+  'requests': {
+    "base": "https://google-analytics.com/collect/"
+    "CWV_EVENT": "${base}?v=1&t=event&tid=${gtag_id}&cid=0&ec=cwv
+  },
   'triggers': {
-    'banner': {
-      'selector': '#top-banner',
-      'on': 'click',
+    'default pageview': {
```
micajuine-ho, line 28: I'm not sure if we allow whitespaces in trigger names, but also I think we should follow (with what I think is naming convention) and use camel case here.

micajuine-ho (approved):

frontend/templates/views/partials/analytics.j2
patrickkettner: done! thanks

PR #5278, opened 2021-01-29
remove noop code

patrickkettner:
@sebastianbenz does this do anything? it looks like a bunch of noops


sebastianbenz (approved):
thanks for cleaning this up!

PR #5280, opened 2021-01-30
resolve css variables to their static values for amp4email examples

### LINKED ISSUE #5227: Missing Templates for AMP for Email Components in AMP Playground
## What's the issue?

It looks like the templates for AMP for Email components have disappeared from the AMP Playground.

https://playground.amp.dev/?runtime=amp4email

1. Open the AMP Playground
2. Choose AMP for Email
3. Click untitled in the top/middle of the page to open the Select Template modal. 

Only `amp-autocomplete-email`, and `amp-bind-macro` components are shown.

Previously you would see `amp-form`, `amp-list`, etc.

## What browsers are affected?

Chrome, Firefox, possibly all.

## Which AMP version is affected?

This is a new issue, as templates would always show beforehand. 

## Screenshot
amp components

patrickkettner:
fix #5227


sebastianbenz (approved):
Nice! Can you please rebase against master.

PR #5286, opened 2021-02-01
Improve stage rendering on homepage

sebastianbenz:
* Enable blurry placeholders (only for homepage to save CPU)
* Transform all stage amp-img to img by increasing the # of hero images
to 5

### DIFF
--- a/platform/lib/routers/growPages.js
+++ b/platform/lib/routers/growPages.js
@@ -244,6 +244,11 @@ growPages.get(/^(.*\/)?([^\/\.]+|.+\.html|.*\/|$)$/, async (req, res, next) => {
         experimentEsm,
         preloadHeroImage,
       };
+      // Enable blurred placeholders and render all stage images for homepage
+      if (req.path === '/') {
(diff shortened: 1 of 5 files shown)


platform/lib/routers/growPages.js
```diff
@@ -244,6 +244,11 @@ growPages.get(/^(.*\/)?([^\/\.]+|.+\.html|.*\/|$)$/, async (req, res, next) => {
         experimentEsm,
         preloadHeroImage,
       };
+      // Enable blurred placeholders and render all stage images for homepage
+      if (req.path === '/') {
+        params.blurredPlaceholders = true;
+        params.maxHeroImageCount = 5;
```
patrickkettner, line 250: Any reason 5 was chosen?

patrickkettner (approved):

platform/lib/routers/growPages.js
sebastianbenz: yes, five images in the stage

PR #5288, opened 2021-02-01
Corrected sizing

CrystalOnScript:


alankent (commented):
The video section looks good, but you might want to check the "Images" section. I don't know if the image is a different resolution on purpose (I don't know - but it is not automatic that the image and video should be the same size).

CrystalOnScript:
@newmuis @Enriqe Would either of you mind confirming the image resolution? Thanks in advanced!

sebastianbenz (approved):

PR #5330, opened 2021-02-09
💄 Update tip style

lluerich:
Another update of the tip styling.


matthiasrohmer (approved):

frontend/scss/components/molecules/tip.scss
```diff
@@ -24,6 +24,7 @@ The tip molecule shows text alongside a contextual icon
   margin-bottom: 20px;
   border-radius: 4px;
   background-color: color('athens-gray');
```
matthiasrohmer, line 26: Shall they stay gray?

lluerich: Yes, but now they have a slightly darker border.

PR #5337, opened 2021-02-10
start rolling-formats mid animation to reduce change

patrickkettner:
As the LCP is being triggered around 3 seconds, and is localized to the header, I was thinking the rolling formats may be the culprit. If this is the case, having text show immediately, or elongating the time that the initial the word is shown could help?


sebastianbenz (approved):
Good idea! Let's try! I actually prefer the animation this way. Fixed a small glitch in the animation.

patrickkettner:
So this raised LCP from 3 seconds, to 4.5. Which would seem to confirm that the animation is what is causing the issue - it just needs a longer delay window. will look into this, after high priority stuff is taken care of 🙃

PR #5341, opened 2021-02-10
Add `sizes` for hero images in home stage

sebastianbenz:
@kristoferbaxter noticed that browsers would always download the largest version of the image, which is way bigger than needed.

The matching `srcset` will be generated by AMP Optimizer.


patrickkettner (approved):

patrickkettner:
Is this something that should be added to optimizer? Or do we not have a good way to expose it?

sebastianbenz:
That definitely should be added to optimizer.

PR #5350, opened 2021-02-12
Update husky

patrickkettner:
closes #5327 

There was some [breaking changes](https://dev.to/typicode/what-s-new-in-husky-5-32g5), so more work than just updating the module was required


patrickkettner:
sorry for the noise - actually good to review now

sebastianbenz (approved):
Thanks!

PR #5353, opened 2021-02-12
enable ESM version of AMP runtime

sebastianbenz:


platform/lib/routers/growPages.js
```diff
@@ -243,13 +243,7 @@ growPages.get(/^(.*\/)?([^\/\.]+|.+\.html|.*\/|$)$/, async (req, res, next) => {
...
     const optimize = req.query.optimize !== 'false';
     if (optimize) {
-      const experimentEsm = !!req.query.esm || false;
-      const optimizeHeroImages =
-        req.query.hero === undefined ? true : !!req.query.hero;
-      const params = {
-        experimentEsm,
-        optimizeHeroImages,
-      };
+      const params = {};
```
patrickkettner, line 246: this isn't being used, its just being passed to `transformHtml`,  which is [passing it untouched](https://github.com/ampproject/amp-toolbox/blob/94956207cd0de658da9553424c4881481e96a612/packages/optimizer/lib/DomTransformer.js#L125-L127) to `transformTree`, which has a [default of `{}`](https://github.com/ampproject/amp-toolbox/blob/94956207cd0de658da9553424c4881481e96a612/packages/optimizer/lib/DomTransformer.js#L136).

Is there a reason this needs to be here?

.husky/.gitignore
```diff
@@ -0,0 +1 @@
+_
```
patrickkettner, line 1: ?

platform/lib/routers/growPages.js
sebastianbenz: It's used here `params.maxHeroImageCount = 5;`

```
      const params = {};
      // Enable blurred placeholders and render all 5 stage images of the homepage
      if (req.path === '/') {
        // Disabled until we know why it fails after build
        // params.blurredPlaceholders = true;
        params.maxHeroImageCount = 5;
      }
      renderedTemplate = await optimizer.transformHtml(
        renderedTemplate,
        params

```

patrickkettner (approved):

platform/lib/routers/growPages.js
patrickkettner: oh pft. totally missed it. lgtm!

PR #5356, opened 2021-02-12
use html preload rather than the header variant

patrickkettner:


patrickkettner:
total shot in the dark, but before this change I am not getting the font preloaded, after I am. I am not sure if the Header variant isn't supported? Or even if this is a way that is acceptable - its just how they were previously preloaded

sebastianbenz (approved):

PR #5407, opened 2021-02-21
Prevent keyboard popup in iOS

ckybonist:
## Bug Description
In iOS safari / chrome, keyboard was popped up after clicking the copy button.

## Changes
Enable `readonly` mode on input tag


patrickkettner:
thanks!

PR #5414, opened 2021-02-22
Fixed typo `allow<-same-origin`

ibnujakaria:
Fixed typo `allow<-same-origin` to `allow-same-origin`


sebastianbenz (approved):
Thanks for fixing!

PR #5443, opened 2021-02-25
preload critical font noto

sebastianbenz:


sebastianbenz:
I've noticed that nodes in the first viewport still use noto so it might be better to preload as well.

patrickkettner:
SGTM

does woff2 cover the usage of browsers that are using the site? Like could we be not preloading a font that a non trivial amount of people are needing?

patrickkettner (approved):

sebastianbenz:
You mean many browsers don't support woff2?

sebastianbenz:
For reference: 

* WPT https://webpagetest.org/result/210225_DiPB_26f1c29dc96b0ebce959a7e39f63be1e/

PR #5490, opened 2021-03-08
Add "Video in AMP" to spec.json

alanorozco:
This document gets referenced enough that it may need a place on amp.dev

### DIFF
--- a/platform/config/imports/spec.json
+++ b/platform/config/imports/spec.json
@@ -82,6 +82,15 @@
 		"formats": ["websites","email","stories", "ads"]
 	},
 	{
+		"title": "Video in AMP",
+		"order": -1,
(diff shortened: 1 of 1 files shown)


platform/config/imports/spec.json
```diff
@@ -81,6 +81,15 @@
 		"toc": true,
 		"formats": ["websites","email","stories", "ads"]
 	},
+	{
+		"title": "Video in AMP",
+		"order": -1,
```
alanorozco, line 86: I'm not sure what to do with this property. 🙂

matthiasrohmer: It defines the order of the pages in the sidebar: 
image

See the order attributes of the other docs in [this collection ](https://github.com/ampproject/amp.dev/tree/future/pages/content/amp-dev/documentation/guides-and-tutorials/learn/spec)or pick something extremely high to add it to the end of the list.

matthiasrohmer (approved):

PR #5503, opened 2021-03-10
Migrate to official packager docker image

sebastianbenz:
If using both AMP Packager and a 2.8.0 canary release of AMP Optimizer, there is a conflict that will lead to invalid AMP pages.

### DIFF
--- a/packager/Dockerfile
+++ b/packager/Dockerfile
@@ -22,10 +22,6 @@ ENV GO111MODULE=on
 # RUN go get -v github.com/ampproject/amppackager/cmd/amppkg@master
 RUN go get -v github.com/ampproject/amppackager/cmd/amppkg
 
-# Install git
-RUN apt-get update \
-    && apt-get install -y git
-
 # Clone amppackager files so we can use b3 directory
 WORKDIR /data
 # Run this if you clone from master branch.
--- a/packager/Dockerfile
+++ b/packager/Dockerfile
@@ -1,23 +1,56 @@
-# Maybe switch to one of the Google-provided images? But seems to be go 1.8, not 1.10:
-# https://cloud.google.com/appengine/docs/flexible/custom-runtimes/build#base
-FROM golang:1.15
+# Copyright 2018 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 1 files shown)


sebastianbenz:
See https://github.com/ampproject/amppackager/blob/releases/docker/Dockerfile for docker reference

packager/Dockerfile
```diff
@@ -1,23 +1,56 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
 
-ADD https://github.com/Yelp/dumb-init/releases/download/v1.2.2/dumb-init_1.2.2_amd64 /usr/local/bin/dumb-init
-RUN chmod +x /usr/local/bin/dumb-init
-ENTRYPOINT ["dumb-init", "--"]
+# Use an official Go runtime as a parent image
+FROM golang:1.13
```
patrickkettner, line 16: why are we downgrading?

packager/Dockerfile
```diff
@@ -1,23 +1,56 @@
...
+# Install AMP Packager
+# Run this if you want to go off master branch.
+# RUN go get -v github.com/ampproject/amppackager/cmd/amppkg@master
+RUN go get -v github.com/ampproject/amppackager/cmd/amppkg
+
+# Install git
+RUN apt-get update \
+    && apt-get install -y git
+
+# Clone amppackager files so we can use b3 directory
```
patrickkettner, line 29: NIT: "b3 directory" is ambiguous, and redundant as it is immiedtly pointed out 5 lines later. It would be great to say _what_ the b3 directory is for

packager/Dockerfile
```diff
@@ -1,23 +1,56 @@
...
-# Copy whatever certs are available in the certs/ directory  into the container. 
-COPY certs/*pem* /amppkg/
+# Install AMP Packager
+# Run this if you want to go off master branch.
+# RUN go get -v github.com/ampproject/amppackager/cmd/amppkg@master
+RUN go get -v github.com/ampproject/amppackager/cmd/amppkg
+
+# Install git
+RUN apt-get update \
+    && apt-get install -y git
```
patrickkettner, line 27: the go docker images include `git` - do we need to have a specific version?

patrickkettner (approved):

packager/Dockerfile
sebastianbenz: @banaag do you know why we include git here?

banaag: The Dockerfile uses ./seedcache.sh which uses amppackager/testdata/b3. We include git so we can get the testdata/b3 directory into the container.

sebastianbenz: As Patrick said, the git installation step is not needed (just verified it).

PR #5510, opened 2021-03-12
Add pixi cloud function

sebastianbenz:
This is the first step in adding the new puppeteer based pixi checks to
amp.dev. This PR adds:

* a google cloud function integrating the new
toolbox-page-experience module
* deploy and bootstrap tasks
* basic api integration test

Subsequent PRs will integrate this into the pix API endpoint. This
version is currently deployed here: https://us-central1-amp-dev-230314.cloudfunctions.net/checkPageExperience?url=https://axios.com


pixi/cloud-function/index.js
```diff
@@ -0,0 +1,42 @@
...
+  if (!url) {
+    res.status(400).send('Missing url parameter');
+    return;
+  }
+  try {
+    new URL(url);
+  } catch (e) {
+    res.status(400).send(`invalid URL "${url}"`);
+    return;
+  }
```
matthiasrohmer, line 39: This method of URL validation has a shortcoming which made us fall back to a RegEx based solution for the Pixi frontend - I can't remember the reason though...

platform/lib/utils/project.js
```diff
@@ -68,6 +68,7 @@ const paths = {
   RECENT_GUIDES_DEST: 'pages/shared/data/recent-guides.yaml',
   BUILD_INFO_PATH: absolute('platform/config/build-info.yaml'),
   THUMBOR_ROOT: absolute('thumbor'),
+  PIXI_ROOT: absolute('pixi/cloud-function'),
```
matthiasrohmer, line 71: `PIXI_CLOUD_ROOT` maybe? I consider `pixi/` as Pixi's root.

gulpfile.js/build.js
```diff
@@ -167,6 +168,15 @@ async function buildPixi() {
   return sh('npm run build:pixi');
 }
 
+/**
+ * Builds the pixi cloud functions project
+ */
+function buildPixiFunctions() {
+  return sh('npm install', {
+    workingDir: PIXI_ROOT,
+  });
```
matthiasrohmer, line 177: That's just a stub, right?

matthiasrohmer (approved):
Only nits 🙂

gulpfile.js/build.js
sebastianbenz: how do you mean?  It needs to be its own npm module.

pixi/cloud-function/index.js
sebastianbenz: It doesn't check the protocol maybe?

platform/lib/utils/project.js
sebastianbenz: good suggestion! done

pixi/cloud-function/index.js
sebastianbenz: seems like it, using the same approach now

PR #5523, opened 2021-03-16
🚸  Fix errors around Pixi webpack dev config

matthiasrohmer:
/cc @lluerich

This enables the same notifications as for the frontend to know when a build succeeded.
image

But in the first place this fixes not building the bundle incrementally for `npm run start:pixi`.

### DIFF
--- a/pixi/webpack.config.js
+++ b/pixi/webpack.config.js
@@ -15,8 +15,6 @@ module.exports = (env, argv) => {
   return {
     entry: path.join(__dirname, 'src/ui/PageExperience.js'),
     output: {
-      filename: 'pixi.[name].[contenthash].js',
-      chunkFilename: 'pixi.[name].[chunkhash].bundle.js',
       filename: isDevelopment
         ? 'pixi.[name].js'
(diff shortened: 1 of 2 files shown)


pixi/webpack.config.js
```diff
@@ -3,18 +3,26 @@ const webpack = require('webpack');
...
 
 module.exports = (env, argv) => {
   const mode = argv.mode || 'production';
+  const isDevelopment = mode == 'development';
 
   return {
     entry: path.join(__dirname, 'src/ui/PageExperience.js'),
     output: {
       filename: 'pixi.[name].[contenthash].js',
       chunkFilename: 'pixi.[name].[chunkhash].bundle.js',
```
lluerich, line 19: Shouldn't this be removed in favour of the next lines?

lluerich (approved):
Looks reasonable if that one remark turns out to be wrong 😃

pixi/webpack.config.js
matthiasrohmer: Ugh...

PR #5533, opened 2021-03-18
functional survey

patrickkettner:


frontend/templates/views/partials/survey.j2
```diff
@@ -1,11 +1,14 @@
...
+  [% set url = "{{ doc.url }}" %]
+
   {% do doc.styles.addCssFile('css/components/organisms/survey.css') %}
 
   <amp-state id="surveyQuestions">
     <script type="application/json">
-
-      {% do survey_questions.update({"_url": request.url })%}
-      {{ survey_questions | tojson }}
+        [% set survey = surveys|random|survey(fixedQuestions, url) %]
```
sebastianbenz, line 10: What does the random do here?

sebastianbenz (approved):
What was the problem? Looks good once the build passes.

frontend/templates/views/partials/survey.j2
```diff
@@ -1,11 +1,14 @@
 {% if survey_questions %}
+  [% set surveys = {{ survey_questions|tojson }} %]
+  [% set fixedQuestions = {{fixed_questions|tojson }} %]
+  [% set url = "{{ doc.url }}" %]
```
matthiasrohmer, line 4: Alternatively you could also do

```suggestion
  [% set url = requestPath %]
```

This is given to the SSR context here:

https://github.com/ampproject/amp.dev/blob/2e3a1fb39f46110065c01ae0f8c8e4a8e11f1965/platform/lib/templates/index.js#L44

Not too sure about the backend but you also get the format the user is on, which might help putting the answers into context:

https://github.com/ampproject/amp.dev/blob/2e3a1fb39f46110065c01ae0f8c8e4a8e11f1965/platform/lib/templates/index.js#L54

frontend/templates/views/partials/first_survey.j2
```diff
@@ -0,0 +1,32 @@
+{% set survey_questions = [{
```
matthiasrohmer, line 1: No need to localize them just yet?

platform/lib/templates/SurveyFilter.js
```diff
@@ -0,0 +1,32 @@
+/**
+ * Copyright 2019 The AMP HTML Authors. All Rights Reserved.
```
matthiasrohmer, line 2: Nit: 2021

frontend/templates/views/partials/survey.j2
patrickkettner: `surveys` is an array of survey objects. `random` picks one of the available objects at runtime

frontend/templates/views/partials/first_survey.j2
patrickkettner: haven't sent the questions to the localization folks

patrickkettner: will add it seperatly

platform/lib/templates/SurveyFilter.js
patrickkettner: fixed!

patrickkettner:
@sebastianbenz 

> What was the problem? Looks good once the build passes.

The issue I was hitting is still an issue, but it was completely orthogonal to getting it to work. @matthiasrohmer was able to set me straight, I had to pass the object from grow to SSR, and then use different tags to pull in a version at request time, rather than at deploy time.

PR #5553, opened 2021-03-19
add pixi cli

sebastianbenz:


pixi/rollup.config.js
```diff
@@ -0,0 +1,37 @@
+import replace from '@rollup/plugin-replace';
+import resolve from '@rollup/plugin-node-resolve';
+import commonjs from '@rollup/plugin-commonjs';
+
+const config = require('./config.js');
```
matthiasrohmer, line 5: Nit: consistently use `require` or `import`.

matthiasrohmer (commented):
Looks good so far, only two remarks: 

- What's the idea behind using both Roll Up and Webpack? 
- I think instructions on API-Authentication (i.e. on how to retrieve `AMP_DEV_PIXI_APIS_KEY`) needs to be clearer or even make the build fail

pixi/rollup.config.js
sebastianbenz: This is intentional to get rollup to work (if I remember correctly).

sebastianbenz:
Not happy about mixing rollup and web pack either. Rollup was a lot easier to get to work for building a node compatible bundle. I'm sure that's also possible with web pack, but haven't had the time yet to investigate.

sebastianbenz:
Will merge this for now, but we should migrate to either rollup or webpack.

PR #5558, opened 2021-03-22
Pixi: Fix code snippet in tap deplay recommendation

### LINKED ISSUE #5417: Pixi: Disable tap delay has code formatting problem
## 🐛 Bug Report

The Disable tap delay advice has a problem with the code formatting:

image

### Current behavior

The `<code>` tags are rendered out, apparently being HTML-escaped.

### Expected behavior/code

Instead of:

```
<code><span class="err"><meta name="viewport" content="width=device-width"></span>
</code>
```

It should display:

```
<meta name="viewport" content="width=device-width">
```

sebastianbenz:
* Use `<code>` instead of `<pre>`
* Style `<code>` in recommendations
* Also use snarkdown instead of marked (saves 42kb)

Fixes #5417


lluerich (approved):
Nice and easy fix. 👍🏻

PR #5561, opened 2021-03-22
Pixi: add new hero images check

### LINKED ISSUE #5416: Pixi: Transform hero images advice given even when there is no amp-img in first viewport
## 🐛 Bug Report

On https://amp-wp.org we do not have any `amp-img` in the first viewport. All images are inline as SVGs. Nevertheless, doing a PX Guide check results in the “Transform hero images on the page” advice being given:

image

### Current behavior

Advice is given

### Expected behavior/code

Advice should not be given since not appropriate

sebastianbenz:
Use check based on Puppeteer to avoid false positives.

Fixes #5416

### DIFF
--- a/pixi/cloud-function/package.json
+++ b/pixi/cloud-function/package.json
@@ -10,6 +10,6 @@
 	"author": "AMPHTML Authors",
 	"license": "Apache 2",
 	"dependencies": {
-		"@ampproject/toolbox-page-experience": "2.8.0-canary.11"
+		"@ampproject/toolbox-page-experience": "2.8.0-canary.12"
 	}
 }
--- a/pixi/cloud-function/package.json
+++ b/pixi/cloud-function/package.json
@@ -10,6 +10,6 @@
 	"author": "AMPHTML Authors",
 	"license": "Apache 2",
 	"dependencies": {
-		"@ampproject/toolbox-page-experience": "2.8.0-canary.12"
+		"@ampproject/toolbox-page-experience": "2.8.0-canary.13"
 	}
 }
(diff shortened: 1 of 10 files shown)


pixi/cloud-function/package.json
```diff
@@ -10,6 +10,6 @@
 	"author": "AMPHTML Authors",
 	"license": "Apache 2",
 	"dependencies": {
-		"@ampproject/toolbox-page-experience": "2.8.0-canary.8"
+		"@ampproject/toolbox-page-experience": "2.8.0-canary.11"
```
patrickkettner, line 13: any reason we are using 11 here and [9 in the main package](https://github.com/ampproject/amp.dev/pull/5561/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R63)?

patrickkettner (approved):

pixi/cloud-function/package.json
sebastianbenz: optimizer is not yet on 11

sebastianbenz:
Just noticed that our isElementInFirstViewport check has one problem - it doesn't care about visibility. For example, it will surface iframes or images hidden inside an amp-sidebar, e.g. hero images reported for https://www.axios.com/trump-filibuster-senate-mcconnell-republican-933e9ed7-abd9-4dca-a702-1c97a858c0b5.html are not useful at all.

pixi/cloud-function/package.json
patrickkettner: oh pft. misread. apologizes.

sebastianbenz:
This has been fixed in https://github.com/ampproject/amp-toolbox/pull/1174

PR #5573, opened 2021-03-23
Fix results being overridden resulting from canary merge

sebastianbenz:


pixi/backend/api.js
```diff
@@ -142,3 +134,10 @@ api.get('/lint', async (request, response) => {
 });
 
 module.exports = api;
+async function runPageExperienceChecks(url) {
```
matthiasrohmer: Nit: having something after `module.exports` feels weird 😅

matthiasrohmer (approved):
Was just about to nag over formatting. LGTM now 😁

pixi/backend/api.js
sebastianbenz: 😰

PR #5581, opened 2021-03-25
🐛 Fix product page sample

### LINKED ISSUE #2835: Product Page sample leads to 404
This preview page seems to have lost a lot of its CSS. Also, there's a little box next to the word "Product" that's actually a form submission for a search API. Clicking that box submits the form to a URL that's no longer supported:

https://amp.dev/documentation/examples/api/?SEARCH=

lluerich:
The request now leads to a valid endpoint, but this endpoint redirects to a non existing page. @sebastianbenz does this suffice in terms of showing that a search endpoint could be behind a form?

Fixes #2835.

### DIFF
--- a/examples/source/e-commerce/Product_Browse_Page.html
+++ b/examples/source/e-commerce/Product_Browse_Page.html
@@ -420,7 +420,7 @@ author: aghassemi
             [value]="autosuggest.query || ''"
             value="">
           <!--~ Keep this just for creating the change event ~-->
-          <input type="submit" value="Search">
+          <input type="submit" value="">
         </div>
         <div class="suggest">
             <div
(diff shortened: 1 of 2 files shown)


sebastianbenz:
If I remember correctly this sample used to work and redirect to the product listing page.

sebastianbenz:
This page:
https://amp.dev/documentation/examples/e-commerce/product_browse_page/preview/?format=websites

>

lluerich:
Alright. Thanks for the link. Should we just redirect to that page then? Afaik we can't pass the query parameter to the preview iframe, so we can't actually search for an item from the product page.
Any other ideas?

sebastianbenz (approved):

PR #5582, opened 2021-03-25
🎨  Further enhance GitHub workflows

matthiasrohmer:
This tries to enhance workflows: actually running the tests on `push` should be enough; we need to investigate if that works for PRs created after everything has been pushed.

Also this fixes the nightly workflow - the required triggers weren't allowed for the release workflows.

Additionally this tries to implement skipping for concurrent release jobs on `future`. As this is not a core feature we need to rely on a 3p action which has [confusing documentation](https://github.com/fkirc/skip-duplicate-actions/).

All in all: fingers crossed 😅

### DIFF
--- a/.github/workflows/check-grow-link-integrity.yaml
+++ b/.github/workflows/check-grow-link-integrity.yaml
@@ -5,9 +5,6 @@ on:
   push:
     paths:
       - 'pages/content/**'
-  pull_request:
-    paths:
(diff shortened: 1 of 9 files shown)


matthiasrohmer:
(Also this bothers me: I didn't alter files affected by any checks, hence this PR will have pending checks ~forever)

Issue #5578, opened 2021-03-24
Debug and fix thumbor failover for /static/remote specially


matthiasrohmer:
When proxying requests over to thumbor and thumbor doesn't respond we actually try to failover to the original files either in the container image or blog.amp.dev - this doesn't seem to work.

PR #5601, opened 2021-03-29
🎨  Use node-fetch instead of proxy to get optimized images

matthiasrohmer:
Fixes https://github.com/ampproject/amp.dev/issues/5578 - getting response handling back from the proxy implementation is a little cumbersome, using `resEnd` events et. al. - I think the original implementation never had a chance to write to the response.

Hence I switched to a `node-fetch` based implementation which might be a little bit slower but easier to read and maintain.


sebastianbenz (approved):
Makes sense - thanks!

PR #5612, opened 2021-03-30
Advise on how to tweak CSP for service worker

westonruter:
As pointed out by @jono-alderson, the CSP directive blocks a service worker from being installed. This adds a note to explain what needs to be changed in the CSP to accomodate a service worker.

### DIFF
--- a/pages/content/amp-dev/documentation/guides-and-tutorials/optimize-measure/secure-pages.md
+++ b/pages/content/amp-dev/documentation/guides-and-tutorials/optimize-measure/secure-pages.md
@@ -24,7 +24,7 @@ Implement a CSP by adding the appropriate meta tag to the head of your pages. Be
 
 [You can view the full example here](https://github.com/ampproject/amphtml/blob/master/examples/csp.amp.html).
 
-Please note that if you are using a service worker (via [`amp-install-serviceworker`](https://amp.dev/documentation/components/amp-install-serviceworker/)) then the above CSP will block the service worker script from being installed. In such case, you'll need to add the URL to the service worker script to the allowed list of `script-src` URLs.
+Please note that if you are using a service worker (via [`amp-install-serviceworker`](../../components/reference/amp-install-serviceworker.md)) then the above CSP will block the service worker script from being installed. In such case, you'll need to add the URL to the service worker script to the allowed list of `script-src` URLs.
 
 [tip type="read-on"]
 Learn more about [protecting about security vulnerabilities and CSPs here](https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP).
--- a/pages/content/amp-dev/documentation/guides-and-tutorials/optimize-measure/secure-pages.md
+++ b/pages/content/amp-dev/documentation/guides-and-tutorials/optimize-measure/secure-pages.md
@@ -24,6 +24,8 @@ Implement a CSP by adding the appropriate meta tag to the head of your pages. Be
 
 [You can view the full example here](https://github.com/ampproject/amphtml/blob/master/examples/csp.amp.html).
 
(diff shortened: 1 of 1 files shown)


CrystalOnScript (approved):
Thanks!

westonruter:
Oh hold on. It turns out that this is not even valid AMP!

image

In order for this to work in AMP, you have to send it as a `Content-Security-Policy` HTTP response header and _not_ as a `meta[http-equiv]` tag.

This CSP example is also invalid: https://github.com/ampproject/amphtml/blob/master/examples/csp.amp.html

westonruter:
I also realize that the CSP will no longer work as-is with the new `script[amp-error]`. See https://github.com/ampproject/amphtml/issues/22543#issuecomment-673603872 and https://github.com/ampproject/amp-toolbox/issues/1146.

patrickkettner:
@kristoferbaxter is there anyone on the eng team that would be able to give best practices for CSP?

westonruter:
As for `script[amp-onerror]`, the CSP is indicated in https://github.com/ampproject/amphtml/issues/22543#issuecomment-673603872.

twifkak:
This PR LGTM, but I suggest some additional actions per the comments here.

To start, maybe update the above to what the Google AMP Cache outputs now for AMP documents:

```
$ curl -si https://amp-dev.cdn.ampproject.org/c/s/amp.dev/ | grep -i content-security-policy | cut -d\  -f2-
default-src * blob: data:; script-src 'sha256-5CxqAdDXlHviOy7zxeRpMobzRK/JNpLvkS+k8Zj3L3A=' 'sha256-FIBGC/wl1Qfnh2Fb5NPFHmRty7BHJdDpWW1FZ8egppI=' 'sha256-UXYprBCAtnqoL5acf14iemip/+HI+gDFh92yyXkM3XI=' 'sha256-dKn2nAtwgzaaXC8ZM58hhldxNyeuu4qrzW4H9//9YMA=' 'sha256-yAAlWuem9ue55JEvxkWhcWWA1Zu0p6cgbYtDWJjsdvs=' blob: https://cdn.ampproject.org/lts/ https://cdn.ampproject.org/rtv/ https://cdn.ampproject.org/sw/ https://cdn.ampproject.org/v0.js https://cdn.ampproject.org/v0.mjs https://cdn.ampproject.org/v0/ https://cdn.ampproject.org/viewer/; object-src 'none'; style-src 'unsafe-inline' https://cdn.ampproject.org/rtv/ https://cdn.materialdesignicons.com https://cloud.typography.com https://fast.fonts.net https://fonts.googleapis.com https://maxcdn.bootstrapcdn.com https://p.typekit.net https://pro.fontawesome.com https://use.fontawesome.com https://use.typekit.net; report-uri https://csp.withgoogle.com/csp/amp
```

This should cover the amp-onerror case. (I'm not sure what the other sha256's are about, off-hand.)

I also recommend a follow-up PR to fix https://github.com/ampproject/amphtml/blob/main/examples/csp.amp.html.

Further future:
- Looks like updating these two docs should become part of the process for changes that require updating the CSP on the AMP Caches.
- Maybe we should recommend a place to follow changes to the CSP that is required by optimized AMP. (It might change its value of `amp-onerror` one day?) This would also helpfully pick up any additions to the font allowlist that may occur for AMP (optimized or not).

twifkak:
Also, maybe we should recommend that publishers replace `report-uri` with one that they can monitor.

twifkak:
OK, looked into the various sha256s. I think all can be omitted except `5Cxq...` and `UXYp...` which are for the onerror handler, pre-and-post mod/nomod, respectively.

westonruter:
> This PR LGTM, but I suggest some additional actions per the comments here.

Also, the page needs to be updated to advise on sending a `Content-Security-Policy` HTTP response header rather than a `meta[http-equiv="Content-Security-Policy"]` since the latter is invalid AMP, yeah?

twifkak:
> Also, the page needs to be updated to advise on sending a `Content-Security-Policy` HTTP response header rather than a `meta[http-equiv="Content-Security-Policy"]` since the latter is invalid AMP, yeah?

Oh oops, yeah.

CrystalOnScript (approved):

Issue #5614, opened 2021-03-31
Disable tests for PRs not changing tested files


matthiasrohmer:
PRs that don't apply to any of the test suites have dangling yellow `test` check. Like this one: https://github.com/ampproject/amp.dev/pull/5612

patrickkettner:
fixed in #5616

PR #5616, opened 2021-03-31
add default workflow to run tests

patrickkettner:
PRs like #5612 are currently hung because we require a `test` job to pass in checks

image

but the existing workflows all filter on filename.

This PR adds a default workflow to run `npm test` on all PRs as the requisite `test` job. Since this should be run on [`pre-push`](https://github.com/ampproject/amp.dev/blob/future/.husky/pre-push#L4), this shouldn't ever really fail, but folks can force push a [`--no-verify`](https://git-scm.com/docs/git-commit#Documentation/git-commit.txt---no-verify) to get around it locally, so 🤷‍♀️.

If it takes up too much time, we can change the `run` [bit](https://github.com/ampproject/amp.dev/commit/3aa34ffc8a5fa2b9d275c2aa3df9aa24174c2e12#diff-1b8368a50c1714fe9faf0dfd1bf766142f1f0bb1eedd48540fb4bdde0323f50dR29) to something like `exit 0` and it should work, just be...hacky.

### DIFF
--- a/.github/workflows/test-default.yml
+++ b/.github/workflows/test-default.yml
@@ -1,5 +1,5 @@
 ---
-name: 'Default test job'
+name: 'Test: Default'
 
 on:
   push:
@@ -12,4 +12,3 @@ jobs:
       - name: 'Default'
         run: |
           echo 'we require a job named "test", but not all pull requests trigger an action. So we created this small silly job named "test" that will always pass, just to make sure nothing hangs as a result of not triggering an actual job.'
-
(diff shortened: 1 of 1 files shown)


matthiasrohmer:
If it's one me I'd be fine with just running `exit 0` as things we can actually test are tested with the path matching in place. Because now the problem would come the other way round: tests might be broken for a reason not related to a PR that didn't change anything that is actually matched by the tests. 🙂 

However this would fix https://github.com/ampproject/amp.dev/issues/5614 and I thought about the same approach and couldn't come up with something better.

patrickkettner:
updated to just echo a message explaining why we are doing nothing 🙃

matthiasrohmer (approved):
Thanks Patrick!

PR #5620, opened 2021-03-31
dont use ssr templating in survey script url

patrickkettner:


frontend/templates/views/partials/survey.j2
```diff
@@ -12,7 +12,7 @@
     </script>
   </amp-state>
 
-  <amp-script src="<% hosts.platform %>/static/survey.js" width="450" height="1205"  id="fez-container" sandbox="allow-forms">
+  <amp-script src="{{ hosts.platform }}/static/survey.js" width="450" height="1205"  id="fez-container" sandbox="allow-forms">
```
matthiasrohmer, line 15: In this context you are actually looking for `{{ podspec.base_urls.platform }}` - sorry, it's a mess.

matthiasrohmer (changes requested):

PR #5659, opened 2021-04-08
serve optimized AMP to packager

sebastianbenz:


matthiasrohmer:
What's the intention here? Doesn't the packager optimize pages using the Go optimizer?

sebastianbenz:
This is for testing. amppackager should support optimised AMP as input, but there was a problem when using the mjs runtime.

matthiasrohmer (approved):
I see 🙂 then go for it.

patrickkettner (commented):
👍

PR #5662, opened 2021-04-08
update thumbor URL in prod

patrickkettner:


matthiasrohmer:
That's the question. And why does it never happen for the packager. 

My gut feeling is the thumbor instance is still dying for unknown reasons and hence getting replaced.

patrickkettner:
@sebastianbenz no idea, i just was checking logs on prod and it was spitting out 404s for the thumbor urls

PR #5672, opened 2021-04-09
✅ Add E2E tests for Pixi

matthiasrohmer:
CWV and Mobile Friendliness tests are still a little flaky; need to look into this. Also it takes aaaages...


sebastianbenz:
Nice!

matthiasrohmer:
This works now. The Mobile Friendliness check currently fails with `testStatus: {status: "PAGE_UNREACHABLE"}` for amp.dev - hence I allowed `Analysis failed` to be a valid result.

sebastianbenz (approved):
Nice! How long does the test take to run?

matthiasrohmer:
If all tests succeed then 60-80s. +30s for fails due to the fairly big timeout.

PR #5686, opened 2021-04-13
🔧 Increase timeouts for Pixi E2E tests

matthiasrohmer:
The newly introduced tests currently fail inside the CI: https://github.com/ampproject/amp.dev/runs/2331487733?check_suite_focus=true#step:9:181

Timeouts being too small is more than a wild guess than an exact idea what's wrong though.

### DIFF
--- a/smoke-tests/pixi/Run.e2e-test.js
+++ b/smoke-tests/pixi/Run.e2e-test.js
@@ -42,6 +42,7 @@ describe('Pixi', () => {
     await expect(page).toClick('#input-submit');
     await expect(page).toMatchElement('#status-intro-banner-loading', {
       visible: true,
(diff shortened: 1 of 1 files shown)


smoke-tests/pixi/Run.e2e-test.js
```diff
@@ -42,6 +42,7 @@ describe('Pixi', () => {
     await expect(page).toClick('#input-submit');
     await expect(page).toMatchElement('#status-intro-banner-loading', {
       visible: true,
+      timeout: 60 * 1000,
```
sebastianbenz, line 45: please extract into a variable

sebastianbenz (approved):
LGTM with one nit

PR #5709, opened 2021-04-19
Disable amppackager

sebastianbenz:
The current version of amppackager is curently not compatbile to
optimized AMP + mjs runtime.

### DIFF
--- a/platform/lib/platform.js
+++ b/platform/lib/platform.js
@@ -170,7 +170,7 @@ class Platform {
 
   _configureRouters() {
     this.server.use(routers.cspReport);
-    this.server.use(routers.packager);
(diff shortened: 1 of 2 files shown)


platform/lib/platform.js
```diff
@@ -170,7 +170,7 @@ class Platform {
 
   _configureRouters() {
     this.server.use(routers.cspReport);
-    this.server.use(routers.packager);
+    // this.server.use(routers.packager);
```
patrickkettner, line 173: Is there a reason you not just removing it?￼

patrickkettner (approved):

sebastianbenz:
It'll be re-added soon.

>

PR #5718, opened 2021-04-20
Granular consent sample

morsssss:
I did get an assertion failure while running pre-commit tests:

```
FAIL: test_attributes (amp_example_preview.util.example_document_test.SourceCodeExporterTestCase)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/morss/Sites/docs/pages/extensions/amp_example_preview/util/example_document_test.py", line 33, in test_attributes
    self.assertEqual('0.1', example_doc.imports[1].version) # 0.1 is always the default
AssertionError: '0.1' != '1.0'
```

I imagine this is trying to tell me that I used v1.0 of a component instead of v0.1. As far as I know, though, I didn't. Maybe it's hitting something else in the repo, but, if so, the error doesn't tell me where.

Also... for some reason the reprompt UI isn't showing up on the sample page, because `amp-consent` doesn't remove the `hidden` attribute after the dialog is dismissed. I'm not sure why. Interestingly this is not a problem in the playground version - so I suspect it has to do with the HTML that the sample page surrounds the `amp-consent` with.

### DIFF
--- a/examples/source/user-consent/Granular_User_Consent.html
+++ b/examples/source/user-consent/Granular_User_Consent.html
@@ -1,0 +1,294 @@
+<!---
+preview: true
+author:
+- morss
+--->
+
+<!--
+  ## Introduction
+
+  *Granular consent* allows users to make a set of consent choices on your page. *Global consent* allows them to make a single choice for the page. The `amp-consent` component allows you to set up granular consent, global consent, or both simultaneously. In this sample, we'll show you how to use granular consent.
+-->
+<!-- -->
+<!doctype html>
+<html ⚡ lang="en">
+
+<head>
+  <meta charset="utf-8">
+  <title>Granular User Consent</title>
+  <link rel="canonical" href="self.html">
+  <meta name="viewport" content="width=device-width,minimum-scale=1,initial-scale=1">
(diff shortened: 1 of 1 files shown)


morsssss:
@micajuine-ho I don't know if you've ever built amp.dev, but I'm not sure why the reprompt `<div>` remains `hidden` on this page.

examples/source/user-consent/Granular_User_Consent.html
```diff
@@ -0,0 +1,328 @@
...
+    <!-- Then, you'll create an `<amp-consent>` component containing the following:
+  - a JSON configuration object
+  - the consent UI, which lets the user make consent choices
+  - optionally, a reprompt UI, which allows the user to show the consent UI again
+-->
+    <!--
+  ## Configuration object
+  [The documentation](https://amp.dev/documentation/components/amp-consent/#consent-configuration) lists all the keys that this object can contain. Here, we've used these:
+
+- [`consentInstanceId`](https://amp.dev/documentation/components/amp-consent/#consent-instance-id). This is required.
```
sebastianbenz, line 143: Super small nit, I'd prefer a colon instead of a dot here.

sebastianbenz (approved):
Great example!

PR #5723, opened 2021-04-22
fix the 404 page

### LINKED ISSUE #5722: 404 page doesn't show up
Type into your browser a bogus URL like `https://amp.dev/idonotexist`.

You'd expect to get a 404 page. Instead, the server response just hangs, then throws a 502.

image

patrickkettner:
fixes #5722


sebastianbenz (approved):
Thanks! Did you check whether we make the same mistake somewhere else?

PR #5743, opened 2021-04-27
add libyaml/pyYAML information to README, and make it a part of the default installation setup

patrickkettner:
doing builds locally sped up ~500% when it was added, should not make it seem quite as unimportant in the README

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,8 +75,8 @@ We welcome contributions to amp.dev.
 
   **Mac**
   ```sh
-    LDFLAGS="-L$(brew --prefix)/lib" CFLAGS="-I$(brew --prefix)/include" pipenv run pip3 install --global-option="--with-libyaml" --force pyyaml
-    pipenv run pip3 install --user grow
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -65,14 +65,23 @@ We welcome contributions to amp.dev.
...
       ```
 
 1. Install [Grow](http://grow.io), the static site generator used to build amp.dev. Do so by using `pip` instead of its installer. Using `pip` will enable importing from the `grow` package in Python later on.
 
    **Note**: Be sure to use the `pip` command associated with Python 3 as Grow 1 depends on Python 3.
 
+
+  **Mac**
+  ```sh
+    LDFLAGS="-L$(brew --prefix)/lib" CFLAGS="-I$(brew --prefix)/include" pipenv run pip3 install --global-option="--with-libyaml" --force pyyaml
```
sebastianbenz, line 78: pipenv is never mentioned before

matthiasrohmer: Also when using pipenv `pip` commands should never be run with either `pipenv run` or `pipenv shell` but `pipenv install` should be used instead 🙂

patrickkettner: apologizes, that was copied straight from the [grow page](https://grow.io/libyaml/), and was left in mistakenly

PR #5766, opened 2021-05-04
✨ About Websites: Stage

matthiasrohmer:
localhost_8080_about_websites-2021_(iPad) (1)
localhost_8080_about_websites-2021_(iPad)
localhost_8080_about_websites-2021_(iPhone 6_7_8)


frontend21/scss/amp-dev.scss
```diff
@@ -4,6 +4,8 @@
 @import '../../frontend/scss/components/atoms/_color.scss';
 
 @import 'variables.scss';
+@import 'variables.scss';
```
lluerich, line 7: Please remove the duplicate import.

lluerich (approved):
One nit, otherwise LGTM 🚀

frontend21/scss/amp-dev.scss
```diff
@@ -70,15 +71,18 @@ h5,
 h6 {
   margin-top: 0;
   margin-bottom: 0.3em;
+  color: currentColor;
```
patrickkettner, line 74: why is this being done? [per spec](https://drafts.csswg.org/css-color/#currentcolor-color)

```
If the ‘currentColor’ keyword is set on the ‘color’ property itself, it is treated as ‘color: inherit’.
```

Its technically not invalid, but just sorta confusing. Are you wanting to inherit the value but not cascade it?

pages/content/amp-dev/about/websites-2021.html
```diff
@@ -0,0 +1,110 @@
...
+          <h2 class="ap-h1">{{ _('Guaranteed Page Experience') }}</h2>
+          <p>
+            {{ _('AMP provides you with the tools needed to build websites that
+            meet Page Experience ranking signals while remaining flexible and
+            customizable to your needs. Page experience is a set of signals from
+            Google that measure how users perceive their experience interacting
+            with a web page beyond its pure information value. AMP is a
+            straightforward way to a high page experience score. Domains built
+            with AMP are 5 times more likely to pass* Core Web Vitals metrics
+            than non-AMP!') }}
```
patrickkettner, line 58: Would you be able to add these strings to the [localization files](https://github.com/ampproject/amp.dev/tree/future/pages/translations), so they can be picked up automatically for localization?

frontend21/scss/amp-dev.scss
lluerich: True, it makes sense to use ```color: inherit;``` then, unless @matthiasrohmer had some reason to implement it that way.

pages/content/amp-dev/about/websites-2021.html
lluerich: Done.

patrickkettner: Awesome! Does grow recognize it still with the whitespace being used here? I am foggy on if it can or not

lluerich: To be save I took them out.

jaygray0919:
Does this version use the single-menu design used on
  https://amp.dev/about/page-experience/

lluerich:
@jaygray0919 Yes, it does use the same template/structure as the page-experience page.

jaygray0919:
tres tres bien!
the 'ole 2-menu-solution was double trouble
we followed the page-experience page here: https://ontomatica.io/a/12370110501010001950/
reduced the maintenance expense **dramatically**
(site still under-construction, but language menus work)

jaygray0919:
@lluerich is there a schedule for putting this new design in production?

PR #5768, opened 2021-05-04
Fix hide surveys when they are inside of amp-next-page

sebastianbenz:
Remove CSS as targeting i-amphtml-* is invalid AMP. Use amp-next-page's
built-in `next-page-hide` element instead to make sure that surveys are
not repeated.


sebastianbenz:
Force merge to unblock build.

PR #5772, opened 2021-05-04
🐛 Optimize inline example previews to add component scripts

### LINKED ISSUE #5761: Inspect component imports issued by [example] tags
Surfaced with https://github.com/ampproject/amphtml/pull/34110/. Interestingly almost all examples in the component docs explicitly define their imports, like for amp-twitter: `[example preview="inline" playground="true" imports="amp-twitter" imports="amp-twitter"]` amp-accordion didn't until #34110 but actually the required dependency should have get injected automatically.

A lot of the bento examples also don't specify their imports, like [amp-social-share](https://amp.dev/documentation/components/amp-social-share-v1.0/?format=websites) as well. Here the dependencies seem to get injected correctly though the example still fail due to CSP errors and certain experiments required to be enabled.

This needs further investigation.

matthiasrohmer:
Fix #5761.

### DIFF
--- a/platform/lib/routers/inlineExamples.js
+++ b/platform/lib/routers/inlineExamples.js
@@ -22,20 +22,30 @@ const {
   getFormatFromRequest,
   DEFAULT_FORMAT,
 } = require('../amp/formatHelper.js');
+const {promisify} = require('util');
+const fs = require('fs');
+const readFileAsync = promisify(fs.readFile);
+const {optimize} = require('@lib/utils/ampOptimizer.js');
+const path = require('path');
 
 // eslint-disable-next-line new-cap
 const inlineExamples = express.Router();
-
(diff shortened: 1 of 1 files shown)


sebastianbenz (approved):

sebastianbenz:
Thanks! Nice and dirty :-)

sebastianbenz:
BTW - are these cached?

matthiasrohmer:
Huh... you asking made me give this a second thought: they are not. I thought the CDN would take care of this but those are also affected by the `vary: AMP-Cache-Transform` problem and therefore are not.

I'll go ahead and add caching as in:

https://github.com/ampproject/amp.dev/blob/ab711ad980fdfc6c5a7ef2c5049e3163cfcf69ca/platform/lib/routers/growPages.js#L247

sebastianbenz:
👍

>

PR #5778, opened 2021-05-05
update upstream docs

patrickkettner:
update docs to recent changes


sebastianbenz:
These look like older versions as URLs point to master instead of main.

patrickkettner:
gah, sorry. I ran `npm run bootstrap` and read the diff backwards (I thought it was adding `main`, not `master`). 

Now to figure out why bootstrap is doing that...

PR #5780, opened 2021-05-05
dashes, not underscores

### LINKED ISSUE #5775: Playground suggesting invalid layout
### Current behavior
Playground is suggesting `fixed_height` instead of `fixed-hight` which is an invalid layout attribute value - same with `flex_item/flex-item`

Screen Shot 2021-05-04 at 12 13 37 PM

patrickkettner:
fixes #5775


patrickkettner:
fwiw, these have been in there since @filip added it back on amp by example, not sure if there was ever a practical reason as to why

sebastianbenz:
I don't think so...this list might be outdated as well

@fstanis how did you generate this?

matthiasrohmer (approved):

PR #5785, opened 2021-05-06
🚧  New homepage

lluerich:
Work in progress.


matthiasrohmer (commented):
No major remarks.

frontend/icons/logo-outline.svg
```diff
@@ -0,0 +1,13 @@
+<symbol id="logo-outline" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 194 194">
```
matthiasrohmer, line 1: Inline this SVG as it's one-off use currently. There also is a new icon mechanism in place that uses an external and hence cacheable SVG document.

jaygray0919: @matthiasrohmer is there documentation for `new icon mechanism in place that uses an external and hence cacheable SVG document`? Is that in place of using a sprite structure? We control color-reversing and gradients from CSS; which is the only way we know how to do it. We're not aware of a way to control an external SVG using `<style-amp>` elements. But perhaps you are not doing any special treatment for `logo-outline`. But other amp-menu symbols (SVG) are in a sprite, nu?

PR #5828, opened 2021-05-12
Fix bento component-version.json generation 

sebastianbenz:


platform/lib/pipeline/componentReferenceImporter.js
```diff
@@ -198,8 +199,9 @@ class ComponentReferenceImporter {
     }
 
     return Promise.all(
-      documents.map((doc) => {
-        return this._createGrowDoc(doc);
+      documents.map(async (doc) => {
```
patrickkettner, line 202: any concern about this failing? Or would the function calling the promise be catching that?

patrickkettner (approved):
one nit/question, otherwise lgtm

platform/lib/pipeline/componentReferenceImporter.js
sebastianbenz: good catch, forgot to clean this up

Issue #5777, opened 2021-05-04
Thumbor requests fail


sebastianbenz:
Haven't had time yet to investigate. The instance id is correct though.

```
{
  "insertId": "r6nq0pork2fknp549",
  "jsonPayload": {
    "cos.googleapis.com/container_id": "3f017a3c997fc40af2086f58bd405ac695ba9df022733af2623eeb16021fbf0c",
    "cos.googleapis.com/stream": "stdout",
    "message": "[9:40:45 PM] [Thumbor] › ✖  error     Failed connecting to thumbor FetchError: request to http://ig-amp-dev-thumbor-n0gc.c.amp-dev-230314.internal:8080/unsafe/https://amp.dev/static/samples/img/image2.jpg%3Foriginal=true failed, reason: getaddrinfo ENOTFOUND ig-amp-dev-thumbor-n0gc.c.amp-dev-230314.internal\n"
  },
  "resource": {
    "type": "gce_instance",
    "labels": {
      "instance_id": "6385818354535104611",
      "zone": "europe-west2-c",
      "project_id": "amp-dev-230314"
    }
  },
  "timestamp": "2021-05-04T21:40:45.692615377Z",
  "labels": {
    "compute.googleapis.com/resource_name": "f237494a92cd"
  },
  "logName": "projects/amp-dev-230314/logs/cos_containers",
  "receiveTimestamp": "2021-05-04T21:40:50.815806749Z"
}
```

matthiasrohmer:
Insert inappropriate language here... I'll have a look tomorrow.

sebastianbenz:
As part of this we should also make `/static/remote/?url=...` fallback gracefully to the actual image.

sebastianbenz:
This problem also affects amppackager. Both, amppackager and thumbor, instances currently cannot be reached even though instanceids are correct ([logs](https://console.cloud.google.com/logs/viewer?project=amp-dev-230314&rif_reserved&minLogLe[%E2%80%A6]%3A23.031073264Z%22%0AinsertId%3D%22pnecxhqub5322thip%22). 

//cc FYI @patrickkettner

patrickkettner:
I shelled into the docker containers for the site to watch for logs, and they were both getting an ENOTFOUND error for the both thumbor and packager. I jumped over to each of the containers, and got their hostname with 

```sh
curl "http://metadata.google.internal/computeMetadata/v1/instance/hostname" \
  -H "Metadata-Flavor: Google"
```

[...]

PR #5831, opened 2021-05-12
Thumbor: fallback for remote images

sebastianbenz:
Gracefully fail if thumbor is not available for remote images. See https://github.com/ampproject/amp.dev/issues/5777#issuecomment-832694224

### DIFF
--- a/platform/lib/routers/thumbor.js
+++ b/platform/lib/routers/thumbor.js
@@ -58,7 +58,8 @@ thumborRouter.get(imagePaths, async (request, response, next) => {
 
   // We allow certain remote images to be optimized;
   // they mount on the virtual /static/remote
-  if (request.url.includes(REMOTE_STATIC_MOUNT)) {
+  const isRemoteImage = request.url.includes(REMOTE_STATIC_MOUNT);
+  if (isRemoteImage) {
     imageUrl = new URL(request.query.url);
   } else {
     imageUrl.searchParams.set('original', 'true');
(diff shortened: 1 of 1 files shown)


platform/lib/routers/thumbor.js
```diff
@@ -68,24 +69,40 @@ thumborRouter.get(imagePaths, async (request, response, next) => {
...
+          res.status(404).send('Not found');
+          return;
+        }
+      } catch (e) {
+        log.error('Failed downloading remote image', e);
+        res.status(404).send('Not found');
+        return;
+      }
+    } else {
+      // If Thumbor did not respond, fail over to default static middleware
```
patrickkettner, line 98: Is that what this is doing? It looks like it would return the 404 if thumbor has a failure. Are we relying on the upstream server to determine a time out?￼ Or am I just not seeing the larger picture of how this fits in￼

sebastianbenz: For local images we fall back to the static middleware and serve the large version of the image (this works already). What's new is that we also handle the case where we use Thumbor to optimize remote images from the blog. In that case we proxy the request.

patrickkettner: Ah, so in the case of an error in thumbor with remote images we send a 404, rather than 302 to the origin URL or whatever?

patrickkettner (approved):

matthiasrohmer (approved):

PR #5837, opened 2021-05-12
Refactor grap and go component

dlemm:
- Remove style duplicates
- add condition to `amp-animation` example in order to render translatable example with markup in example code.


dlemm:
@sebastianbenz The changes to the `animation`-example has been pushed a minute too late. So here is another PR for adding the markup to `animation`-example.

lluerich (approved):
LGTM

PR #5839, opened 2021-05-12
redirect instead of proxy

sebastianbenz:
Patrick was of course right! Redirect is the better choice here.


patrickkettner (approved):
Just to confirm, we don't want to even touch remote images with thumbor?￼ As opposed to trying thumbor and then failing back to a redirect￼. If so, then LGTM

sebastianbenz:
By default thumbor will optimize remote images, but if thumbor is unavailable we redirect to the remote image.

PR #5844, opened 2021-05-13
    include zone in the instance information URLs for thumbor and amppackager

patrickkettner:


patrickkettner:
cc @sebastianbenz @matthiasrohmer, merging to get prod up to date

sebastianbenz:
Awesome! Thanks for looking into this!

PR #5846, opened 2021-05-14
Revert "✨  Overwrite home and about page for default language with ne…w versions (#5841)"

patrickkettner:
build is failing currently, merging to unblock before debugging this


patrickkettner:
cc @matthiasrohmer

PR #5867, opened 2021-05-18
Add Statamic to the CMS list

jackmcdade:
Statamic supports AMP and is used on huge sites like https://spiegel.de and https://opentable.com.


sebastianbenz (approved):
Thanks!

PR #5870, opened 2021-05-18
🎨 Move serving path determination to component ref importer

### LINKED ISSUE #5830: Bento components are listed as latest component
For example: https://amp.dev/documentation/components/amp-timeago/?format=websites

The problem is that validation rules now include the bento version as the latest version even though they're experimental. As a side effect, the link from the non-bento version to the bento version is also broken as it assumes a versioned doc URL.

matthiasrohmer:
Fixes #5830, tested with the provided `amp-timeago` case while observing `amp-accordion` (where the bento version is not specified as latest in the validation rules) and `amp-carousel` which doesn't have a bento but a regular 0.2 version.

This change is necessary as one needs to know about all existing versions and the `experimental` flag specified in the frontmatter.


matthiasrohmer:
@sebastianbenz, if you want to test this locally make sure to delete `~www-amp-dev/pages/.grow/routescache.json`. Then run a fresh import with `node platform/lib/pipeline/componentReferenceImporter.js --queue-imports`.

sebastianbenz:
Nice! One more things we need to fix: The version dropdown no longer works in this case, e.g. select v0.1 [here](http://localhost:8080/documentation/components/amp-accordion-v1.0/?format=websites).

sebastianbenz:
@matthiasrohmer should we merge this to get the first fix in?

matthiasrohmer:
Now merged with #5879.

matthiasrohmer:
Blocked by #5913.

sebastianbenz:
Good to merge now?

matthiasrohmer:
Yup, checks are passing!

PR #5877, opened 2021-05-19
✨ Add text-aside component to new Get Started page

lluerich:


frontend/templates/views/2021/partials/content/link.j2
```diff
@@ -0,0 +1,13 @@
+<a class="ap-link{% if square %} --square{% endif %}"
```
matthiasrohmer, line 1: Instead of having to manually add the `svg` elements can we instead use an `:before` element with `.--internal` and `.--external` modifier classes?

The thing is I don't want contributors need to care over how icons are used on amp.dev but things to just work.

lluerich: Please have another look. ```.--internal``` is the default. ```.--external```, ```.--square``` and ```.--icon``` are optional modifiers.

frontend/templates/views/2021/partials/content/link.j2
```diff
@@ -0,0 +1,10 @@
+<a class="ap-link{% if external %} --external{% else %} --internal{% endif %}{% if square %} --square{% endif %}{% if icon %} --icon{% endif %}"
+  href="{{ href }}">
+  {% if icon %}
+    <div class="ap-icon ap-link-icon">
```
matthiasrohmer, line 4: Do we ever use links in this form with an icon other than the specified errors?

matthiasrohmer (approved):
Looks good, one last question 🙂

frontend/templates/views/2021/partials/content/link.j2
lluerich: I think you mean arrows, not errors? :) 
I guess I was mislead, since the class ```ap-m-lnk-icon``` is used in other parts of the page (e.g. https://amp.dev/community/working-groups/infra/ -> channel links in the communication channels section).

I will remove it.

matthiasrohmer: Whoops, yes 😁 thanks!

PR #5879, opened 2021-05-19
Fix component auto import

sebastianbenz:
This fixes the problem that the validation rules include experimental
component versions causing bento components to be auto imported.

Instead of generating the component versions from validation rules, we
extract the latest version from the component docs.

This PR also removes the buildComponentVersion task as it is no longer
needed.


sebastianbenz:
@matthiasrohmer build is passing now, not super happy about commiting the component versions file, but I don't see another way.

pages/content/amp-dev/about/websites-2021.html
```diff
@@ -112,7 +112,7 @@ <h2 class="ap-h1">{{ _('Grab-and-go components') }}</h2>
         <p>
           {{ _('Want just the components without the AMP ecosystem? ')}}
           <a
-            href="{{ g.doc('/content/amp-dev/documentation/guides-and-tutorials/start/bento_guide', locale=doc.locale).url.path }}"
+            href="{{ g.doc('../documentation/guides-and-tutorials/start/bento_guide.md', locale=doc.locale).url.path }}"
```
patrickkettner, line 115: im assuming relative URLs don't break anything?

pages/extensions/amp_example_preview/util/example_document_test.py
```diff
@@ -43,7 +43,7 @@ def test_different_attributes(self):
     self.assertEqual('none', example_doc.preview)
     self.assertEqual(1, len(example_doc.imports))
     self.assertEqual('amp-accordion', example_doc.imports[0].name)
-    self.assertEqual('1.0', example_doc.imports[0].version) # 0.1 is always the default
```
patrickkettner, line 46: lol

pages/static/files/component-versions.json
```diff
@@ -0,0 +1,132 @@
+{
```
patrickkettner, line 1: is this just going to be manually updated? What calls componentReferenceImporter to update it?

patrickkettner (approved):
not clear on some stuff that I ask about, but looks good!

pages/static/files/component-versions.json
matthiasrohmer: `componentReferenceImporter` is called with with every build and by running `npx gulp bootstrap` or `npx gulp importAll` for devs running amp.dev locally.

patrickkettner:
just as a devil's advocate, did this issue get discussed with the validator folks? I assume they had no workarounds?

pages/content/amp-dev/about/websites-2021.html
matthiasrohmer: Yeah that should be fine.

matthiasrohmer (approved):
Hm. I liked having the validator rules as a single point of truth but they have always been tedious to work with... no better idea.

PR #5888, opened 2021-05-20
fix playground amp-script hash generation

### LINKED ISSUE #5887: Script hash generation broken in playground
`Uncaught ReferenceError: Buffer is not defined` in

```
calculateHash: function(t, {algorithm: e="sha384"}={}) {
                    const n = e.toLowerCase();
                    if (!i.has(n))
                        throw Error("Unsupported algorithm for CSP: " + n);
                    return "string" == typeof t && (t = Buffer.from(t, "utf8")),
                    `${n}-${function(t) {
                        return t.replace(/=/g, "").replace(/\+/g, "-").replace(/\//g, "_")
                    }(r.createHash(n).update(t).digest("base64"))}`
```

See https://github.com/ampproject/amphtml/issues/34469

patrickkettner:
fixes #5887

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -16,6 +16,7 @@
 				"@ampproject/toolbox-optimizer": "2.8.0-canary.20",
 				"@google-cloud/datastore": "6.4.0",
 				"acorn": "8.2.4",
+				"buffer": "5.6.0",
 				"casual": "1.6.2",
 				"cheerio": "1.0.0-rc.3",
 				"client-sessions": "0.8.0",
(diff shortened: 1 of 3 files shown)


patrickkettner:
webpack 5 removed automatic polyfilling of node apis. fixing the Buffer error that was being logged exposed an issue where `createHash` didn't exist (`crypto-es`, which was being mapped to `crypto` doesn't provide it). This probably breaks something else, but I am not sure what is using `crpyto-es` fallbackj currently

sebastianbenz (approved):

kristoferbaxter:
Thanks for fixing so quickly!

PR #5926, opened 2021-05-31
🚑 Add file extension to component docs serving path

matthiasrohmer:
Compare [L456](https://github.com/ampproject/amp.dev/runs/2713031896?check_suite_focus=true#step:10:456) to L457. This slipped #5870's review.

Without the `.html` in the serving path Grow creates files without an extension which don't get picked up by `gulp.src`.


sebastianbenz (approved):
Good catch!

PR #5989, opened 2021-06-14
💥 Start working on production ready Docker image with page building at runtime

matthiasrohmer:
This is still far from finished, but by popular demand here is what I have so far. Doing a fresh clone from this branch and running `npx gulp build` (or `npx gulp build --queue-imports` if there are problems with the GitHub API) followed by

```
docker build . -t amp.dev
docker run --publish 7777:8080 -t amp.dev
```

Should start amp.dev on http://localhost:7777 without building all the pages first. In order to get this to work I needed to break some things, others I haven't looked into yet:

- Error pages: until now this was a static page, of course. And it should stay this way as redirecting to the Grow dev server is not an option because this could lead to endless loops if Grow itself has an error
- Static files: they 404 - probably just copied to the wrong location
- Sitemaps: they probably work, needs to be checked
- Image optimization: the image index generating the hashes isn't build/found
- Correct absolute URLs: this could already work, not quite sure if I catched all needed places
- Docker image size: it's still 2.4GB. Might be that I missed something in the .dockerignore, but node_modules alone are 746MB, collected static files are 429MB


matthiasrohmer:
Build context is only 736MB.

matthiasrohmer:
Sadly this project failed. I can't get Grow to run on app engine. There are random errors (currently `RouteContflictErrors`) I can't reproduce in any other environment.

I'll extract the good parts of this PR (removing unused deps, updating deps, condensing the build files, development dockerfile) into separate cleaned-up PRs.

sebastianbenz:
😢 thanks for looking into this!

PR #6001, opened 2021-06-17
header fixes for new frontend

patrickkettner:


sebastianbenz:
Thanks for looking into this Patrick! I've created https://github.com/ampproject/amp.dev/pull/6003 which fixes the root cause. Your PR gave me the right idea where to look.

PR #6008, opened 2021-06-21
🐛 Fix order of imports

### LINKED ISSUE #5998: subscribe link listed twice on homepage
image

[its here](https://github.com/ampproject/amp.dev/pull/5928/files#diff-78abf0c057f3f1f6f5c13fecd0e938472ee3c130895011443a137744eff586d5R141-R155) - is there something missing from the `with` @lluerich @matthiasrohmer  ?

lluerich:
Link styles overruled the special case for the newsletter due to the order in which they are imported.

Fixes #5998.


sebastianbenz:
Thanks for fixing!

Issue #5570, opened 2021-03-22
Amp Analytics CLS w/ GA add precision tip


micajuine-ho:
When tracking CLS using GA, they state that they can only collect integer values and so it is recommended that you multiply the collected CLS value by 1000 to get greater precision. 

https://github.com/GoogleChrome/web-vitals#using-analyticsjs

A tip [here](https://amp.dev/documentation/examples/advertising-analytics/tracking_core_web_vitals/?format=websites#tracking-core-web-vitals-in-amp-using-google-analytics), would be helpful.

Something along the lines of:

```
TIP: For greater precision, multiply the CLS value collected by 1000 to be sent to GA:

$CALC(${cumulativeLayoutShift}, 1000, multiply)
```

FYI: new $CALC macro was just added: https://github.com/ampproject/amphtml/pull/33383 which lead to this convo.

PR #6020, opened 2021-06-25
update analytics CWV docs

### LINKED ISSUE #5708: Fix event for CVW analytics
We currently recommend to track CWV on pageview, however, most values won't be available yet when the pageview event is fired. We need to update the sample with the correct approach:

https://amp.dev/documentation/examples/advertising-analytics/tracking_core_web_vitals/?format=websites&optimize=false

patrickkettner:
fixes #5708

### DIFF
--- a/examples/source/advertising-analytics/Tracking_Core_Web_Vitals.html
+++ b/examples/source/advertising-analytics/Tracking_Core_Web_Vitals.html
@@ -37,7 +37,9 @@ The `amp-analytics` component provides the following macros:
 * [Cumulative Layout Shift](https://web.dev/cls/): `${cumulativeLayoutShift}`
 * [First Input Delay](https://web.dev/fid/): `${firstInputDelay}`
 
-These macros can attach to any triggers and resolve asynchronously.
+While you can attach these macros to any trigger, they resolve individually and asynchronously.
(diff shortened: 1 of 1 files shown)


micajuine-ho (approved):
This looks great! Love the clarification--it should help devs understand best usages!

patrickkettner:
also fixes #5570

patrickkettner:
thanks @micajuine-ho!

sebastianbenz (approved):
Thanks for the update! One question for my understanding: when does the `visible` event trigger? Is sounds like it might fire too early for FID or CLS or am I misunderstanding it?

micajuine-ho:
> when does the visible event trigger? Is sounds like it might fire too early for FID or CLS or am I misunderstanding it?
 
Firing early is not a problem, since it'll wait for FID or CLS to resolve and then the request to be sent.

And if it fires after FID or CLS is calculated, it'll send out the request immediately.

patrickkettner:
It fires after `visible`, once the macros have resolved

cajotafer:
I need help, I don't understand how to finally see these events in GA, now that the example changed I am even more confused. Do you have any example about how to receive or see this in GA? 🤞

PR #6027, opened 2021-06-26
New amp-render samples page

morsssss:
Ok! I cut my more ambitious sample - showing how to do list-style pagination - because it turned out the approach I was taking with `amp-script` wouldn't have worked, as `amp-render` doesn't support the `refresh` action when data comes from `amp-script`. I could have done it in a different way, but I'm going to leave that for later.

I couldn't run pre-commit tests because some Pixi tests were failing. So this hasn't gone through the usual Prettier linting.

### DIFF
--- a/examples/source/1.components/amp-render.html
+++ b/examples/source/1.components/amp-render.html
@@ -1,0 +1,150 @@
+<!---
+standaloneSnippets: true
+preview: true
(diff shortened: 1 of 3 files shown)


sebastianbenz (approved):
Really nice!

PR #6032, opened 2021-07-02
Update Tracking_Core_Web_Vitals for better understanding

cajotafer:
Hi guys,
I have no experience with `amp-analytics` and looking for some docs to track core web vitals, I found this, but it was confusing for me. The last example is not related to the other examples.
I am not completely sure these changes are OK, but it makes sense to me and I just want it to make sense to other people.


sebastianbenz (approved):
Good catch! This makes a lot of sense.

sebastianbenz:
Thanks!

cajotafer:
Hi, 
I am fixing this error since the examples are using single quotes `'` instead of double quotes `"`, so it was not a valid json.

PR #6063, opened 2021-07-22
Specify .amp-carousel-button so it does not pollute rendered sample

alanorozco:
We're inadvertently modifying and breaking [rendered code samples.](https://amp.dev/documentation/examples/components/amp-lightbox-gallery/?format=websites)

image (3)

### DIFF
--- a/frontend/scss/base.scss
+++ b/frontend/scss/base.scss
@@ -26,8 +26,12 @@ html, body {
   }
 }
 
-/* Default amp-carousel button styley */
-.amp-carousel-button {
+/*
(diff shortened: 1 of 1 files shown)


alanorozco:
I should not that this is the only selector I've seen that pollutes the sample namespace, but I did not verify any others.

alanorozco:
I should note that this is the only selector I've seen that pollutes the sample namespace, but I did not verify any others.

sebastianbenz (approved):
Thanks a lot!

PR #6070, opened 2021-07-26
Fix invalid homepage

sebastianbenz:
Blog posts without title image did render without fallback image due to
mixing {{ }} and [= =] template literals by importing teaser.j2.

Brute force fix by duplicating teaser.j2 in index-2021.html.

### DIFF
--- a/frontend21/scss/components/content/teaser.scss
+++ b/frontend21/scss/components/content/teaser.scss
@@ -74,5 +74,24 @@
       color: color('river-bed');
       text-transform: capitalize;
     }
+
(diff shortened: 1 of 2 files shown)


sebastianbenz:
///cc @lluerich  @matthiasrohmer I'm sure there's a cleaner approach. 

@lluerich is this potentially also the reason why you couldn't get the carousel arrows to work correctly?

lluerich:
I'll check that, but I believe the reason for the arrows to not work correctly was them being outside of the carousel and only controlling it from the outside.

PR #6087, opened 2021-08-05
Specify namespace to avoid pollution, for real this time

alanorozco:
Fixes an issue where carousel button selectors for the host document pollutes rendered example snippets.

A followup to #6063, where I did not test that the exclusion selector actually worked. (I verified this time, apologies!)

This selector could be somewhat inefficient since it traverses up to `body`. Alternatively, we could scope it to the `amp-carousel` elements that we want by classname, but I'm not familiar with the parts of the `amp.dev` site that we would like to target.


sebastianbenz (approved):
Thanks!

PR #6115, opened 2021-08-27
Update plumrocket links due to website migration

06romix:
We at Plumrocket Inc have migrated our website and changed all links. So we would like to update our links on amp.dev to new ones.


andreylipattsev:
@CrystalOnScript @sebastianbenz could you take a look or route correctly when you can? Thanks!

matthiasrohmer (approved):
Thanks for the update!

Issue #6216, opened 2021-11-11
Playground: seeing `embedMode is undefined` error


samouri:
### Current behavior
When using the playground in mobile emulation mode, I am seen an error `embedMode is undefined` in the console.

Please also provide:
- Page link: https://playground.amp.dev/?url=https%3A%2F%2Fpreview.amp.dev%2Fdocumentation%2Fexamples%2Fcomponents%2Famp-lightbox-gallery&format=websites
- (Optional) Screenshots/videos:

![Screen Shot 2021-11-11 at 10 38 54 AM](https://user-images.githubusercontent.com/4656974/141326106-efd701d0-d65f-414c-8016-7f731ea46944.png)

### Reproduction procedure

Browser: Chrome

1. Open the provided URL in Chrome with Dev Tools Console optn
2. (maybe): enable mobile emulation
3. Notice the error
3.

PR #6218, opened 2021-11-11
playground bugfix

patrickkettner:
fix for #6216

Looks like when the modes stuff was updated in #4138, this wasn't.


sebastianbenz (approved):
Thanks!

PR #6269, opened 2022-01-25
Update language on courses and roadshow

morsssss:
Changing roadshows page to make it more clear that we're not doing roadshows right now.

And removing my email address and language about certification program from the courses area.


sebastianbenz (approved):
Thanks Ben!

PR #6327, opened 2022-03-03
Added email client FairEmail

M66B:


patrickkettner (approved):

M66B:
Thanks for merging this @patrickkettner

How does it work: emails with AMP content will be shown as usual, but with an AMP icon (the "flash") at the top right of the email, which allows users to switch to the AMP content with one tap.

For security reasons, the script location is restricted to the domain *cdn.ampproject.org* for now.

PR #6393, opened 2022-06-24
Fix bullet indentation level and link URLs

zhangsu:
1. Currently the indentation of the sub-bullets are done using 2 spaces,
   which apparently isn't enough to make them render as sub-bullets. We
   need 4 spaces instead.

2. Fix the `byte limit on the size of the email` link. It's missing
   `email-spec` in the path.

3. Make action and event and amp-bind links point to amp.dev links
   instead of GitHub.


zhangsu:
@sebastianbenz anyway to preview the changes with rendered Markdown? In hindsight, I shouldn't have skipped this when adding those changes in the first place...

sebastianbenz:
Thanks for the fixes! There's unfortunately no simple way to preview the rendered page. You need to setup a local amp.dev dev environment as described in the readme.

sebastianbenz (approved):

PR #6399, opened 2022-06-29
Fix relative links

zhangsu:
The link needs to point to a `.md` file, not the URL path when the
Markdown is rendered.

/to @sebastianbenz


sebastianbenz (approved):
Thanks!

PR #6472, opened 2022-10-12
create a static version of amp.dev

patrickkettner:


patrickkettner:
netlify bot hasnt been configured, so their failures are expected

patrickkettner:
ping @sebastianbenz

netlify/configs/amp.dev/netlify.toml
```diff
@@ -0,0 +1,754 @@
...
+
+[[redirects]]
+  from = "/documentation/courses/*"
+  to = "/documentation/courses/:level/:splat/?redirected"
+  status = 301
+  force=true
+  query = {level = ":level"}
+
+#################
+# Imported from platform/config/go-links.yaml
```
sebastianbenz, line 53: Is this manually maintained?

gulpfile.js/build.js
```diff
@@ -355,24 +387,63 @@ function buildPages(done) {
...
-    async function publishPages() {
-      if (!config.options.locales.includes(config.getDefaultLocale())) {
+    async function copyBuildFiles(done) {
+      if (!config.options?.locales?.includes(config.getDefaultLocale())) {
         console.log(
           'Skipping page publishing. Default language is not build, only:',
           config.options.locales
         );
-        return;
+        return done();
```
sebastianbenz, line 396: curiosity: do you still need to call `done` when using returning promises?

gulpfile.js/build.js
```diff
@@ -392,10 +463,404 @@ function buildPages(done) {
...
+      'A ship sailing the sea at sunset.',
+      'landscape_ship_1280x853.jpg',
+      10
+    ),
+  ];
+
+  const configObj = {
+    time: new Date().toLocaleTimeString(),
+    timestamp: Number(new Date()),
+    // send a random list of blog items to make it also work on the cache
```
sebastianbenz, line 592: does this still make sense?

gulpfile.js/build.js
```diff
@@ -392,10 +463,404 @@ function buildPages(done) {
...
+              const origURL = $this.attr('href');
+
+              $this.attr('href', getUpdatedURL(origURL, format));
+            });
+
+            const renderedPage = $.root().html();
+
+            if (
+              format === FORMAT_WEBSITES &&
+              !file.path.endsWith('tools.html')
```
sebastianbenz, line 686: add a comment to explain why not tools

gulpfile.js/build.js
```diff
@@ -392,10 +463,404 @@ function buildPages(done) {
...
+  const requestPathRegex = new RegExp(
+    `${project.paths.PAGES_DEST}|(index)?.html`,
+    'g'
+  );
+
+  const generatedFormats = SUPPORTED_FORMATS.map((format) => {
+    const f = (cb) => {
+      const env = nunjucksEnv();
+
+      const f = filter([`${project.paths.PAGES_DEST}/**/*html`, 'index'], {
```
sebastianbenz, line 638: nested variables with the same name are a bit confusing, can you find a better name?

gulpfile.js/build.js
```diff
@@ -392,10 +463,404 @@ function buildPages(done) {
...
+              }
+
+              file.contents = Buffer.from(result);
+              callback(null, file);
+            });
+          })
+        )
+        .pipe(
+          through.obj(async function (file, enc, callback) {
+            // Rewrite links inside of each of the pages
```
sebastianbenz, line 670: This rewrites only "known" format specific links, but format specific links in docs etc won't be rewritten, but rely on redirects - correct?

gulpfile.js/build.js
```diff
@@ -554,14 +1017,18 @@ exports.zipTemplates = zipTemplates;
...
 exports.unpackArtifacts = unpackArtifacts;
 exports.collectStatics = collectStatics;
+exports.generateWhoAmI = generateWhoAmI;
 exports.buildPixiFunctions = buildPixiFunctions;
 exports.buildFinalize = gulp.series(
   gulp.parallel(collectStatics, persistBuildInfo),
   thumborImageIndex
 );
 
+exports.zzz = renderExamples;
```
sebastianbenz, line 1030: is this used?

gulpfile.js/build.js
```diff
@@ -17,6 +17,7 @@
 'use strict';
```
sebastianbenz, line 17: This file did get quite large, is there a reasonable way to split it up?

gulpfile.js/build.js
```diff
@@ -355,24 +387,63 @@ function buildPages(done) {
...
+      if (!config.options?.locales?.includes(config.getDefaultLocale())) {
         console.log(
           'Skipping page publishing. Default language is not build, only:',
           config.options.locales
         );
-        return;
+        return done();
       }
 
+      await copyFile(
```
sebastianbenz, line 399: would it make sense to parallelise copying the files?

gulpfile.js/build.js
```diff
@@ -392,10 +463,404 @@ function buildPages(done) {
...
+      through.obj((file, encoding, callback) => {
+        const rawJSON = file.contents.toString();
+        const data = JSON.parse(rawJSON);
+
+        for (const page of data.pages) {
+          page.url = new URL(page.url, config.hosts.platform.base).toString();
+        }
+
+        for (const page of data.components) {
+          // addExampleAndPlaygroundLink(page, locale);
```
sebastianbenz, line 810: Should the comments be left here?

gulpfile.js/build.js
```diff
@@ -392,10 +463,404 @@ function buildPages(done) {
...
+      project.absolute('examples/static/**/*'),
+      project.absolute('frontend21/dist/static/**/*'),
+    ])
+    .pipe(gulp.dest(`${project.paths.PAGES_DEST}/static`));
+
+  await gulp
+    .src(`${project.paths.STATICS_DEST}/**/*`)
+    .pipe(gulp.dest(() => `${project.paths.PAGES_DEST}/static`));
+
+  return new Promise((resolve, reject) => {
```
sebastianbenz, line 839: Couldn't we also return gulp.series instead of a new promise? ....my gulp is a bit rusty.=

examples/source/1.components/amp-list.html
```diff
@@ -213,7 +213,7 @@
             noloading
             layout="fixed-height"
             height="654"
-            src="/documentation/examples/api/slow-json-with-items/?delay=10000"
+            src="/documentation/examples/api/slow-json-with-items/?delay=5000"
```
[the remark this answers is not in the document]
patrickkettner: The timeout for netlify's functions on the tier being used for testing (i.e. free) is 10 seconds, so the `slow-json-with-items` demo failed before I reduced the delay

netlify/configs/amp.dev/netlify.toml
patrickkettner: It was added manually, yes. would be trivial to add it during the build step from the contents of the go.yml file if you wanted that ability however.

gulpfile.js/build.js
patrickkettner: I _think_) it was broken without it. If memory serves, `done` needed to be within the function's arity, and thereby also needs to be called once added. `done()` returns a promise as well that gulp is expecting. But I could be misremembering it - gulp kicked my butt on this for a bit

netlify/configs/amp.dev/netlify.toml
sebastianbenz: that'd be good

gulpfile.js/build.js
patrickkettner: I can double check if it is nolonger needed, but I believe it needs to be there

The rendering of the pages happens in two steps - the initial render (that would normally be handled by grow at serve time). That step outputs all of the files, which include a handful of files that have our own templating inside of it that is rendered with nunjucks currently. The `blogItems` bit is there for that initial render. I could easily be wrong, though.

gulpfile.js/build.js
patrickkettner: ah, [this](https://github.com/ampproject/amp.dev/blob/b1e69255e480c62ea20015342360fd36305d4226/gulpfile.js/build.js#L503) got moved around at some point. Thanks!

gulpfile.js/build.js
patrickkettner: nope!

gulpfile.js/build.js
patrickkettner: sure

gulpfile.js/build.js
patrickkettner: since it's two files each, wasnt sure if it made sense to do it more than this. but will update to do that

gulpfile.js/build.js
patrickkettner: nope! left over cruft. thanks!

gulpfile.js/build.js
patrickkettner: No, because we need to `resolve` the outer promise we can't return a series. If it wasn't for [this line](https://github.com/ampproject/amp.dev/blob/b1e69255e480c62ea20015342360fd36305d4226/gulpfile.js/build.js#L845) we could do that, however

gulpfile.js/build.js
sebastianbenz: I've definitely been there as well....

gulpfile.js/build.js
sebastianbenz: What I mean is that the random part does not make sense if we're rendering at build time

gulpfile.js/build.js
sebastianbenz: that'd be great to do before we get this in

sebastianbenz:
@patrickkettner What's the status of this? From my point of view this looks good, once build.js has been split up.

patrickkettner:
@sebastianbenz build passes

sebastianbenz (approved):
thanks! well done!

PR #6521, opened 2022-11-14
dont overwrite rewrites when adding go links

patrickkettner:
Introduced this bug when I switched over to programmatically adding go links


patrickkettner:
@sebastianbenz the build is triggered by this branch. You can follow the `Release: Static Production` action to know when it is released to the demo url

PR #6531, opened 2022-11-21
build updates

patrickkettner:
handful of build changes.
This should build a staging version of the static site if a push or PR is added by someone with write access on the repo.
If it is a commit the branch `production-amp-dev` (doesn't currently exist, just a random branch name I put in), then it will trigger a production deploy.
If the push is only modifying the netlify configs, then we reuse the last build cache, so server fixes should be an order of magnitude faster to deploy as all files would be cached.

### DIFF
--- a/.github/workflows/release-static-production.yml
+++ b/.github/workflows/release-static-production.yml
@@ -214,7 +214,7 @@ jobs:
       - name: Deploying
         run: |
           if ${{ github.ref_name }} != "production-amp-dev" ; then
-            npx gulp staticDeploy
-          else
             npx gulp staticDeploy --stage
+          else
+            npx gulp staticDeploy
           fi
--- a/.github/workflows/release-static-production.yml
+++ b/.github/workflows/release-static-production.yml
@@ -3,15 +3,13 @@ name: 'Release: Static Production'
 
 on:
   workflow_dispatch:
-  push:
-    branches:
-      - static-production-amp-dev
+  pull_request:
 
 jobs:
   queue:
     runs-on: ubuntu-latest
     outputs:
-      should_skip: ${{ steps.skip_check.outputs.should_skip }}
+      should_skip: ${{ steps.skip_check.outputs.should_skip || steps.check_user.check-result == 'false' }}
     steps:
       - id: skip_check
         uses: fkirc/skip-duplicate-actions@master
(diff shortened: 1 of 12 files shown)


.github/workflows/release-static-production.yml
```diff
@@ -3,15 +3,13 @@ name: 'Release: Static Production'
...
-    branches:
-      - static-production-amp-dev
+  pull_request:
 
 jobs:
   queue:
     runs-on: ubuntu-latest
     outputs:
-      should_skip: ${{ steps.skip_check.outputs.should_skip }}
+      should_skip: ${{ steps.skip_check.outputs.should_skip || steps.check_user.check-result == 'false' }}
```
sebastianbenz, line 12: What is this check for?

.github/workflows/release-static-production.yml
```diff
@@ -200,4 +213,8 @@ jobs:
 
       - name: Deploying
         run: |
-          npx gulp staticDeploy
+          if ${{ github.ref_name }} != "production-amp-dev" ; then
```
sebastianbenz, line 216: Do I understand this right, if we're not merging to production-amp-dev, then we should should deploy to prod? 

Sidenote: I find the flag `stage` confusing. Does this mean `staging`? Maybe `prod` would make this more clear.

.github/workflows/release-static-production.yml
patrickkettner: it gets the output of [`this`](https://github.com/ampproject/amp.dev/blob/b42e077300cc035ccfd0c788d750810a44784ad3/.github/workflows/release-static-production.yml#L22-L25). It is checking if the creator has write permission on the repo

.github/workflows/release-static-production.yml
patrickkettner: ah, flipped the logic in gulp, but missed this. I felt like the expected default of `deploy` is to prod, so adding a flag to modify that to the less expected staging was the reasoning. I will change it to deploy to staging by default and to prod only when flagged as `--prod`

sebastianbenz: both is fine as long as the logic is right :-)

sebastianbenz:
Is this still relevant?

patrickkettner:
Yes, other or was just to get future up to date
-- 
patrick

PR #6627, opened 2023-05-20
Fix non-constant variable definition

pothitos:
translation can be overwritten if the language isn't recognized.


patrickkettner:
Thanks!!

PR #6680, opened 2023-10-05
html encode the reflected query parameters

erwinmombay:


banaag (approved):
Please fix the errors before submitting.

PR #6767, opened 2024-02-15
Cleanup GitHub Actions workflows

danielrozenberg:
- Remove jobs that have been manually disabled or were never configured correctly
- Remove reference to the non-existent `ACT` env variable
- Replace `npm install` with `npm ci` (CI jobs should not regenerate package-lock.json)
- Renamed *.yml to *.yaml to keep consistent :)
- ~Update NodeJS version from `14` to `lts`~ _(reverted for now, will deal with this later)_
- Remove the custom call to `action/cache` and replace with the built-in caching mechanism in `action/setup-node`
- Add `workflow-dispatch:` trigger (manual trigger) to all remaining workflows, so we can trigger them on PRs that we want to double-check, even if the PR itself doesn't trigger those workflows automatically


powerivq (approved):
Thanks! Don't know if latest LTS has been tested successfully though? I think some won't work on latest.

danielrozenberg:
> Thanks! Don't know if latest LTS has been tested successfully though? I think some won't work on latest.

Fair enough, I don't know why I assumed for one second that this would Just Work with a simple version bump... Let me double-check this on my fork 😅

danielrozenberg:
Invalidating the approval so I won't accidentally merge this

danielrozenberg:
Let's revert the `lts` change in this PR, I'll deal with that specific change in a separate PR. Re-approval required :)

powerivq (approved):

PR #6770, opened 2024-02-20
Remove dependencies that do not seem to be referenced anywhere in code

danielrozenberg:
None of the packages removed in this PR seem to be used anywhere in the codebase, even with partial string matching. Assuming all tests pass, I think it's safe to merge this PR (we can always partially revert, if anything fails)


powerivq:
Gotta love this. Thanks @danielrozenberg

PR #6775, opened 2024-02-28
Update Node version to 16 and fix other config / API usage for newly-broken deps

danielrozenberg:
See individual commits for each step

Locally, all `npm run test:...` commands pass for me and some but not all the `npm run build:...` commands. It's really difficult to tell which ones are still relevant and which ones are legacy code :/

Importantly, I ran through the steps in `release-static-production.yaml` manually and it does seem to pass with Node 16, so I am optimistic about this PR :D


powerivq (approved):
Thanks!

Issue #6776, opened 2024-03-01
Relative URL used in AMP4email example code


M-J-Robbins:
### Current behavior
The code in the [Advanced Server Request example](https://amp.dev/documentation/examples/email/interactivity-dynamic-content/advanced_server_request/) uses a relative URL which is not valid in AMP4Email and shows an error when viewed in the playground. 

![Error message: The relative URL '/static/inline-examples/data/amp-list-animals.json' for attribute 'src' in tag 'amp-list' is disallowed.](https://github.com/ampproject/amp.dev/assets/7871206/27eb10e0-5b8e-4e40-b37b-3fc8a983f23b)

[Code in GitHub](https://github.com/ampproject/amp.dev/blob/main/examples/source/interactivity-dynamic-content/Advanced_Server_Request.html#L64)

### Expected behavior/code
Expect example code to pass validation

### Reproduction procedure

1. [Open Advanced Server Request example the in playground](https://playground.amp.dev/?url=https://preview.amp.dev/documentation/examples/interactivity-dynamic-content/advanced_server_request)
2. Look at error message

dkjonzy2:
Fix is to add "https://amp.dev" to the beginning of the url so that the full url is "https://amp.dev/static/inline-examples/data/amp-list-animals.json"

I'll submit a PR

dkjonzy2:
PR here waiting review https://github.com/ampproject/amp.dev/pull/6778

PR #6778, opened 2024-03-01
Fix Issue #6776 "Relative URL used in AMP4email example code"

dkjonzy2:
Added "https://amp.dev/" to the beginning of the url so that the full url is "https://amp.dev/static/inline-examples/data/amp-list-animals.json" making it AMP valid


erwinmombay (approved):

erwinmombay:
Fixes https://github.com/ampproject/amp.dev/issues/6776

PR #6788, opened 2024-03-28
Perform PR build test

powerivq:
Currently, there is no PR build tests. We can merge something that breaks. This PR adds the test.


.github/workflows/test-default.yaml
```diff
@@ -7,7 +7,22 @@ jobs:
...
+        run: npm ci
+
+      - name: Running tests
         run: |
-          echo 'we require a job named "test", but not all pull requests trigger an action. So we created this small silly job named "test" that will always pass, just to make sure nothing hangs as a result of not triggering an actual job.'
+          npx gulp updateTestResources
+          npx gulp lintNode
+          npx gulp lintYaml
+          npm run test:platform
+          npm run test:playground
```
danielrozenberg, line 28: These two lines are also in `test-platform.yaml` and `test-playground.yaml` - let's avoid duplication here. We can probably just remove the conditionals on these other files, since these tests are relatively fast and it's better to be safe than sorry here

(basically, replace the `on:` block in these three files with just `on: [dispatch_workflow, push, pull_request]`

.github/workflows/lint.yaml
```diff
@@ -1,13 +1,7 @@
 ---
 name: 'Lint: YAML'
 
-on:
-  workflow_dispatch:
```
danielrozenberg, line 5: Make sure you keep `workflow_dispatch`, this will allow us to manually execute these workflows on arbitrary commits

.github/workflows/lint.yaml
```diff
@@ -27,3 +21,7 @@ jobs:
       - name: Linting YAML files
         run: |
           gulp lintYaml
+
+      - name: Linting JavaScript files
+        run: |
+          gulp lintNode
```
danielrozenberg, line 27: Nit: for single-line run commands you can just put them in the same line:

```suggestion
      - name: Linting YAML files
        run: gulp lintYaml

      - name: Linting JavaScript files
        run: gulp lintNode
```

danielrozenberg (approved):

powerivq:
Updated PR, but @danielrozenberg can you do a final check to make sure I understand that correctly?

danielrozenberg (commented):
Looks good! Ship it :ship:

PR #6826, opened 2024-07-22
Fix amp4email demo XHR endpoint inconsistent header

powerivq:
Gmail team has found out that this endpoint is returning inconsistent Access-Control-Allow-Origin header. It could be because it enables `public` cache. The problem is that this header depends on the origin header and therefore it cannot properly be cached without messing the response header up. This PR disables the cache on the Netlify. Given it is only a demo, I do not expect this to become a significant cost problem.

@zhangsu

### DIFF
--- a/netlify/functions/examples_static_samples_json_cart/examples_static_samples_json_cart.js
+++ b/netlify/functions/examples_static_samples_json_cart/examples_static_samples_json_cart.js
@@ -2,7 +2,7 @@ const cartJSON = require('./cart.json');
 
 const handler = async (ev) => {
   const headers = {
-    'Cache-Control': 'public, max-age=604800, stale-while-revalidate=0',
+    'Cache-Control': 'no-cache',
     'Content-Type': 'application/json',
     'Access-Control-Allow-Origin': ev.headers?.origin || '',
   };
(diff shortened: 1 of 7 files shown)


netlify/functions/examples_static_samples_json_cart/examples_static_samples_json_cart.js
```diff
@@ -2,7 +2,7 @@ const cartJSON = require('./cart.json');
 
 const handler = async (ev) => {
   const headers = {
-    'Cache-Control': 'public, max-age=604800, stale-while-revalidate=0',
+    'Cache-Control': 'no-cache',
```
zhangsu, line 5: Thanks for the fix! It looks like the echo endpoint may have the same issue:

https://github.com/ampproject/amp.dev/blob/cfc06c57705e3466a6ff823bfa875ec7a218a3d2/netlify/functions/examples_api_echo/examples_api_echo.js#L10

That endpoint is also used by the Gmail playground.

powerivq: Modified a few more endpoints.

erwinmombay (approved):

PR #6831, opened 2024-07-24
Do not ignore .git folder since it is actually being used

powerivq:
In the container, it is throwing bunch of errors due to .git is not there.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,7 +1,6 @@
 # If you would like to upload your .git directory, .gitignore file or files
 # from your .gitignore file, remove the corresponding line
 # below:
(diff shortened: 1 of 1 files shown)


erwinmombay (approved):

danielrozenberg:
Hmm... what directory level is being exposed publicly from the container? It's best to avoid exposing .git via the server, so I want to make sure it remains inaccessible

powerivq:
@danielrozenberg The Docker container is a dev container and is not a prod container. `.git` will be in amp.dev repo and it generates a self sustained dist bundle without .git.

